PR #20, opened 2016-10-03
Use sample timeout checking interval from config instead of hard-code…

iamigo:
…d so we can adjust it as needed

### DIFF
--- a/config.js
+++ b/config.js
@@ -35,6 +35,9 @@ const defaultDbUrl = 'postgres://' + pguser + ':' + pgpass + '@' + pghost +
 const ipWhitelist = pe.IP_WHITELIST || '[[0.0.0.0,255.255.255.255]]';
 const iplist = configUtil.parseIPlist(ipWhitelist);
 
+// Check for timed-out samples every 30 seconds if not specified in env var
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -35,6 +35,9 @@ const defaultDbUrl = 'postgres://' + pguser + ':' + pgpass + '@' + pghost +
 const ipWhitelist = pe.IP_WHITELIST || '[[0.0.0.0,255.255.255.255]]';
 const iplist = configUtil.parseIPlist(ipWhitelist);
 
+// Check for timed-out samples every 30 seconds if not specified in env var
+const DEFAULT_CHECK_TIMEOUT_INTERVAL_MILLIS = 30000;
+
```
shriramshankar: 29 seconds or 30 seconds ?

pallavi2209:
Code looks good to me.

shriramshankar:
ok to merge

config.js
iamigo: Heh I went back on what I said... I realized it is just easier for people to understand what they're looking at in the code if it's a round number :)

PR #23, opened 2016-10-05
moving tags to an array field in aspect model + data migration for tags

shriramshankar:
tags refactoring for aspects along with data migration for subject and aspect tags.

### DIFF
--- a/api/v1/swagger.yaml
+++ b/api/v1/swagger.yaml
@@ -343,9 +343,12 @@ paths:
               tags:
                 type: array
                 items:
-                  $ref: "#/definitions/TagRequest"
+                  type: string
+                  maxLength: 60
+                  pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,60}$
                 description: >
-                  Tags associated with this model.
+                  Array of tag names. Tag name: Max 60 characters; case insensitive; allows alpha-numeric characters,
+                  underscore (_) and dash (-). Tag names cannot start with a dash (-).
               relatedLinks:
                 type: array
                 items:
@@ -569,9 +572,12 @@ paths:
               tags:
                 type: array
                 items:
-                  $ref: "#/definitions/TagRequest"
+                  type: string
+                  maxLength: 60
+                  pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,60}$
                 description: >
-                  Tags associated with this model.
+                  Array of tag names. Tag name: Max 60 characters; case insensitive; allows alpha-numeric characters,
+                  underscore (_) and dash (-). Tag names cannot start with a dash (-).
               relatedLinks:
                 type: array
                 items:
@@ -728,9 +734,12 @@ paths:
               tags:
                 type: array
                 items:
-                  $ref: "#/definitions/TagRequest"
+                  type: string
+                  maxLength: 60
+                  pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,60}$
                 description: >
-                  Tags associated with this model.
+                  Array of tag names. Tag name: Max 60 characters; case insensitive; allows alpha-numeric characters,
+                  underscore (_) and dash (-). Tag names cannot start with a dash (-).
               relatedLinks:
                 type: array
                 items:
@@ -792,7 +801,7 @@ paths:
           $ref: "#/responses/genericError"
 
   # ---------------------------------------------------------------------------
-  /aspects/{key}/tags/{akey}:
+  /aspects/{key}/tags/{tagName}:
     x-swagger-router-controller: aspects
     delete:
       security:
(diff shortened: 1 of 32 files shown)


api/v1/swagger.yaml
```diff
@@ -343,9 +343,12 @@ paths:
               tags:
                 type: array
                 items:
-                  $ref: "#/definitions/TagRequest"
+                  type: string
+                  maxLength: 60
+                  pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,60}$
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -569,9 +572,12 @@ paths:
...
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -728,9 +734,12 @@ paths:
...
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -3162,10 +3171,10 @@ paths:
                 items:
                   type: string
                   maxLength: 60
-                  pattern: ^[0-9A-Za-z_\\-]{1,60}$
+                  pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,60}$
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -3331,10 +3340,10 @@ paths:
...
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -3436,10 +3445,10 @@ paths:
...
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -3541,10 +3550,10 @@ paths:
...
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -4835,12 +4803,14 @@ definitions:
         description: >
           TODO
       tags:
-        readOnly: true
         type: array
         items:
-          $ref: "#/definitions/TagResponse"
+          type: string
+          maxLength: 60
+          pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,60}$
```
iamigo: I think this should be {0,59} now

api/v1/swagger.yaml
```diff
@@ -5416,10 +5386,10 @@ definitions:
         items:
           type: string
           maxLength: 60
-          pattern: ^[0-9A-Za-z_\\-]{1,60}$
+          pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,60}$
```
iamigo: I think this should be {0,59} now

db/model/aspect.js
```diff
@@ -110,6 +110,11 @@ module.exports = function aspect(seq, dataTypes) {
         },
       },
     },
+    tags: {
```
iamigo: validation?

shriramshankar: All the validations are in the API layer. Doing it in the db layer will be redundant, I guess.

shriramshankar: Let me know if you want me to add it in the db layer too.

shriramshankar:
@iamigo PR updated with the swagger changes.

pallavi2209:
I think the admin front end tag (get and post) related things needs to be updated with this change. Do we handle that separately?

shriramshankar:
@iamigo  @pallavi2209 
1. filters work
2. perspective loads
3. hierarchy loads about 40% faster
there is something wrong with migration, the tags data get migrated correctly, but it looks like tags wit h '_' in the name get copied multiple times.
http://refocus-pr-25.herokuapp.com/v1/subjects/Salesforce.SFDC_Core.CHI.SP1.CS10    I tweaked the migrating code a couple of times but do not see a better result. Any pointers?
I created a another branch to play with the migration code,
https://github.com/salesforce/refocus/blob/b801984561a5a3655017a8ad62470ccaf2f87d40/migrations/20161002203022-moveTagDataToField.js

iamigo:
Any progress on the data migration? If you want to go over this together on a hangout let me know.

shriramshankar:
@iamigo I have another idea that I want to try. Have trust1 meeting for ssl till 11.30am. Will try it after that and let you know.

shriramshankar:
@iamigo updated the pr with migrations fixes.

iamigo:
OK to merge.

shriramshankar:
@pallavi2209 can you merge it?

PR #24, opened 2016-10-05
#Refocus /static optimization

pallavi2209:
> Requests for all content gets tested for whether the content exists in the /static folder. We can improve app server performance by prefixing all our static contents with /static and then only check for existence of those files on disk when we know we're expecting them to be on disk.

Prefixed /static for all static contents in public folder to increase performance.


iamigo:
Don't the files need to actually move into a /static directory, too?

pallavi2209:
@iamigo This line in index.js `app.use('/static', express.static(path.join(__dirname, 'public')));` says,  you can load the files that are in the public directory from the /static path prefix.

PR #27, opened 2016-10-07
Cache /GET /v1/perspectives/:key

pallavi2209:
Cache /GET /v1/perspectives/:key using redis.

### DIFF
--- a/api/v1/helpers/nouns/perspectives.js
+++ b/api/v1/helpers/nouns/perspectives.js
@@ -15,11 +15,11 @@ const Perspective = require('../../../../db/index').Perspective;
 const config = require('../../../../config');
 
 const m = 'perspective';
-let cacheEnabled = true;
+let cacheEnabled = false;
 
-if (config.enableCachePerspective === 'false' ||
- config.enableCachePerspective === false) {
-  cacheEnabled = false;
+if (config.enableCachePerspective === 'true' ||
+ config.enableCachePerspective === true) {
+  cacheEnabled = true;
 }
 
 module.exports = {
--- a/api/v1/helpers/nouns/perspectives.js
+++ b/api/v1/helpers/nouns/perspectives.js
@@ -9,12 +9,18 @@
 /**
  * api/v1/helpers/nouns/perspectives.js
  */
-'use strict';
+'use strict'; // eslint-disable-line strict
 
 const Perspective = require('../../../../db/index').Perspective;
+const config = require('../../../../config');
 
 const m = 'perspective';
-const cacheEnabled = true;
+let cacheEnabled = true;
+
+if (config.enableCachePerspective === 'false' ||
+ config.enableCachePerspective === false) {
+  cacheEnabled = false;
+}
 
 module.exports = {
   apiLinks: {
--- a/api/v1/helpers/nouns/perspectives.js
+++ b/api/v1/helpers/nouns/perspectives.js
@@ -14,6 +14,7 @@
 const Perspective = require('../../../../db/index').Perspective;
 
 const m = 'perspective';
+const cacheEnabled = true;
 
 module.exports = {
   apiLinks: {
@@ -29,4 +30,5 @@ module.exports = {
   },
   model: Perspective,
   modelName: 'Perspective',
+  cacheEnabled,
 }; // exports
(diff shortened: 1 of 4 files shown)


api/v1/helpers/nouns/perspectives.js
```diff
@@ -26,4 +27,5 @@ module.exports = {
   baseUrl: '/v1/perspectives',
   model: Perspective,
   modelName: 'Perspective',
+  cacheEnabled,
```
theswamis: This should come from an Environment variable, not hardcoded, right?

pallavi2209:
@iamigo updated  the PR

pallavi2209:
@ rebase with master

pallavi2209:
I forgot to assign this PR, and I was like why this PR is waiting so long!

shriramshankar:
A general question/comment on caching. 

I was wondering if we really have an overall gain by caching any of the GET/{object}/:key endpoints? the keys are indexed(or can be indexed if not)  and the response is just one object. probably there is not much difference between hitting the database with index on the key and getting the object from redis ?

Caching the GET/{object} will have more gain, as in this case the database will have to do a full table scan?

shriramshankar:
Also, just noticed that, both GET/perspectives and GET/perspectives/:key load under  ~150 ms. Wondering if we really need to cache them ?

pallavi2209:
Yes, I noticed that too, but I was okay because caching never hurts (I intend to make sure that we ae not increasing response time) and we can have many more entries in perspectives table in future, so it might benefit then. Currently it makes no difference. That is why it was last in my priorities of stories.

pallavi2209:
@iamigo Updated cache to be off by default.

api/v1/helpers/verbs/doGet.js
```diff
@@ -29,11 +30,33 @@ const httpStatus = require('../../constants').httpStatus;
  *  resource type to retrieve.
  */
 function doGet(req, res, next, props) {
-  u.findByKey(props, req.swagger.params)
-  .then((o) => {
-    res.status(httpStatus.OK).json(u.responsify(o, props, req.method));
-  })
-  .catch((err) => u.handleError(next, err, props.modelName));
+  if (props.cacheEnabled) {
+    redisCache.get(req.swagger.params.key.value, (cacheErr, reply) => {
```
iamigo: Let's say perspective "XYZ" was cached based on a call to `GET /v1/perspectives/XYZ?fields=name`. Now when I call `GET /v1/perspectives/XYZ` I'm going to get the cached response, which means the perspective I get back only contains the "name" attribute, but `GET /v1/perspectives/XYZ` without any `?fields=...` is supposed to return *all* the perspective fields...

pallavi2209: Yeah, I did not factor in the possibility of fields here. What if I create the cache key- a string which is combination of object and fields required. And I will remove the cache entry when some update happens, to keep it simple, so that it is fetched from db next time.

iamigo: sounds reasonable

pallavi2209:
@iamigo @shriramshankar Updated PR, now caching with an expiry of 1 minute. We cannot use sequelize hooks to update/delete cache entry because various factors play role in there, like, association updates, cache by name and id, fields options etc. I think caching with an expiry is the simplest and suggested solution here.

api/v1/helpers/verbs/doGet.js
```diff
@@ -29,11 +31,41 @@ const httpStatus = require('../../constants').httpStatus;
...
-  u.findByKey(props, req.swagger.params)
-  .then((o) => {
-    res.status(httpStatus.OK).json(u.responsify(o, props, req.method));
-  })
-  .catch((err) => u.handleError(next, err, props.modelName));
+  if (props.cacheEnabled) {
+    const reqParams = req.swagger.params;
+    let cacheKey = reqParams.key.value;
+
+    // cache key is combination of key and fields in specified order.
```
iamigo: Let's say I call `GET /v1/perspectives/ABC?fields=name,rootSubject` and the response is cached. After 45 seconds, another user calls `GET /v1/perspectives/ABC?fields=rootSubject,name`... will they get the cached response or hit the DB?

iamigo: I don't think it will hurt us to set the cache key to include the list of fields in alphabetical order, but I'm also OK with leaving this as is for now (i.e. accepting that there may be fewer cache hits) and getting performance numbers later to see if the extra time to alphabetize the  field list would have any impact vs. getting a higher cache hit rate.

PR #39, opened 2016-10-12
enable socketio clients to communicate with client using only the protocols set in the config

shriramshankar:


shriramshankar:
All the changes are here. Will do the any other requested changes after my training tomorrow.

viewConfig.js
```diff
@@ -26,10 +26,14 @@ const DEFAULT_THROTTLE_MILLISECS = 4000;
 const realtimeEventThrottleMilliseconds =
  pe.realtimeEventThrottleMilliseconds || DEFAULT_THROTTLE_MILLISECS;
 
+const socketIOtransportProtocol = pe.SOCKETIO_TRANSPORT_PROTOCOL || 'undefined';
```
iamigo: Why using string 'undefined' here? It seems like it has potential for misunderstanding given javascript's "undefined" reserved word. I would rather see something like this:

`const socketIOtransportProtocol = pe.SOCKETIO_TRANSPORT_PROTOCOL || null;`

or

`const socketIOtransportProtocol = pe.SOCKETIO_TRANSPORT_PROTOCOL || '';`

... and then in app.js line 105, just check for `if (clientProtocol) {`

view/perspective/app.js
```diff
@@ -96,10 +96,21 @@ function setupSocketIOClient(persBody) {
...
 
   /*
+   * if the transprotocol is set, initialize the socketio client with
+   *  the transport protocol options. The transProtocol variable is set in 
+   *  perspective.pug
+   */
+  const options = {};
+  const clientProtocol = transProtocol; 
+  if (clientProtocol !== 'undefined') {
+    options.transports = clientProtocol.replace(/\s*,\s*/g, ',').split(',');
```
iamigo: Can you add a comment about this regex and the format the socket.io options.transports requires?

shriramshankar:
pr updated
@theswamis added the option to force client to use websocket using the query param. 
@iamigo used '' instead of undefined.

shriramshankar:
?protocol=websocket is the query param to force the client to use communicate over websocket.

shriramshankar:
@pallavi2209 @iamigo  web concurrency also seems to be working. Can you check if it set the config var right ?

pallavi2209:
@shriramshankar Thats great! The config var looks right to me.

PR #40, opened 2016-10-12
Make Upserts More Efficient to improve performance

pallavi2209:
Previously:
----------
1. Find subject and aspect from sample name - 2 db calls
2. Get sample from db, 1 db call
3. If sample exists, update sample, else create sample, 1 call

Total:
Sample create: 4 calls
Sample update: 3 calls

Now:
-------
1. Get sample by sample name. 1 db call
2. If sample exists, update (1 call), else, get subject and aspect from db, and create sample (3 calls)

Total:
Sample create: 4 calls
Sample update: 2 calls

We  cannot do Sample.update directly because it does not return sample object, just the number of rows updated, so we will have to make a sample.find call anyway.

### DIFF
--- a/db/model/sample.js
+++ b/db/model/sample.js
@@ -170,31 +170,43 @@ module.exports = function sample(seq, dataTypes) {
        *  encountered while performing the sample upsert operation itself.
        */
       upsertByName(toUpsert, isBulk) {
-        let subjasp;
+        let sampleExists = true;
         return new seq.Promise((resolve, reject) => {
-          u.getSubjectAndAspectBySampleName(seq, toUpsert.name)
-          .then((sa) => {
-            subjasp = sa;
-            toUpsert.subjectId = sa.subject.id;
-            toUpsert.aspectId = sa.aspect.id;
-          })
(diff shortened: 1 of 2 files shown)


iamigo:
Looks good. OK to merge once @shriramshankar has reviewed.

db/model/sample.js
```diff
@@ -170,32 +170,44 @@ module.exports = function sample(seq, dataTypes) {
...
-          .then(() => Sample.findOne({
+          // get sample by name
+          Sample.findOne({
             where: {
-              subjectId: subjasp.subject.id,
-              aspectId: subjasp.aspect.id,
-              isDeleted: NO,
+              name: {
+                $iLike: toUpsert.name,
+              },
```
shriramshankar: what about using eq vs like ?

pallavi2209: @shriramshankar After recent changes, we are indexing only name, so we are using iLike to match case-insensitive name.
And I am not able to find an operator to match case-insensitive without regex. So, using iLike only.

PR #42, opened 2016-10-13
GET /v1/perspectives?fields=name should NOT include the lens association

iamigo:


api/v1/helpers/verbs/utils.js
```diff
@@ -371,18 +372,37 @@ function deleteAJsonArrayElement(jsonArray, elementName) {
...
-      const f = fields[i];
-      if (props.fieldScopeMap && props.fieldScopeMap[f]) {
-        scopes.push(props.fieldScopeMap[f]);
+    if (props.fieldAbsenceScopeMap) {
+      const keys = Object.keys(props.fieldAbsenceScopeMap);
+      for (let i = 0; i < keys.length; i++) {
+        const fieldName = keys[i];
+        if (fields.indexOf(fieldName) === NOT_FOUND) {
+          const scopeName = props.fieldAbsenceScopeMap[fieldName];
+          scopes.push(scopeName);
```
shriramshankar: Do we need to check if props.fieldAbsenceScopeMap[fieldName] returns null or not?

iamigo: I'll add that.

pallavi2209:
@iamigo merge the master changes?

api/v1/helpers/verbs/utils.js
pallavi2209: Given the fieldName is coming from fieldAbsenceScopeMap only, I dont think the check for props.fieldAbsenceScopeMap[fieldName] was necessary.

pallavi2209:
@iamigo So with these changes GET /v1/perspectives?fields=name do not return lens association part, which is this one:
screen shot 2016-10-13 at 3 15 34 pm
But do you want http://localhost:3000/v1/perspectives/:key also not return lens association part??
If yes, then why not change default scope only? Sorry, I am a bit confused on intentions here.

iamigo:
If the user doesn't specify ?fields=... then I want to return the extra lens fields, but if they are explicitly narrowing the list of fields, then I don't want to add those fields since the user hasn't explicitly asked for them.

pallavi2209:
And what about GET /v1/perspectives/:key. With this change, the response does not include lens fields anymore. Do you intend to make that change?

iamigo:
@pallavi2209 good catch that was not my intention, adding test and fixing now

iamigo:
thank you for your diligence looking at this, @pallavi2209!

PR #43, opened 2016-10-13
fixed API GET operations should return results in alphabetical order and respects query parameter ordering

harshkothari410:
- Now if defaultOrder is defined in props then it will take that defaultOrder and return the result as per that. If defaultOrder is not defined then it will take sort order based on model. 

- Fixed all swagger parameter sort option

### DIFF
--- a/api/v1/helpers/verbs/findUtils.js
+++ b/api/v1/helpers/verbs/findUtils.js
@@ -145,9 +145,9 @@ function toSequelizeOrder(sortOrder, modelName) {
 function options(params, props) {
   const opts = u.buildFieldList(params);
 
-  // Specify the sort order. If defaultOrder is defined in props then
-  // update sort order otherwise take value from model defination
-  if (props.defaultOrder) {
+  // Specify the sort order. If defaultOrder is defined in props or sort value
+  // then update sort order otherwise take value from model defination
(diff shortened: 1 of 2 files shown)


api/v1/helpers/verbs/findUtils.js
```diff
@@ -145,9 +145,12 @@ function toSequelizeOrder(sortOrder, modelName) {
 function options(params, props) {
   const opts = u.buildFieldList(params);
 
-  // Specify the sort order
-  const ord = (params.sort ? params.sort.value : null) || props.defaultOrder;
-  opts.order = toSequelizeOrder(ord, props.modelName);
+  // Specify the sort order. If defaultOrder is defined in props then
+  // update sort order otherwise take value from model defination
+  if (props.defaultOrder) {
+    const ord = (params.sort ? params.sort.value : null) || props.defaultOrder;
```
iamigo: No... it should still sort by params.sort.value even if props.defaultOrder is empty or undefined.

harshkothari410: Thats good point. I missed it.

pallavisingh9:
Can we have some corresponding tests?

pallavi2209:
Can we have some corresponding tests?

PR #44, opened 2016-10-13
moved the sample timeout query to a worker dyno

shriramshankar:


shriramshankar:
@iamigo  changes to move the timeout query to a worker dyno were fairly simple, so, created a PR instead of doing a spike.

pallavi2209:
@shriramshankar Just to confirm, with these changes, doTimeout will run just on one worker dyno and server will run on multiple dynos?

shriramshankar:
@pallavi2209 yes, that is the goal.

shriramshankar:
this can be extended to handle the bulk upsert , without much effort.

shriramshankar:
@iamigo  a config variable asking asking if we want to use worker dynos or not?

iamigo:
Yeah, maybe an environment variable like "HEROKU_WORKER_DYNOS" (true|false). The *absence* of that variable means do it like we always have, but when that environment variable exists and is set to true, then do it this new way.
Would that make sense?

shriramshankar:
Sure, will do it this way.  Thanks.

shriramshankar:
@iamigo Wondering if we can move the the doTimeout query at line 97 and 102 , to somewhere in the bottom of the index.js file?

iamigo:
Yeah, @shriramshankar, it doesn't matter where in the sequence it is. End is fine.

shriramshankar:
@iamigo pr updated with the requested changes.

worker/index.js
```diff
@@ -18,5 +18,7 @@ const env = conf.environment[conf.nodeEnv];
 
 const dbSample = require('../db/index').Sample;
 
-setInterval(() => dbSample.doTimeout(), env.checkTimeoutIntervalMillis);
+if (conf.enableWorkerDynos === true || conf.enableWorkerDynos === 'true') {
```
iamigo: in general, I think we should put all of these 'true'/true conditions for env vars in config.js so that the modules which use them can just do a simple `if (conf.enableWorkerDynos)...`

index.js
```diff
@@ -102,6 +102,12 @@ function start() { // eslint-disable-line max-statements
     });
   }
 
+  // if worker dynos are not enabled run the Sample timeout query here.
+  if (conf.enableWorkerDynos === false || conf.enableWorkerDynos === 'false') {
```
iamigo: in general, I think we should put all of the conditions for env vars in config.js so that the modules which use them can just do a simple `if (conf.enableWorkerDynos)...` or `if (!conf.enableWorkerDynos)...`

config.js
```diff
@@ -37,7 +37,8 @@ const iplist = configUtil.parseIPlist(ipWhitelist);
 
 // Check for timed-out samples every 30 seconds if not specified in env var
 const DEFAULT_CHECK_TIMEOUT_INTERVAL_MILLIS = 30000;
-const enableWorkerDynos = pe.HEROKU_WORKER_DYNOS || false;
+const enableWorkerDynos = pe.HEROKU_WORKER_DYNOS === 'true' ||
```
iamigo: I know this is really nitpicky, but what about case sensitivity?

/true/i

Procfile
```diff
@@ -1 +1,2 @@
 web: npm start
+worker: node worker/index.js
```
iamigo: add Vijay's GC settings here

shriramshankar:
@iamigo updated the PR with gc options for worker dyno.

shriramshankar:
@iamigo @pallavi2209  merge conflicts resolved.

shriramshankar:
@iamigo can we merge this? This PR keeps having a merge conflict every time, some other pr is merged.

PR #45, opened 2016-10-14
Do db sync on server start for table and index creation

iamigo:
(since table creation via migration doesn't do everything we need it to do)

Also add a migration to remove the vestigial GlobalConfig table if it had already been partially created.

### DIFF
--- a/migrations/20161013215314-dropglobalconfig.js
+++ b/migrations/20161013215314-dropglobalconfig.js
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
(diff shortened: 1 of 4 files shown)


migrations/20161013215314-dropglobalconfig.js
```diff
@@ -0,0 +1,39 @@
...
+ * time, this migration tests for the existence of the "createdAt" field and
+ * only drops the table if that field does not exist. After this migration,
+ * when the server starts up and index.js calls sync on all the models, the
+ * GlobalConfig table will be created properly.
+ */
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    return queryInterface.describeTable('GlobalConfig')
```
pallavi2209: I got this error when I do postinstall:
```
Using environment "development".
== 20161013215314-dropglobalconfig: migrating =======

stderr: No description found for "GlobalConfig" table. Check the table name and schema; remember, they _are_ case sensitive.
```
The error goes when I change the name of the table to "GlobalConfigs"

pallavi2209:
I get Unhandled rejection SequelizeDatabaseError when I do resetdb. Can you check that on your end?

PR #46, opened 2016-10-14
Added feature flag for upsert optimization

pallavi2209:
Turned off by default so that we can go back to our previous implementation.

### DIFF
--- a/db/model/sample.js
+++ b/db/model/sample.js
@@ -170,8 +170,7 @@ module.exports = function sample(seq, dataTypes) {
        *  encountered while performing the sample upsert operation itself.
        */
       upsertByName(toUpsert, isBulk) {
-        if (config.optimizeUpsert === 'true' ||
-         config.optimizeUpsert === true) {
+        if (config.optimizeUpsert) {
           let sampleExists = true;
           return new seq.Promise((resolve, reject) => {
             // get sample by name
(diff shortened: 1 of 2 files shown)


db/model/sample.js
```diff
@@ -170,49 +170,84 @@ module.exports = function sample(seq, dataTypes) {
...
        */
       upsertByName(toUpsert, isBulk) {
-        let sampleExists = true;
-        return new seq.Promise((resolve, reject) => {
-          // get sample by name
-          Sample.findOne({
-            where: {
-              name: {
-                $iLike: toUpsert.name,
+        if (config.optimizeUpsert === 'true' ||
```
iamigo: We should put the truthiness check (`(config.optimizeUpsert === 'true' || config.optimizeUpsert === true)`) in config.js so that all the other places that might need to check the variable can just do the simpler easier-to-read `if (config.optimizeUpsert)`

pallavi2209:
@iamigo updated the PR

iamigo:
@harshkothari410 please merge this once you review and approve the changes

iamigo:
@shriramshankar

iamigo:
@shriramshankar can you review this change? need to get this in before we can do another deploy. thx!

shriramshankar:
@iamigo the changes look good but wondering if it is a good to use a feature flag for even optimizations. Kind of think, there are going to be feature flags all over the code. If we still need this will merge it.

iamigo:
> use a feature flag for even optimizations
@shriramshankar  -- yes, we want to build in the ability to turn it on/off in production without rolling back a deployment
once we are comfortable with it, we can delete all the old code and the env. var.

PR #50, opened 2016-10-18
Trace perf

shriramshankar:


index.js
```diff
@@ -19,6 +19,20 @@ const WORKERS = process.env.WEB_CONCURRENCY || 1;
  */
 function start() { // eslint-disable-line max-statements
   const conf = require('./config');
+
+  if (conf.traceAPIKey) {
+    /*
```
iamigo: add something to readme about the Trace addon, link to it, how to configure it, etc.

shriramshankar:
@iamigo this one and the worker dyno changes are on the same files. If there are no more changes for this, can we merge this first please?

PR #54, opened 2016-10-18
do not throw error when trace-service-name env variable is not set

shriramshankar:


shriramshankar:
@iamigo removed the error. Sorry about that!

PR #60, opened 2016-10-19
changing image size

meynet:


meynet:
The image was too big for the site

PR #66, opened 2016-10-20
introduce temporary /perspectiveBeta to improve perspective page load time

iamigo:
unwinds some of the promises to load all the various pieces asynchronously and render the right pieces at the right time


iamigo:
Note: the changes are almost all in app.js, just a little in PerspectiveController.js -- all the other files are identical to their counterparts under /view/perspective

shriramshankar:
tested this locally, works fine.

PR #68, opened 2016-10-21
#Refocus get /subjects filtering by tag working

pallavi2209:
Using env var for new changes.
When env var us enables, subjects are filtered by tags, like GET /v1/subjects?tags=tag1,tag2
Added tests behind config var.

### DIFF
--- a/api/v1/swagger.yaml
+++ b/api/v1/swagger.yaml
@@ -3094,6 +3094,17 @@ paths:
             Filter by parentAbsolutePath; asterisk (*) wildcards ok.
           required: false
           type: string
+        -
+          name: tags
+          in: query
+          items:
+            type: string
+            maxLength: 60
+            pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,59}$
+          description: >-
+            Filter by tags.
+          required: false
+          type: array
       responses:
         200:
           description: >-
--- a/tests/api/v1/subjects/get.js
+++ b/tests/api/v1/subjects/get.js
@@ -19,6 +19,7 @@ const u = require('./utils');
 const Subject = tu.db.Subject;
 const path = '/v1/subjects';
 const expect = require('chai').expect;
+const filterSubjByTags = require('../../../../config').filterSubjByTags;
 
 describe(`api: GET ${path}`, () => {
   let token;
@@ -30,10 +31,12 @@ describe(`api: GET ${path}`, () => {
   const us = {
     name: `${tu.namePrefix}UnitedStates`,
     description: 'country',
+    tags: ['US'],
   };
   const vt = {
     name: `${tu.namePrefix}Vermont`,
     description: 'state',
+    tags: ['US', 'NE'],
   };
 
   before((done) => {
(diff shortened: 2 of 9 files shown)


api/v1/swagger.yaml
```diff
@@ -3094,6 +3094,17 @@ paths:
             Filter by parentAbsolutePath; asterisk (*) wildcards ok.
           required: false
           type: string
+        -
+          name: tags
+          in: query
+          items:
+            type: string
+            maxLength: 60
```
iamigo: Why 60 here? (Might need to allow for longer since you could build up a list of includes or excludes.)

tests/api/v1/subjects/get.js
```diff
@@ -19,6 +19,7 @@ const u = require('./utils');
 const Subject = tu.db.Subject;
 const path = '/v1/subjects';
 const expect = require('chai').expect;
+const filterSubjByTags = require('../../../../config').filterSubjByTags;
```
iamigo: will always be false unless we explicitly set env var when we call the test (see examples in package.json)

api/v1/swagger.yaml
pallavi2209: Each tag length is max 60 char long, as restricted by model.
Also, the changes in this PR only handles includes at sql level like `v1/subjects?tags=mytag`. 
Do we need excludes here as well? I assumed the story was about includes only to get trust data working. I can create another PR if we need excludes as well.

iamigo: ok we can get this through as is, but i think we want to follow up with story to allow `v1/subjects?tags=tag1,tag2,tag3` or `v1/subjects?tags=-tag10,tag11` for parity with the analogous functionality on the /hierarchy endpoint

iamigo:
uh oh coverage decreased! :(

pallavi2209:

api/v1/swagger.yaml
shriramshankar: the /hierarchy endpoint filters are case insensitive too.

pallavi2209:
@iamigo So, I looked into it because I have tests for the new code. I think the test coverage does not cover the tests written for specific environment. So, coverage identifies the new code as not covered. @harshkothari410 can you confirm this?

iamigo:
Yes, that's right--Harsh is going to add all those in this week I hope.

api/v1/swagger.yaml
pallavi2209: @shriramshankar Just checked, we can add different case tags to a subject, like I can have
```
 tags = [
    "mytag",
    "Cali",
    "mytaG",
    "MYTAG"
  ]
```
Is this expected behavior?
Also, if we need to filter case insensitive, there are 2 ways:
1. Filter on API layer instead of query level, like hierarchy endpoint. Disadvantage: get all subjects first and then extra processing to filter.
2. store tags in lowercase in db. This is more efficient way to do, if we agree.

I can add this feature to the new story to filter by excludes as well.

harshkothari410:
Yes environment test are remaining. It covers API and DB test only right now. I will add those because that requires some research because we need to compile some stuff for view test or change environment for env related test. @pallavi2209 @iamigo

api/v1/swagger.yaml
shriramshankar: yes you can have mixed cases for  tags.

shriramshankar: Since filtering for the hierarchy end point was done in the API layer, it made sense to store the tags the way it was created by the user.

PR #78, opened 2016-11-01
corrected hierarchy filter query on perspective

shriramshankar:


shriramshankar:
@harshkothari410 looks like with the code coverage changes, the db tests does not run any more. Also, the db reset does not happen any more. Not sure if these were the intended; if not can you take a look any time you are free?

view/perspective/app.js
```diff
@@ -249,36 +249,41 @@ function handleLibraryFiles(res) {
  */
 function getFilterQuery(p) {
   let q = '?';
+
   if (p.aspectFilter && p.aspectFilter.length) {
-    const sign = p.aspectFilterType === 'INCLUDE' ? '=' : '=-';
-    q += 'aspect' + sign + p.aspectFilter.join();
+    const sign = p.aspectFilterType === 'INCLUDE' ? '' : '-';
+    q += 'aspect' + '=' + sign + 
+        p.aspectFilter.join().replace(/,/g, ',' + sign);
```
iamigo: Does it really need the sign before each one?
wouldn't `...?aspect=-A,B,C` set up an EXCLUDE filter for all three aspects without having to put the minus sign before each one `...?aspect=-A,-B,-C`?

iamigo: The API docs for GET /v1/subjects/.../hierarchy are ambiguous about this:

> Comma-separated list of aspect names to include/exclude. For example, ?aspect=FOO,BAR will only return subjects in the hierarchy with samples for those two aspects (and all those subjects' ancestors up to the specified root of the requested hierarchy). Prefix the aspect name with a negative sign to indicate that a sample with that aspect should be excluded. For example, ?aspect=-BAZ will return only the subjects (and its hierarchy) that have samples with aspect name not equal to BAZ. Subjects without samples are not included in the result set

The api docs don't explicitly say anything about excluding multiple items.

PR #80, opened 2016-11-02
fix the problem where the value of DISABLE_HTTP would be ignored

iamigo:


pallavi2209:
@iamigo Do you think we can improve test coverage by having a test for toggle.js? Possible?

iamigo:
Test coverage isn't measuring all the tests yet so we can't trust these percentages. There's a story to fix the "test" script in package.json.

PR #92, opened 2016-11-08
Get OKTA SAML app working from OKTA admin

pallavi2209:
A Refocus deployment using OKTA for auth was not getting redirected when clicking on SAML app in OKTA admin (which was expected and was happening before Relay state redirection came in).

Added a condition that if relay state is set (which gets set when we click SSO button), redirect to the url requested, else redirect to home page (when SAML app is clicked from other source).


iamigo:
"spare time" -- nope!
create new user story and let the team prioritize :)

PR #95, opened 2016-11-11
Minify app.js

shriramshankar:
minifying app.js reduces the size by another 30KB and the content download time reduces by about 30ms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "supertest": "^1.1.0",
     "swagger-tools": "^0.10.1",
     "throng": "^1.0.1",
+    "uglifyify": "^3.0.4",
     "underscore": "^1.8.3",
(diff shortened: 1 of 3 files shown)


shriramshankar:
@iamigo @pallavi2209 looks like minifying app.js reduced its content download time by a bit.

package.json
```diff
@@ -124,6 +124,7 @@
     "supertest": "^1.1.0",
     "swagger-tools": "^0.10.1",
     "throng": "^1.0.1",
+    "uglifyify": "^3.0.4",
```
iamigo: Awesome! I had been hung up on a version conflict with the "uglify" library so I thought we wouldn't be able to minify stuff until that got resolved.

iamigo: @meynet license ok on this new external dependency?

PR #97, opened 2016-11-11
#Refocus api token database updates

pallavi2209:
New Model: Token

Columns:
- name - allowNull: false, validate using db/constants.nameRegex, use db/constants.fieldlen.normalName for field length
- token - allowNull: false, string
- isDeleted - dataTypes.BIGINT, defaultValue: 0, allowNull: false
- isDisabled - follow the same pattern we use for "isDeleted" so a record is considered "disabled" when this field has a non-zero value

Indexes:
- unique on lower(name)+createdBy+isDeleted

Associations:
- createdBy association to User

Scopes:
- default scope should exclude the token field and should order by name ascending

Hooks:
- beforeCreate - reuse db/helpers/userUtils.hashPassword to store the encrypted version of the token

Permissions:
- Profile checks will be in the API layer, not the DB layer


db/model/token.js
```diff
@@ -0,0 +1,103 @@
...
+  const Token = seq.define('Token', {
+    id: {
+      type: dataTypes.UUID,
+      primaryKey: true,
+      defaultValue: dataTypes.UUIDV4,
+    },
+    name: {
+      type: dataTypes.STRING(constants.fieldlen.normalName),
+      allowNull: false,
+    },
```
shriramshankar: Do we need a name field ?

db/model/token.js
```diff
@@ -0,0 +1,103 @@
...
+    token: {
+      type: dataTypes.STRING,
+      allowNull: false,
+      unique: true,
+    },
```
shriramshankar: @iamigo @pallavi2209 Any thoughts on having a default values for all the data types? We might be able to get ride of the responsify function at one point of time.

db/model/token.js
iamigo: If a user has multiple tokens, the name field makes it possible to identify a particular token one might wish to disable or delete

db/model/token.js
iamigo: Shriram can you create a spike for team to discuss this? (But let's not let it hold up this particular PR.)

db/model/token.js
```diff
@@ -0,0 +1,103 @@
...
+        return new seq.Promise((resolve, reject) =>
+          u.hashPassword(seq, inst.get('token'))
+          .then((hash) => inst.set('token', hash))
+          .then(() => resolve(inst))
+          .catch((err) => reject(err))
+        );
+      },
+    },
+    name: {
+      singular: 'Lens',
```
iamigo: ?

pallavi2209:
@iamigo @shriramshankar updated the PR.

PR #99, opened 2016-11-14
Db script cleanup

iamigo:
Resolve the timing error when doing a sync on server start.
Refactor all the scripts under ./db.
Get rid of resetdb--we only need checkdb now.
Also: update node version from 4 to 6.


pallavi2209:
When I start node, I get this:

> pallavi.singh$ node .
> [./db/utils.doImport] Import 11 models... OK
> Listening on port 3000
> [./db/utils.doImport] Post-import... OK


Required console logs?

iamigo:
console log is only when node env is dev

pallavi2209:
Right :thumbsup:

pallavi2209:
code looks good to me. I tried deploying on my personal app, there I get, error: permission denied for relation SequelizeMeta (maybe cz it is not on private space?) Trying now on staging.

iamigo:
> permission denied for relation SequelizeMeta

Hmmm. I haven't encountered that :(

pallavi2209:
Works on staging, yay! :raised_hands:

pallavi2209:
So, this is what I tested:
1. Instead of resetdb, now we can do `npm run dropdb && npm run checkdb` to get our db in latest state (but without any data).
2. Keep the data, but update db (with possibly new migrations), do `npm run checkdb`
3. If we create new table, it will be automatically added in db when we start node.

Ian, does this sounds correct?

iamigo:
exactly, yes

db/createOrDropDb.js
```diff
@@ -9,59 +9,58 @@
...
+  if (keys.includes('init')) {
+    u.createOrDropDb(pgtools.createdb)
+    .then((res) => {
+      u.clog('createOrDropDb', '', res);
+    })
+    .catch((err) => {
+      u.clog('createOrDropDb', '', err.message);
+    });
+  } else {
+    u.createOrDropDb(pgtools.dropdb)
```
shriramshankar: Not sure if I am reading it right, but does this mean if the key includes anything other than "init", it will drop the db?

iamigo: This isn't a change in behavior, but also note that it validates the args so the only allowable options are "init" and "drop" and it has to be one of these. So, if you try to pass argument "foo" it will error out and if you try to pass no arg, it will error out.

shriramshankar:
When I start the app I get "[./db/utils.doImport] Import 11 models... OK" . Is this fine ?

iamigo:
> When I start the app I get "[./db/utils.doImport] Import 11 models... OK" . Is this fine ? Ok to merge if we are fine with this.

@shriramshankar that console logging is only when node env is development

PR #108, opened 2016-11-17
Delete 00-welcome.md

meynet-salesforce:
There is nothing here and it is frustrating to do an extra click for the quickstart. Prefer to start simple


docs/docs/00-welcome.md
```diff
@@ -1,6 +0,0 @@
----
```
iamigo: need to update /docs/_layouts/default.html too

iamigo: @meynet can you add the change to /docs/_layouts/default.htm to this PR so we can close this out and merge it?

meynet-salesforce: sorry missed this comment. Added that change

PR #91, opened 2016-11-07
remove old prop from db update

annyhe:
This change aims to improve performance by reducing data sent from scoket.io.
Originally updated objects would be sent with {
key: { old: oldObj, new: newObj }
}
This change takes out the old object, so the updated object is sent with {
key: { new: newObj }
}

PR #113, opened 2016-11-23
send ADD+DELETE realtime events when subject tags change (instead of UPDATE event)

harshkothari410:
Without this fix, perspectives which filter on subject tags will not get the right realtime events.


db/model/subject.js
```diff
@@ -342,6 +342,12 @@ module.exports = function subject(seq, dataTypes) {
           );
         }
 
+        // Send event for updation of tags, first delete and then add
```
iamigo: With this block of code as it is here, if tags changed, we will basically send a delete and an add, then if the record is published, we'll send an update, and if it's not published, another delete. 

This is wrong.

Your new code has to be inside of the "isPublished" logic below. Follow all the existing isPublished logic, but if tags were changed, send out a "delete" event followed by an "add" event INSTEAD OF the normal "update" event. If tags were not changed, just send out the normal "update" event as it does now.

harshkothari410: Yeah I got the issue.

db/model/subject.js
```diff
@@ -344,8 +344,14 @@ module.exports = function subject(seq, dataTypes) {
 
         if (inst.getDataValue('isPublished')) {
           if (inst.previous('isPublished')) {
-            common.publishChange(inst, eventName.upd, changedKeys,
+            // Send event for updation of tags, first delete and then add
```
iamigo: Yes! Better!

One small thing in this comment line: "updation" is not a word. I think the comment should be:
```
/*
 * If tags were updated, send a "delete" event followed by an "add" event
 * so that perspectives using subject tag filters will get the right realtime
 * events. If subject tags were not updated, just send the usual "update"
 * event.
 */
```

harshkothari410:
@iamigo I don't think so we can test this thing. In tests/db/helper/common.js there is test related to publishChange function. But we have not modified that function just used that thing to generate the event in Subjects. Is there any tests we have written for events ?

iamigo:
We'll need to come up with a way to mock the socket.io stuff so we can test this properly. We'll create a new user story for that.

db/model/subject.js
```diff
@@ -344,8 +344,19 @@ module.exports = function subject(seq, dataTypes) {
...
           if (inst.previous('isPublished')) {
-            common.publishChange(inst, eventName.upd, changedKeys,
+            /*
+             * If tags were updated, send a "delete" event followed by an "add"
+             * event so that perspectives using subject tag filters will get
+             * the right realtime events. If subject tags were not updated,
+             * just send the usual "update" event.
+             */
+            if (inst.changed('tags')) {
+              common.publishChange(inst, eventName.del);
```
iamigo: Please add a TODO comment here describing the difficulties in writing unit tests for this right now.

shriramshankar:
Ok to merge, once Ian's  requested are done.

harshkothari410:
@shriramshankar as per our talk I made changes just review them.

db/model/subject.js
```diff
@@ -344,8 +344,24 @@ module.exports = function subject(seq, dataTypes) {
...
+            if (inst.changed('tags')) {
+              common.publishChange(inst, eventName.del, ['tags'],
+                ignoreAttributes);
+              common.publishChange(inst, eventName.add, ['tags'],
+                ignoreAttributes);
+            } else {
+              common.publishChange(inst, eventName.upd, changedKeys,
               ignoreAttributes);
+            }
           } else {
```
shriramshankar: I dont think we need these changes.

realtime/utils.js
```diff
@@ -12,6 +12,11 @@
 
 'use strict'; // eslint-disable-line strict
 const constants = require('./constants');
+const eventName = {
+  add: 'refocus.internal.realtime.subject.add',
+  upd: 'refocus.internal.realtime.subject.update',
+  del: 'refocus.internal.realtime.subject.remove',
+};
```
shriramshankar: We have to similar changes for aspects too. So, may be make it generic ?

db/model/subject.js
harshkothari410: Why ? So we don't need to send delete and add events ?

shriramshankar: The changes you did in the utils will take care of everything.  Can you test without these changes to make sure?

realtime/utils.js
```diff
@@ -47,8 +52,12 @@ function getNewObjAsString(key, obj) {
...
-function parseObject(messgObj) {
-  if (messgObj.new) {
+function parseObject(messgObj, key) {
+  if (key === eventName.del && messgObj.new) {
+    return messgObj.old;
+  }
+
+  if (key === eventName.add && messgObj.new) {
     return messgObj.new;
   }
```
shriramshankar: You might want to add some comments here to say why we ended by doing like this. Also, you can combine these "if" conditions to use else if's and else.

db/model/subject.js
harshkothari410: We need to send add and delete event @shriramshankar otherwise it will send update event.

shriramshankar: What I was asking is do you need specifically add ['tags'] in the place of changedKeys? changedKeys by itself will include tags and you may not need to specifically add "tags".

iamigo:
Approved but waiting to merge until I have fixed a MultiTable lens bug: 

```
Uncaught TypeError: Cannot read property 'toLowerCase' of undefined
    at Function.isUnderRootSubject (<anonymous>:19206:33)
    at onSubjectRemove (<anonymous>:18852:21)
    at Function.handle (<anonymous>:18878:8)
    at evt.detail.forEach (<anonymous>:188:56)
    at Array.forEach (native)
    at HTMLDivElement.onRealtimeChange (<anonymous>:188:17)
    at createAndDispatchLensEvent (http://localhost:3000/static/perspective/app.js:21829:540)
    at http://localhost:3000/static/perspective/app.js:21829:636
isUnderRootSubject @ no-conflict.js:13
onSubjectRemove @ no-conflict.js:13
handle @ no-conflict.js:13
evt.detail.forEach @ no-conflict.js:13
onRealtimeChange @ no-conflict.js:13
createAndDispatchLensEvent @ eventsQueue.js:56
(anonymous) @ eventsQueue.js:68
```

iamigo:
My previous comment about a lens bug was incorrect--it was a bug in the realtime event layer where we were wrapping adds and removes inside a "new" property, but only updates should be wrapped that way.

iamigo:
@harshkothari410  @annyhe  @shriramshankar can you please give this PR another look before we merge it? While testing the fix for this, Harsh ran into a bug we introduced back in https://github.com/salesforce/refocus/pull/91 where we wrapped the object in "new" for every event type. Lenses actually expect the realtime event to only be wrapped inside "new" if it's an "update" event (not an add or remove event), so we were seeing lens errors.

So... I added a fix to realtime/utils.js getNewObjAsString function to resolve this.

shriramshankar:
Sure, will look at it.

realtime/utils.js
```diff
@@ -47,8 +61,16 @@ function getNewObjAsString(key, obj) {
  * @param {Object}  messgObj - Message object received from the redis channel.
  * @returns {Object} - returns the parsed message object.
  */
-function parseObject(messgObj) {
+function parseObject(messgObj, key) {
+  // If event is subject delete then send the old subject so that namespace
+  // filter can send the delete event to perspectives
   if (messgObj.new) {
+    return key === eventName.del ? messgObj.old : messgObj.new;
+  }
```
annyhe: this is confusing. we only need the changes in getNewObjAsString.

realtime/utils.js
```diff
@@ -30,17 +30,26 @@ function isThisSubject(obj) {
...
   const wrappedObj = {};
-  wrappedObj[key] = { new: obj };
+  if (key.endsWith('update')) {
+    wrappedObj[key] = { new: obj };
+  } else {
+    wrappedObj[key] = obj;
+  }
+
   return JSON.stringify(wrappedObj);
 }
```
shriramshankar: Can we do something like, if the object has a new key then return the object with the new key, if not return the object itself?

shriramshankar: @iamigo @annyhe Just went through the realtime module code. Looks like we do not need to call getNewObjAsString. So we can remove the following line 

const newObjectAsString = rtUtils.getNewObjAsString(key, obj);

and at line 32 when we emit the object, we just need to emit "obj" that was created in line 18.

iamigo: @shriramshankar we have to transform the object to a string before emitting it, and if we do that as part of the parseObject implementation it breaks other uses of that function

iamigo: @annyhe @shriramshankar I'm inclined to leave the parseObject and getNewObjAsString functions as they are here now--automated tests and manual browser tests are passing with subject tag changes pushing out the right sequence of realtime events now so that a perspective filtered by subject tag will get the right updates

shriramshankar: Just looked the file history. Look like we converted the object to string just before emitting something like 
io.of(nsp).emit(key, JSON.stringify(mssgObj));
If you want to leave it the way it is now, no problem with me. Ok to merge.

PR #114, opened 2016-11-28
add front end validation for tags

annyhe:
improved error message for invalid input

### DIFF
--- a/view/admin/components/pages/Detail.js
+++ b/view/admin/components/pages/Detail.js
@@ -20,18 +20,28 @@ import Form from '../common/Forms';
 import PageHeader from '../common/PageHeader';
 import Modal from '../common/Modal';
 import ButtonRowWhenRead from '../common/ButtonRowWhenRead';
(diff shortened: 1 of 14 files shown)


annyhe:
I tried creating a review app but ran into problems when connecting to redis. @meynet can you create a review app and share the url? I can push subjects and aspects data to it.
Here is what it should look like on invalid input:
screen shot 2016-11-27 at 7 47 33 pm

view/admin/components/pages/Detail.js
```diff
@@ -59,19 +69,59 @@ class Detail extends React.Component {
...
+  showError(error) {
+    const errorMessage = (error && typeof error === 'object') ?
+      JSON.stringify(error) : error;
+    this.setState({ error: errorMessage });
+  }
+  closeError() {
+    this.setState({ error: '' });
+  }
+  // obj contains invalid fields and their data
+  setIinvalidFieldObj(invalidFieldObj) {
```
iamigo: typo in function name?

PR #120, opened 2016-12-01
Create token api: POST /v1/token

pallavi2209:
Updated Controller logic
Updated Swagger docs
Updated tests

### DIFF
--- a/api/v1/controllers/token.js
+++ b/api/v1/controllers/token.js
@@ -7,52 +7,54 @@
  */
 
 /**
- * api/v1/controllers/apiaccess.js
+ * api/v1/controllers/token.js
  */
-
-const configuredPassport = require('../../../index').passportModule;
 const httpStatus = require('../constants').httpStatus;
 const u = require('../helpers/verbs/utils');
 const apiErrors = require('../apiErrors');
 const jwtUtil = require('../../../utils/jwtUtil');
+const helper = require('../helpers/nouns/tokens');
 
 const resourceName = 'token';
 
 module.exports = {
 
   /**
-   * Authenticates user and sends token in response with status code 200
-   * if authenticated else responds with error.
+   * Authenticates user using provided token and creates new token with given
+   * name. Saves created token to db and sends token in response with status
+   * code 201 if token craeted, else responds with error.
    * @param {IncomingMessage} req - The request object
    * @param {ServerResponse} res - The response object
    * @param {Function} next - The next middleware function in the stack
    *
    */
-
-  // this endpoint will be replaced with new create token endpoint.
   postToken(req, res, next) {
-    configuredPassport.authenticate('local-login', (err, user/* , info */) => {
-      if (err) {
-        return u.handleError(next, err, resourceName);
-      }
-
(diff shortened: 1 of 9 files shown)


api/v1/swagger.yaml
```diff
@@ -4470,11 +4470,16 @@ paths:
...
+      summary: Create a new api access token
       tags: [ token ]
       description: >-
-        Authenticate a user with email and password and create an api access
-        token.
+        Create a new api access token by providing a token in header.
+        If the Refocus configuration parameter `useAccessToken` is set to 
+        `true`, you must include an `Authorization` request header with your 
+        [JSON Web Token](https://tools.ietf.org/html/rfc7519) (JWT) as the 
+        value. You can get a token using `POST /v1/register` or `POST /v1/token`.
```
iamigo: line 4482: POST /v1/token**s**

api/v1/controllers/token.js
```diff
@@ -7,52 +7,54 @@
...
-      return res.status(httpStatus.OK).json({
-        success: true,
-        message: 'Enjoy your token!',
-        token: createdToken,
+    // req.user is set when verifying token with user details. If req.user is
+    // not set, then return error.
+    if (!req.user || !req.user.name) {
+      const tokenErr = new apiErrors.LoginError({
+        explanation: 'Unable to parse token. Please make sure that you have ' +
+         'provided token in header.',
```
iamigo: "... you have provided token in header."

should be

"... you include an Authorization header (i.e. `Authorization=YOUR_TOKEN`) with your POST request."

api/v1/controllers/token.js
```diff
@@ -7,52 +7,54 @@
...
       });
-    })(req, res, next);
+      tokenErr.resource = resourceName;
+      return u.handleError(next, tokenErr, resourceName);
+    }
+
+    // create tokem to be returned in response.
```
iamigo: "tokem" should be "token"

api/v1/swagger.yaml
```diff
@@ -4484,33 +4489,40 @@ paths:
...
                 type: string
                 maxLength: 254
                 description: >
-                  The user's email address.
-              password:
-                type: string
-                format: password
-                description: >
-                  The user's password.
+                  The name of token. Token name should be unique for a user.
```
iamigo: A name for your new token. Token name must be unique for a user. Best practice is to select a name which reminds you of how you intend to use the token, e.g. "PushSamplesFromSystemXyz".

api/v1/swagger.yaml
pallavi2209: @iamigo Actually /v1/tokens does not work. We had /v1/token endpoint until now because we started with only one token create/return.
Now we can change this endpoint to /v1/tokens, but this will touch a lot of files. So, I did not clutter this PR. I am going to create next PR with just /v1/token -> /v1/tokens changes.

iamigo: Got it. OK.

pallavi2209:
@iamigo updated the PR

PR #121, opened 2016-12-01
Make token endpoint plural, v1/token -> /v1/tokens

pallavi2209:
Also, tried to fix flapping ssoconfig test.


pallavi2209:
This PR is based on #120, should be merged after that.

PR #122, opened 2016-12-02
Writeperms db

iamigo:


shriramshankar:
@iamigo just read your spike and the user doc on this. So, this is just to do with phase 1 correct?

iamigo:
@shriramshankar: correct. We'll tackle Phase II (defining groups of users and assigning permissions by group) later.

db/model/subject.js
```diff
@@ -440,9 +446,7 @@ module.exports = function subject(seq, dataTypes) {
 
               return resolve(inst);
             })
-            .catch((err) => {
-              return reject(err);
-            });
+            .catch(reject);
```
iamigo: A promise's `.catch(...)` function takes a function as an argument. The function can be an anonymous inline function like 

```.catch((err) => { reject(err); })```

or it can be a named function like

```.catch(reject)```

annyhe: ok. somehow coveralls check is stalled. Have you seen this problem before @shriramshankar ?

iamigo: There's a coveralls incident going on: https://github.com/lemurheavy/coveralls-public/issues/892

Let's not let ourselves get blocked by the coveralls calculation. Feel free to approve or reject PRs on their merit :)

PR #123, opened 2016-12-02
sample filter test

annyhe:
- with lint fixes
- api: use $in for enums
- add enum values for resources with enums (aspect, sample)

### DIFF
--- a/api/v1/helpers/verbs/findUtils.js
+++ b/api/v1/helpers/verbs/findUtils.js
@@ -13,6 +13,8 @@
 const u = require('./utils');
 const constants = require('../../constants');
 const defaults = require('../../../../config').api.defaults;
+const ZERO = 0;
+const ONE = 1;
 
 /**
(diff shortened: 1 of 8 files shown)


api/v1/helpers/verbs/findUtils.js
```diff
@@ -94,30 +104,49 @@ function toSequelizeWhere(filter, props) {
       const values = [];
 
       /*
+       * If enum filter is enabled and key is an enumerable field
```
iamigo: thx for the clear comments!

package.json
```diff
@@ -13,6 +13,7 @@
     "url": "https://github.com/Salesforce/refocus.git"
   },
   "scripts": {
+    "cleanup": "npm run dropdb && npm run initdb && npm run checkdb",
```
iamigo: thx for adding this--i got rid of stuff a little too aggressively a couple of weeks ago ;)

PR #124, opened 2016-12-02
When an aspect name is changed all related samples will be removed

harshkothari410:


db/model/aspect.js
```diff
@@ -203,7 +204,8 @@ module.exports = function aspect(seq, dataTypes) {
       afterUpdate(inst /* , opts */) {
         if (inst.changed('isPublished') &&
           inst.previous('isPublished') &&
-          !inst.getDataValue('isPublished')) {
+          !inst.getDataValue('isPublished') ||
```
iamigo: need db tests for this change

harshkothari410: I am on it.

harshkothari410: @iamigo test added

shriramshankar:
The branch is named "W-3453246" is this fine?

PR #125, opened 2016-12-02
set order of subjects in perspectives

annyhe:
include minor lint fixes for CreatePerspective


annyhe:
somehow coveralls says coverage went down, when it did not. Restoring commit to see whether it makes a difference.

tests/view/components/createPerspective.js
```diff
@@ -0,0 +1,150 @@
+/**
```
iamigo: OK

PR #130, opened 2016-12-09
show aspect valueType and relatedLinks when embedded.

jgraff2:


db/model/sample.js
```diff
@@ -126,6 +126,8 @@ module.exports = function sample(seq, dataTypes) {
                 'infoRange',
                 'okRange',
                 'valueLabel',
+                "valueType",
```
iamigo: we use single quotes

(`npm run jscs` will check for code style--you can also set up a sublime editor plugin to show you violations in realtime in your editor, if you use sublime)

jgraff2: my mistake, I should have noticed this. I'll run the code style check in the future.

PR #133, opened 2016-12-12
keep name on create perspective

annyhe:
- add create perspective tests


annyhe:
@shriramshankar @iamigo The nsp check doesn't show up on github, but I still get emails about pass/fail. This pr fails with 4 nsp errors, the same number of errors as on master. I emailed the nsp team about this problem.

PR #134, opened 2016-12-13
Add API endpoints for user tokens

iamigo:
GET /v1/users/{key}/tokens
DELETE /v1/users/{key}/tokens/{tokenName}
GET /v1/users/{key}/tokens/{tokenName}
POST /v1/users/{key}/tokens/{tokenName}/revoke
POST /v1/users/{key}/tokens/{tokenName}/restore


shriramshankar:
@iamigo For some reason,  the new line at the end of all the files are removed.  Wondering if we do not want them anymore. I need to make these changes in my PR too if this is the case.

iamigo:
@annyhe waiting for review

tests/api/v1/userTokens/revokeRestore.js
```diff
@@ -0,0 +1,120 @@
...
+    .then(() => done())
+    .catch(done);
+  });
+
+  after(u.forceDelete);
+
+  it('ok', (done) => {
+    api.post(`${path}/${uname}/tokens/${tname}/revoke`)
+    .set('Authorization', '???')
+    .send({})
```
annyhe: why does the test accept authorization: ????

tests/api/v1/userTokens/utils.js
```diff
@@ -0,0 +1,30 @@
...
+const testStartTime = new Date();
+
+module.exports = {
+  fakeUserCredentials: {
+    email: 'user1@abc.com',
+    password: 'fakePasswd',
+    username: 'user1@abc.com'
+  },
+  forceDelete(done) {
+    tu.forceDelete(tu.db.Token, testStartTime)
```
annyhe: this and the following files look familiar: are they from another pull request?

tests/api/v1/userTokens/delete.js
```diff
@@ -0,0 +1,81 @@
...
+    api.delete(`${path}/${uname}/tokens/${tname}`)
+    .set('Authorization', '???')
+    .expect(constants.httpStatus.OK)
+    .end((err, res) => {
+      if (err) {
+        done(err);
+      } else {
+        expect(res.body).to.have.property('name', tname);
+        expect(res.body.isDeleted).to.not.equal(0);
+        done();
```
annyhe: do we expect isDeleted to === 1?

tests/api/v1/userTokens/revokeRestore.js
iamigo: 'cuz we're not actually enforcing that a token is valid yet

tests/api/v1/userTokens/delete.js
iamigo: when a record is soft-deleted, the "isDeleted" field will have a non-zero long int representing the timestamp of when it was soft-deleted

PR #135, opened 2016-12-13
Augment Sample with Subject and Aspect instances to stop samples creep…

shriramshankar:
…ing into the wrong perspective


iamigo:
@annyhe waiting for review

PR #136, opened 2016-12-13
not allowed to delete system-generated token

iamigo:


shriramshankar:
Just for my understanding, how do we know if a token is system generated or not ?

iamigo:
system-generated token has token name = user name

tests/db/model/token/delete.js
```diff
@@ -40,6 +40,28 @@ describe('db: Token: find', () => {
...
+      t = token;
+      done();
+    })
+    .catch(done);
+  });
+  afterEach(u.forceDelete);
+  it('Not allowed to delete system token', (done) => {
+    Token.findOne({ where: { name: t.name } })
+    .then((token) => token.destroy())
+    .then(() => done('expecting error here'))
```
annyhe: Why do we expect an error here? The user hasn't been deleted yet.

iamigo: We don't want to let a user delete the system-generated token for their user record, so the token.destroy call should send back an error

PR #140, opened 2016-12-14
create new token UI

annyhe:


annyhe:
@shriramshankar Comments and feedback are welcome :)

view/tokens/new.pug
```diff
@@ -0,0 +1,33 @@
+//
```
shriramshankar: any reason this has been named new.pug, instead of tokens.pug?

annyhe: There might be other pages ie. tokens/edit.pug, or we can keep it on the same page.
If we use a page for all CRUD activities, the create would need to happen in a modal (instead of going to a different page).
For simplicity, let's stay with separate pages for now. If needed we can always rename the file later.

PR #143, opened 2016-12-19
Checks whether token-enforced path behaves as expected

annyhe:


annyhe:
@iamigo Somehow GET /v1/docs returns a 301 "Moved Permanently". Is this expected behavior?

iamigo:
GET /v1/docs returns 200 today... why would it be different here?

tests/enforceToken/verbs.js
```diff
@@ -0,0 +1,154 @@
...
+const registerPath = '/v1/register';
+const { OK, CREATED, FORBIDDEN } = constants.httpStatus;
+
+const STABLE_PATH = '/v1/subjects';
+const NEW_SUBJECT = 'NEW_SUBJECT';
+
+describe('API verb token enforced tests', () => {
+  it('GET docs without token should succeed', (done) => {
+    api.get('/v1/docs')
+    .expect(constants.httpStatus.OK)
```
annyhe: @iamigo @shriramshankar This is where the test fails: Error: expected 200 "OK", got 301 "Moved Permanently". Going to the page manually returns a 304 Not Modified, since I already loaded the page previously.
To work around it we can check that the error to be not defined. I am still looking into why this happens.

annyhe:
So the redirect was happening due to the missing / at the end: location in header pointed to /v1/docs/. Modifying the url slightly solved the problem.

package.json
```diff
@@ -13,9 +13,9 @@
     "url": "https://github.com/Salesforce/refocus.git"
   },
   "watch": {
-    "test-view": {
-      "patterns": ["view/perspective", "tests/view/components"],
-      "extensions": "js,jsx",
+    "test-enforced": {
```
iamigo: are these "watch" changes part of this story?

webpack.config.js
```diff
@@ -15,16 +15,16 @@ module.exports = {
   devtool: 'eval-source-map',
   entry: [
     'webpack-hot-middleware/client?reload=true',
-    path.join(__dirname, 'view/perspective/app.js'),
+    path.join(__dirname, 'view/admin/app.js'),
```
iamigo: are these webpack changes part of this story?

package.json
annyhe: We don't have to commit those. I'll take it out.

webpack.config.js
annyhe: Ok I'll take out those too.

annyhe:
Updated as per @iamigo's comments. Feedback and suggestions @shriramshankar?

tests/testUtils.js
```diff
@@ -84,5 +96,10 @@ module.exports = {
...
     .catch((err) => done(err));
   }, // forceDeleteUser
-
+  // delete users
+  forceDeleteSubject(done) {
+    forceDelete(db.Subject, testStartTime)
+    .then(() => done())
+    .catch((err) => done(err));
+  }, // forceDeleteUser
 }; // exports
```
shriramshankar: Do we need forceDeleteSubject here?

annyhe: If we use the same subject, its name will be changed in some tests. Hence it is a good idea to create and delete a separate subject for each test.

iamigo:
@annyhe can you confirm that our realtime event streaming works in heroku environment with USE_ACCESS_TOKEN=true?

tests/testUtils.js
shriramshankar: the tests to delete subjects are in utils.js. There is a utils.js file under each of the folders in /tests

annyhe:
@iamigo Confirmed on local and heroku http://refocus-pr-143.herokuapp.com/perspectives/NorthAmerica?debug=REALTIME

tests/testUtils.js
annyhe: Yes there is, and we have delete subjects in tests/api/v1/subjects/utils.js. However it deletes tags, profiles, user, and aspect. The section I added only deletes subjects.

shriramshankar: testUtils.js has a generic method called "forceDelete", this takes in a model as its first argument. All you need to do is export the subject model doing something like u.db.Subject and pass the subject to forceDelete. This will take care of deleting the subject and you do not have to create another method just to delete the subjects.

PR #147, opened 2016-12-20
up down enter on perspective dropdown

annyhe:
- use arrow key up or down to navigate dropdown options
- press enter on a highlighted option, to go that perspective
- this PR does not include right click: open in new tab


iamigo:
Cool!

Wondering why I can't arrow all the way down to the "+ New Perspective" item.

iamigo:
I like that it loops back around to the top if you keep arrowing down past the bottom :)

annyhe:
@iamigo We can change the PR to allow the arrow to go down to the "+ New Perspective" item. It is not a trivial change since the options and the button are separate components, so I'll need to add tests. 
It's best to add a new story for the feature.

annyhe:
Comments and feedback is welcome @shriramshankar

PR #148, opened 2016-12-21
Aspect tag change realtime events

iamigo:
When an aspect’s tags change, send “delete” and “add” realtime events (instead of “update”)
so that perspectives which filter by aspect tag will get the right sample events.

### DIFF
--- a/db/model/aspect.js
+++ b/db/model/aspect.js
@@ -18,6 +18,11 @@ const assoc = {};
 const timeoutLength = 10;
 const timeoutRegex = /^[0-9]{1,9}[SMHDsmhd]$/;
 const valueLabelLength = 10;
+const sampleEventNames = {
+  add: 'refocus.internal.realtime.sample.add',
+  upd: 'refocus.internal.realtime.sample.update',
+  del: 'refocus.internal.realtime.sample.remove',
+};
 
 module.exports = function aspect(seq, dataTypes) {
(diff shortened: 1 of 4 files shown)


tests/db/model/aspect/tags.js
```diff
@@ -0,0 +1,75 @@
...
+    .then((o) => {
+      o.tags = ['T3'];
+      return o.save();
+    })
+    .then(() => Sample.findOne({
+      name: `${tu.namePrefix}Subject|${tu.namePrefix}A`,
+    }))
+    .then((s) => {
+      expect(s.dataValues.aspect.tags).to.have.members(['T3']);
+      done();
```
annyhe: @iamigo Wouldn't we want the test to assert 
expect(aspect.tags).to.deep.equal(['T3'])?

db/model/aspect.js
```diff
@@ -217,6 +252,27 @@ module.exports = function aspect(seq, dataTypes) {
...
+                  .then((s) => {
+                    common.publishChange(s, sampleEventNames.add);
+                  });
+                }
+
+                return seq.Promise.resolve(true);
+              });
+            })
+            .then(() => resolve(inst))
+            .catch(reject);
```
annyhe: @iamigo do we have a test that asserts on unpublishing an aspect, it loses all of its samples?

PR #149, opened 2016-12-22
Move more tests under coveralls coverage

iamigo:
- Get rid of the special-case test config environments: testTokenReq, testTokenNotReq, testTokenEnforced.
- Use the “enforceApiToken” feature toggle instead of a config environment variable
- Move the jobQueue tests and realtime tests under code coverage
- Consolidate the tests which require USE_ACCESS_TOKEN=true into a single directory
- Move the check for the “useWorkerProcess” feature toggle from the module level (only evaluated at startup) to the function level (evaluated at runtime)
- Add feature toggle “enforceApiToken” to list of “longTermToggles”
- Add “toggleOverride” function to testUtils so that tests can override a feature toggle at runtime
- Update the jobQueue and realtime tests to use “toggleOverride” to set and unset the feature before and after running those tests


config.js
```diff
@@ -175,42 +172,6 @@ module.exports = {
       tokenSecret:
        '7265666f637573726f636b7377697468677265656e6f776c7373616e6672616e',
     },
-    testTokenReq: {
```
annyhe: yay less config to maintain :)

iamigo:
@annyhe 

> npm module feature-toggle is missing from package.json.

`"feature-toggles": "^1.4.0",`
is in package.json

PR #150, opened 2016-12-22
record level permissions: added DELETE enpoint

shriramshankar:
added DELETE /v1/[subjects|aspects|perspectives|lenses]/{key}/writers endpoints and made a few changes to DELETE /v1/[subjects|aspects|perspectives|lenses|uesrs]/{key}


shriramshankar:
@iamigo @annyhe Just started working on 
DELETE /v1/[subjects|aspects|perspectives]/{key}/writers/{u_key}  endpoint. Will create another PR for this.

shriramshankar:
Looks like the check to see if we need to send a 403 or not. I will do that, along with the next PR.

PR #153, opened 2016-12-27
Get rid of unnecessary “test” environment

iamigo:
(was only being used for view tests)—just use the “build” env for this

### DIFF
--- a/config.js
+++ b/config.js
@@ -13,7 +13,6 @@
  */
 'use strict'; // eslint-disable-line strict
 require('./config/toggles'); // Loads the feature toggles
-const featureToggles = require('feature-toggles');
 const configUtil = require('./config/configUtil');
 const defaultPort = 3000;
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -13,7 +13,6 @@
  */
 'use strict'; // eslint-disable-line strict
 require('./config/toggles'); // Loads the feature toggles
```
annyhe: @iamigo Do we need this line? Seems like we are not using toggles in this file.

iamigo: Yeah I saw that too--I meant to take it out in this PR... I'll catch in the next one.

iamigo: Actually having this line here makes sure that the feature flags are loaded *before* the config is loaded.

annyhe: Do we need the feature flags in this module? Where are we using it?

PR #154, opened 2016-12-27
Fix the groupings (token/tokens/users) in the swagger-generated API docs

iamigo:
- Group all the /tokens operations together
- Include all the /users/.../tokens operations in both the "tokens" group and the "users" group
- Include all the /writers operations under a "writers" group as well as in the aspect/subject/perspective/lens group


annyhe:
Tokens section looks good
screen shot 2016-12-27 at 2 55 22 pm

PR #156, opened 2016-12-28
subject tag modifications, with tests

annyhe:
- POST, PUT and PATCH subjects with duplicate tags fails
- case insensitive tag comparison
- implement GET EXCLUDES /subjects with multiple tag, and with single tag

### DIFF
--- a/api/v1/apiErrors.js
+++ b/api/v1/apiErrors.js
@@ -52,6 +52,15 @@ apiErrors.create({
     'previously revoked, and vice versa.',
 });
 
+apiErrors.create({
+  code: 11103,
+  status: 400,
+  name: 'TagValidationError',
+  parent: apiErrors.ValidationError,
+  fields: ['tags'],
+  defaultMessage: 'Tags are case-insensitive. Duplicates found',
+});
+
 // ----------------------------------------------------------------------------
 // Not Found
 // ----------------------------------------------------------------------------
--- a/api/v1/apiErrors.js
+++ b/api/v1/apiErrors.js
@@ -55,7 +55,7 @@ apiErrors.create({
 apiErrors.create({
   code: 11103,
   status: 400,
-  name: 'TagValidationError',
+  name: 'DuplicateFieldError',
   parent: apiErrors.ValidationError,
   fields: ['tags'],
   defaultMessage: 'Tags are case-insensitive. Duplicates found',
--- a/api/v1/controllers/subjects.js
+++ b/api/v1/controllers/subjects.js
@@ -51,12 +51,21 @@ function checkDuplicates(tagsArr) {
 }
 
 /**
- * Checks the given fields. If fail, throws an
- * corresponding error
- * @param {String} absolutePath Should not exist
- * @param {Array} tags Should not contain duplicates
+ * Validates the given fields from request body or url.
+ * If fails, throws a corresponding error.
+ * @param {Object} requestBody Fields from request body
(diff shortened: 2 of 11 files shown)


annyhe:
@iamigo Regarding duplicate parameters, different sites returns two main results
1. Only query the unique param value
ie. https://finance.yahoo.com/quote/TWTR/financials?p=TWTR,TWTR,TWTR **renders the same** as https://finance.yahoo.com/quote/TWTR/financials?p=TWTR
2. Send back 200 NO DATA
MSN money http://www.msn.com/en-us/money/stockscreener/52weeklows?market=USA,USA,USA
gets a 200 'DATA NOT AVAILABLE' response, whereas http://www.msn.com/en-us/money/stockscreener/52weeklows?market=USA renders correctly
The SF library also does not render correctly on duplicate params: http://sfpl.org/index.php?pg=2000565501,2000565501 returns nothing while
http://sfpl.org/index.php?pg=2000565501 returns the expected results.

The implementation in the pull request acts the same as option 1. We can discuss whether we want to keep it the same, or return nothing.

api/v1/controllers/subjects.js
```diff
@@ -23,6 +23,49 @@ const u = require('../helpers/verbs/utils');
...
+  return false;
+}
+
+/**
+ * Checks the given fields. If fail, throws an
+ * corresponding error
+ * @param {String} absolutePath Should not exist
+ * @param {Array} tags Should not contain duplicates
+ */
+function checkRequestBody(absolutePath, tags) {
```
iamigo: maybe call this function "validate" and pass in the request body itself?

api/v1/swagger.yaml
```diff
@@ -3500,9 +3500,15 @@ paths:
...
             pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,59}$
           description: >-
-            Filter by tags.
+             Comma-separated list of tags to include/exclude.
+             For example, ?tags=FOO,BAR will only return subjects with tags
+             FOO or BAR. Prefix each of the tag name with a negative sign to
+             indicate that a subject with that tag should be excluded.
+             For example, ?tags=-BAZ,-FOO will return only the subjects
+             with tag name not equal to BAZ or FOO. Subjects without tags
+             are not included in the result set
```
iamigo: Maybe add a line like "Remember: tag names are case-insensitive!"?

tests/api/v1/subjects/post.js
```diff
@@ -454,19 +455,35 @@ describe(`api: POST ${path}`, () => {
         done();
       });
     });
-    it('posting subject with duplicate tags', (done) => {
+
+    it('posting subject with case sensitive tags should fail',
```
iamigo: 'posting subject with case sensitive (duplicate) tags should fail'

api/v1/apiErrors.js
```diff
@@ -52,6 +52,15 @@ apiErrors.create({
...
 });
 
+apiErrors.create({
+  code: 11103,
+  status: 400,
+  name: 'DuplicateFieldError',
+  parent: apiErrors.ValidationError,
+  fields: ['tags'],
+  defaultMessage: 'Tags are case-insensitive. Duplicates found',
+});
```
annyhe: Changing to throw duplicate error, similar to how Salesforce API.

annyhe: @iamigo Ready for another lookover.

PR #157, opened 2016-12-28
User model's defaultScope should exclude password

iamigo:
By default, the API will now return User records *without* the encrypted password field.
Explicitly use the db model scope “withSensitiveInfo” to include the encrypted password field in the response (e.g. for passport configuration).
Add tests to confirm both the default and the scoped behavior.
Unskip all the “getWriters” tests.


iamigo:
(@shriramshankar review optional)

PR #180, opened 2016-12-30
update clock process

iamigo:
Refactor the clock process to make it cleaner to define scheduled jobs. Add some clock tests.
Make the sample timeout scheduled job enqueue work if worker dyno is enabled. 
Add new job type "SAMPLE_TIMEOUT" and execute the sample timeout job from the jobProcessor.


worker/jobProcessor.js
```diff
@@ -9,20 +9,34 @@
...
 jobQueue.process(jobType.BULKUPSERTSAMPLES, (job, done) => {
   const samples = job.data;
-  console.log('Job id being processed: ' + // eslint-disable-line no-console
-                      job.id);
-  helper.model.bulkUpsertByName(samples).then(() => done());
+  const msg = `Processing ${jobType.BULKUPSERTSAMPLES} job ${job.id} ` +
+    `with ${samples.length} samples`;
+  console.log(msg); // eslint-disable-line no-console
+  helper.model.bulkUpsertByName(samples)
+  .then(() => done());
```
annyhe: @iamigo Can this throw an error? If so, it's best to have a .catch here

iamigo:
@annyhe thx for reviewing, but please don't merge this--I want to have Shriram take a look at it

clock/scheduledJobs/sampleTimeoutJob.js
```diff
@@ -0,0 +1,30 @@
...
+ * Executes the sample timeout process. If worker process is enabled, enqueues
+ * a job, otherwise just executes work directly in this process.
+ */
+const featureToggles = require('feature-toggles');
+
+module.exports = {
+  execute() {
+    if (featureToggles.isFeatureEnabled('useWorkerProcess')) {
+      const jobWrapper = require('../../jobQueue/jobWrapper');
+      const jobType = require('../../jobQueue/setup');
```
shriramshankar: I think this needs to be 
const jobType = require('../../jobQueue/setup').jobType
and  we are getting the undefined error because jobType.SAMPLE_TIMEOUT returns undefined for now.

iamigo: Correct. Will update shortly.

shriramshankar:
@iamigo  It seems to be working after this change but for some reason the clock seems to be queuing a bunch of jobs to the workers every second.

iamigo:
I'll take a look tomorrow--thx @shriramshankar

PR #197, opened 2017-01-11
Check whether the user is an admin user or the owner of a token befor…

iamigo:
…e allowing the user to delete a token.

Check whether the user is an admin user before allowing the user to revoke/restore a token.

### DIFF
--- a/api/v1/controllers/globalconfig.js
+++ b/api/v1/controllers/globalconfig.js
@@ -12,34 +12,13 @@
 'use strict';
 
 const helper = require('../helpers/nouns/globalconfig');
-const Profile = require('../helpers/nouns/profiles').model;
-const User = require('../helpers/nouns/users').model;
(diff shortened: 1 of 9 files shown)


api/v1/controllers/globalconfig.js
```diff
@@ -12,34 +12,13 @@
...
-const User = require('../helpers/nouns/users').model;
 const doDelete = require('../helpers/verbs/doDelete');
 const doPatch = require('../helpers/verbs/doPatch');
 const doPost = require('../helpers/verbs/doPost');
 const doGet = require('../helpers/verbs/doGet');
 const doFind = require('../helpers/verbs/doFind');
 const u = require('../helpers/verbs/utils');
-const jwtUtil = require('../../../utils/jwtUtil');
-
-function isAdmin(req) {
```
iamigo: Extracting this out to a utility module so it can be reused.

PR #198, opened 2017-01-11
Cache /v1/perspectives

pallavi2209:


api/v1/helpers/verbs/doFind.js
```diff
@@ -15,6 +15,8 @@ const u = require('./utils');
 const fu = require('./findUtils');
 const COUNT_HEADER_NAME = require('../../constants').COUNT_HEADER_NAME;
 const httpStatus = require('../../constants').httpStatus;
+const redisCache = require('../../../../cache/redisCache').client;
+const SECS_IN_MIN = 60;
```
iamigo: Can we pull this out to config?

api/v1/helpers/verbs/doFind.js
```diff
@@ -32,19 +34,21 @@ const httpStatus = require('../../constants').httpStatus;
...
-    reqResNext.res.status(httpStatus.OK).json(retval);
-  })
-  .catch((err) => u.handleError(reqResNext.next, err, props.modelName));
+        return u.responsify(row, props, reqResNext.req.method);
+      });
+      resolve(retval);
+    })
+    .catch((err) => reject(err));
+  });
 } // doFindAndCountAll
```
shriramshankar: Why do you need this to be wrapped by a Promise?

package.json
```diff
@@ -38,9 +38,10 @@
     "test-api": "mocha -R dot --recursive tests/api",
     "test-disablehttp": "DISABLE_HTTP=true mocha -R dot --recursive tests/disableHttp",
     "test-enforced": "USE_ACCESS_TOKEN=true mocha -R dot --recursive tests/enforceToken",
+    "test-cache-persp": "ENABLE_CACHE_PERSPECTIVE=true mocha -R dot --recursive tests/enableCache",
```
iamigo: /tests/testUtils.js has a function called "toggleOverride" -- since this feature toggle isn't used to instantiate the app server in index.js, you should be able to set and unset it directly in your test module, see example in tests/jobQueue/v1/bulkUpsert.js. That way you don't have to set up a separate package.json script for it and the tests can be counted by coveralls.

api/v1/helpers/verbs/doFind.js
pallavi2209: Because I wanted to get the retval back in the calling function doFind. Do you think there is better way?

package.json
pallavi2209: @iamigo I tried as you suggested, but I have created a var cacheEnabled using feature toggle in api/v1/helpers/nouns/perspectives.js which is called when app server starts, and hence cacheEnabled calculates to false when checking for feature. This results in failed tests even if I use toggleOverride.

api/v1/helpers/verbs/doFind.js
shriramshankar: Not sure if there is a better way but what about returning u.getScopedModel? This would return a promise and you can do a "then" in the caller to get the value.

package.json
iamigo: Oh, I see.

OK...

Since the perspectives controller uses the api/v1/helpers/nouns/perspectives.js helper and passes it around to the various utility functions, what about setting the value of the helper's cacheEnabled attribute in each of the perspective controller functions which needs it, e.g.

```
  getPerspective(req, res, next) {
    helper.cacheEnabled = featureToggles.isFeatureEnabled('enableCachePerspective');
    doGet(req, res, next, helper);
  },
```

pallavi2209: Yes! That should work. Thanks!

api/v1/helpers/verbs/doFind.js
pallavi2209: Agreed, that should work. No need to wrap in new promise. Updates coming up!

pallavi2209:
@shriramshankar @iamigo updated the PR

iamigo:
@pallavi2209 we currently have the flag enabled in production, so after we merge, please test this in staging with the flag on and off

PR #199, opened 2017-01-13
return the model instance instead of boolean while using req.user to …

shriramshankar:
…check for permission


shriramshankar:
Fix for a bug on the isWritable method while patching an aspect.

PR #201, opened 2017-01-16
Worker activity logging

pallavi2209:


pallavi2209:
Not sure why it says coverage decreased. The tests/logging included in coveralls should cover the logging code lines.

pallavi2209:
Okay, coverage remains same now.

api/v1/controllers/samples.js
```diff
@@ -23,6 +23,45 @@ const doPut = require('../helpers/verbs/doPut');
...
+    if (username) {
+      workerLogObj.user = username;
+    }
+  })
+  .then(() => jwtUtil.getTokenNameFromToken(reqNext.req))
+  .then((tokenName) => {
+    workerLogObj.token = tokenName;
+  })
+  .then(() => {
+    workerLogObj.ipAddress = u.getIPAddrFromReq(reqNext.req);
```
iamigo: seems like this line 50 can be inside of the previous "then" block since line 47 doesn't return a promise which this would depend on

api/v1/helpers/verbs/utils.js
```diff
@@ -678,6 +678,26 @@ function forbidden(next, modelName) {
...
+ * Get IP Address from request.
+ * @param  {Object} req - Request object
+ * @returns {String} - IP Address
+ */
+function getIPAddrFromReq(req) {
+  let ipAddr = null;
+  if (req) {
+    if (req.connection && req.connection.remoteAddress) {
+      ipAddr = req.connection.remoteAddress;
+    }
```
iamigo: Was the intention here for the action taken on line 690 to get overwritten by line 694? i.e. if you get to line 690, should you just return req.connection.remoteAddress, regardless of whether there is an x-forwarded-for header?

utils/activityLogs/workerLog.js
```diff
@@ -0,0 +1,52 @@
...
+  this.ipAddress = null;
+  this.totalTime = null;
+  this.queueTime = null;
+  this.workTime = null;
+  this.dbTime = null;
+  this.recordCount = null;
+  this.errorCount = null;
+
+  // Function to convert worker log object to String format
+  this.toLogString = function toLogString() {
```
iamigo: Any way to add tests which check the output?

pallavi2209: I thought about it. Basically, I would have to export the object created in api/v1/controllers/samples.js. Was not sure if that is a good code practice.

pallavi2209: Let me see if we can do something better.

shriramshankar:
Looks like you missed the logging for the clock process.

api/v1/controllers/samples.js
```diff
@@ -23,6 +23,45 @@ const doPut = require('../helpers/verbs/doPut');
...
+      workerLogObj.queueTime = result.queueTime;
+      workerLogObj.workTime = result.workTime;
+      workerLogObj.dbTime = result.dbTime;
+      workerLogObj.recordCount = result.recordCount;
+      workerLogObj.errorCount = result.errorCount;
+      workerLogObj.toLogString();
+    });
+  })
+  .catch((err) => u.handleError(reqNext.next, err, helper.modelName));
+}
```
shriramshankar: Every time we move a feature to the worker  we will have to copy the this function and put it in its controller.  We already have an job.onComplete lister in jobQueue/jobWrapper.js which acts like a single point of exit for all the jobs. Wondering if we could do something like this there? We could also consider this making it a bit more configurable. We could have an config file were we define the parameters for each of the job type we want to measure. Just after a the job completes, we could use the config to get the parameters from the config for the job type and log them.

utils/activityLogs/workerLog.js
pallavi2209: @iamigo Tried a couple of things, but then after discussion with @shriramshankar, we realized we cannot test the output because the job cannot reach to completion in test mode because there is no worker process.

api/v1/controllers/samples.js
iamigo: @shriramshankar I like this idea. @pallavi2209 I'll leave it up to you to either add this in this story or create a new technical debt story to do this refactoring.

shriramshankar: Ok to merge if we are not logging the clock process and doing the refactoring as a part of this story.

pallavi2209: Working on refactoring in this PR. Will create another PR for clock process.

pallavi2209:
@shriramshankar @iamigo updated the PR

config/activityLog.js
```diff
@@ -0,0 +1,30 @@
+/**
```
iamigo: Yes!!

jobQueue/jobWrapper.js
```diff
@@ -15,6 +15,10 @@
  */
```
iamigo: Maybe export "mapJobResultsToLogObject" and add some tests just sending it mock objects?

jobQueue/jobWrapper.js
```diff
@@ -25,43 +29,127 @@ const jobQueue = require('./setup').jobQueue;
...
     });
   }
 } // removeJobOnComplete
 
 /**
+ * Logs worker activity and remove job when complete
+ * @param  {Object} job - Worker job and job type
+ * @param  {Object} req - Request object
+ */
+function logAndRemoveJobOnComplete(req, job) {
```
iamigo: maybe also add tests for this with a mock request and a mock job?

utils/jwtUtil.js
```diff
@@ -122,8 +122,32 @@ function createToken(tokenName, userName) {
...
+      resolve(null);
+    }
+  });
+} // getTokenNameFromToken
+
 module.exports = {
   verifyToken,
   createToken,
   getUsernameFromToken,
+  getTokenNameFromToken,
```
iamigo: I think we should just have a single function which returns the user name AND token name... I don't think we need to have separate ones.

db/model/sample.js
```diff
@@ -225,7 +225,9 @@ module.exports = function sample(seq, dataTypes) {
           .then((o) => resolve(o))
           .catch((err) => {
             if (isBulk) {
-              resolve(err);
+              // adding isFailed:true to differentiate failed results from
+              // success results in bulk upsert
+              resolve({ explanation: err, isFailed: true });
```
shriramshankar: /*
*/ 
for multi line comments in site a function.

jobQueue/jobWrapper.js
```diff
@@ -25,43 +29,127 @@ const jobQueue = require('./setup').jobQueue;
...
       }, delayToRemoveJobs);
+
+      // if enableWorkerActivityLogs is enabled, update logObject
+      if (featureToggles.isFeatureEnabled('enableWorkerActivityLogs') &&
+       jobResultObj && logObject) {
+        mapJobResultsToLogObject(jobResultObj, logObject);
+
+        // The second argument should match the activity type in
+        // /config/activityLog.js
+        activityLogUtil.printActivityLogString(logObject, 'worker');
```
shriramshankar: /*
*/
for multi line comments in site a function.

utils/jwtUtil.js
```diff
@@ -94,8 +94,12 @@ function getUsernameFromToken(req) {
           return reject(err);
         }
 
-        const username = decodedData.username;
-        return resolve(username);
+        const resObj = {
+          username: decodedData.username,
+          tokenname: decodedData.tokenname,
+          timestamp: decodedData.timestamp,
```
iamigo: we *encode* the timestamp so it generates a new token string for the given username and tokenname, but I don't think we should *return* the timestamp, just the username and tokenname

pallavi2209:
@shriramshankar @iamigo updated the PR

pallavi2209:

pallavi2209:
Conflict emerged after latest commit. Resolved and force pushed again.

utils/jwtUtil.js
annyhe: /**
 * Copyright (c) 2017, salesforce.com, inc.
 * All rights reserved.
 * Licensed under the BSD 3-Clause license.
 * For full license text, see LICENSE.txt file in the repo root or
 * https://opensource.org/licenses/BSD-3-Clause
 */

/**
 * tests/logging/enableWorkerLog.js
 */

const api = supertest(require('../../index').app);
const tu = require('../testUtils');
const u = require('./utils');
const constants = require('../../api/v1/constants');
const Aspect = tu.db.Aspect;
const path = '/v1/samples/upsert/bulk';
const jobQueue = require('../../jobQueue/setup').jobQueue;

// replicated api bulk upsert basic tests with enableWorkerActivityLogs enabled
// to execute worker activity logging and test that it does not interfere with
// the existing code.
describe('api: POST ' + path, () => {
  let token;

  before((done) => {
    jobQueue.testMode.enter();
    tu.toggleOverride('useWorkerProcess', true);
    tu.toggleOverride('enableWorkerActivityLogs', true);
    tu.createToken()
    .then((returnedToken) => {
      token = returnedToken;
      done();
    })
    .catch(done);
  });

  before((done) => {
    Aspect.create({
      isPublished: true,
      name: `${tu.namePrefix}Aspect1`,
      timeout: '30s',
      valueType: 'NUMERIC',
      criticalRange: [0, 1],
    })
    .then(() => done())
    .catch(done);
  });

  after(u.forceDelete);
  after(tu.forceDeleteUser);
  after(() => {
    tu.toggleOverride('useWorkerProcess', false);
    tu.toggleOverride('enableWorkerActivityLogs', false);
    jobQueue.testMode.clear();
  });

  it('all succeed', (done) => {
    api.post(path)
    .set('Authorization', token)
    .send([
      {
        name: `${tu.namePrefix}Subject|${tu.namePrefix}Aspect1`,
        value: '2',
      }, {
        name: `${tu.namePrefix}Subject|${tu.namePrefix}Aspect2`,
        value: '4',
      },
    ])
    .end((err /* , res */) => {
      if (err) {
        done(err);
      }

      done();
    });
  });

  it('some succeed, some fail returns ok', (done) => {
    api.post(path)
    .set('Authorization', token)
    .send([
      {
        name: `${tu.namePrefix}NOT_EXIST|${tu.namePrefix}Aspect1`,
        value: '2',
      }, {
        name: `${tu.namePrefix}Subject|${tu.namePrefix}Aspect2`,
        value: '4',
      },
    ])
    .end((err /* , res */) => {
      if (err) {
        done(err);
      }

      done();
    });
  });

  it('all fail returns ok', (done) => {
    api.post(path)
    .set('Authorization', token)
    .send([
      {
        name: `${tu.namePrefix}NOT_EXIST|${tu.namePrefix}Aspect1`,
        value: '2',
      }, {
        name: `${tu.namePrefix}Subject|${tu.namePrefix}NOT_EXIST`,
        value: '4',
      },
    ])
    .end((err /* , res */) => {
      if (err) {
        done(err);
      }

      done();
    });
  });
});

tests/logging/enableWorkerLog.js
```diff
@@ -0,0 +1,126 @@
...
+        value: '2',
+      }, {
+        name: `${tu.namePrefix}Subject|${tu.namePrefix}Aspect2`,
+        value: '4',
+      },
+    ])
+    .expect(constants.httpStatus.OK)
+    .end((err /* , res */) => {
+      if (err) {
+        done(err);
```
annyhe: @pallavi2209 It seems we are not checking for results here. Is the purpose of the test to check for OK httpStatus with toggle turned on?

jobQueue/jobWrapper.js
```diff
@@ -25,47 +28,127 @@ const jobQueue = require('./setup').jobQueue;
...
 } // removeJobOnComplete
 
 /**
+ * Logs worker activity and remove job when complete
+ * @param  {Object} job - Worker job
+ * @param  {Object} req - Request object
+ */
+function logAndRemoveJobOnComplete(req, job) {
+  // if activity logs are enabled, log activity and remove job
+  if (featureToggles.isFeatureEnabled('enableWorkerActivityLogs')) {
```
annyhe: @pallavi2209 This function can be modified to be used with different activity logs ie. API, realtime, ... Not just worker activity. Once this pull request gets merged I might be able to modify it slightly for reuse.

annyhe:
Multiple changes in this pull request affect the API logging. I changed to work off of this branch, and will do a pull request once this gets merged.

tests/db/model/sample/upsertBulk.js
```diff
@@ -104,7 +104,7 @@ describe('db: sample: upsertBulk: ', () => {
       let successObj;
       for (let i = 0; i < o.length; i++) {
         const s = o[i];
-        if (util.isError(s)) {
+        if (util.isError(s.explanation)) {
           errorCount++;
```
annyhe: @pallavi2209 are we only checking for errorCount in sample/upsertBulk.js?

tests/logging/jobWrapperLogFuncs.js
```diff
@@ -0,0 +1,98 @@
...
+    tu.createToken()
+    .then((returnedToken) => {
+      token = returnedToken;
+      done();
+    })
+    .catch(done);
+  });
+
+  before((done) => {
+    Aspect.create({
```
annyhe: what is the aspect created for?

pallavi2209: Good, we don't need it. I created this while playing with the test and forgot to remove. Let me remove it in my next PR for clock logs.

tests/db/model/sample/upsertBulk.js
pallavi2209: Looks like we are checking for success objs as well and there are only 2 objs in request.

tests/logging/enableWorkerLog.js
pallavi2209: Yes, and the result is always status:OK.

utils/jwtUtil.js
```diff
@@ -80,12 +80,12 @@ function verifyToken(req, cb) {
...
- * Verify jwt token.
+ * Get token details: username, token name from Token.
  * @param  {object}   req - request object
  * @param  {Function} cb - callback function
  * @returns {User}
  */
-function getUsernameFromToken(req) {
+function getTokenDetailsFromToken(req) {
   return new Promise((resolve, reject) => {
     if (req && req.headers && req.headers.authorization) {
```
annyhe: For the function getTokenDetailsFromToken I added tests in the forthcoming pull request. FYI

pallavi2209: Nice, and this function gets called in 'logAndRemoveJobOnComplete function ok' as well as an integration test.

pallavi2209:
@iamigo @shriramshankar Are we waiting on something to merge this?

PR #203, opened 2017-01-19
change to handle job objects with the old shape

shriramshankar:


shriramshankar:
@pallavi2209 Thanks for the help on this.

shriramshankar:
@iamigo @pallavi2209 Changes to fix the worker dyno error.

PR #204, opened 2017-01-19
add API logs to API endpoints

annyhe:
for testing on heroku, and feedback.

### DIFF
--- a/api/v1/helpers/verbs/doDelete.js
+++ b/api/v1/helpers/verbs/doDelete.js
@@ -14,7 +14,7 @@
 const u = require('./utils');
 const httpStatus = require('../../constants').httpStatus;
 const featureToggles = require('feature-toggles');
-const logAPI = require('../../../../utils/loggingUtil').logAPI;
+const logAuditAPI = require('../../../../utils/loggingUtil').logAuditAPI;
 
 /**
  * Deletes a record and sends the deleted record back in the json response
@@ -27,15 +27,18 @@ const logAPI = require('../../../../utils/loggingUtil').logAPI;
  *  resource type to delete.
  */
 function doDelete(req, res, next, props) {
+  const resultObj = { reqStartTime: new Date() };
   u.findByKey(props, req.swagger.params)
   .then((o) => u.isWritable(req, o,
       featureToggles.isFeatureEnabled('enforceWritePermission')))
   .then((o) => o.destroy())
   .then((o) => {
     if (props.loggingEnabled) {
-      logAPI(req, props.modelName, o);
+      logAuditAPI(req, props.modelName, o);
     }
 
+    resultObj.dbTime = new Date() - resultObj.reqStartTime;
+    resultObj.recordCount = 1;
     const assocNames = [];
 
     /**
@@ -50,6 +53,8 @@ function doDelete(req, res, next, props) {
 
     // when a resource is deleted, delete all its associations too
     u.deleteAllAssociations(o, assocNames);
+    resultObj.retval = o.dataValues;
+    u.logAPI(req, resultObj);
     return res.status(httpStatus.OK).json(u.responsify(o, props, req.method));
   })
   .catch((err) => u.handleError(next, err, props.modelName));
--- a/api/v1/helpers/verbs/utils.js
+++ b/api/v1/helpers/verbs/utils.js
@@ -16,6 +16,7 @@ const apiErrors = require('../../apiErrors');
 const constants = require('../../constants');
 const commonDbUtil = require('../../../../db/helpers/common');
 const jwtUtil = require('../../../../utils/jwtUtil');
+const logAPI = require('../../../../utils/apiLog').logAPI;
 
 /**
  * This function adds the association scope name to the as the to all
@@ -680,7 +681,7 @@ function forbidden(next, modelName) {
 // ----------------------------------------------------------------------------
 
 module.exports = {
-
+  logAPI,
   buildFieldList,
 
   /**
--- a/config/toggles.js
+++ b/config/toggles.js
@@ -73,6 +73,10 @@ const longTermToggles = {
   enableRealtimeActivityLogs:
     environmentVariableTrue(pe, 'ENABLE_REALTIME_ACTIVITY_LOGS'),
 
+  // Enable api activity logging
+  enableApiActivityLogs: environmentVariableTrue(pe,
+                                        'ENABLE_API_ACTIVITY_LOGS'),
+
   // Enable worker activity logging
   enableWorkerActivityLogs:
     environmentVariableTrue(pe, 'ENABLE_WORKER_ACTIVITY_LOGS'),
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "start-clock": "node clock/index.js",
     "start-worker": "node worker/jobProcessor.js",
     "test-api": "mocha -R dot --recursive tests/api",
+    "test-api-log": "ENABLE_API_ACTIVITY_LOGS=true mocha -R dot --recursive tests/api",
     "test-disablehttp": "DISABLE_HTTP=true mocha -R dot --recursive tests/disableHttp",
     "test-enforced": "USE_ACCESS_TOKEN=true mocha -R dot --recursive tests/enforceToken",
     "test-db": "npm run checkdb && mocha -R dot --recursive tests/db",
--- a/tests/logging/jwtUtil.js
+++ b/tests/logging/jwtUtil.js
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/logging/jwtUtil.js
+ */
+'use strict'; // eslint-disable-line strict
+
+const sinon = require('sinon');
+const DUMMY_STR = 'COOL_NAME';
+const expect = require('chai').expect;
+const jwtUtil = require('../../utils/jwtUtil');
+const tu = require('../testUtils');
+
+describe('jwtUtil getTokenDetailsFromToken test', (done) => {
+  // setup
+  let token;
+  const FORBIDDEN = 403;
+
+  before((done) => {
+    tu.createToken()
+    .then((returnedToken) => {
+      token = returnedToken;
+      done();
+    })
+    .catch(done);
+  });
+
+  after(tu.forceDeleteUser);
+
+  it('works with valid Authorization from cookie', () => {
+    const cookies = {
+      Authorization: token,
+    };
+
+    jwtUtil.getTokenDetailsFromToken(cookies).then((res) => {
+      expect(res.tokenname).to.equal(tu.userName);
+      expect(res.usernname).to.equal(tu.userName);
+      done();
+    })
+    .catch(done);
+  });
+
+  it('works with valid Authorization from request header', () => {
+    const req = {
+      headers: {
+        authorization: token,
+      }
+    };
+
+    jwtUtil.getTokenDetailsFromToken(req).then((res) => {
+      expect(res.tokenname).to.equal(tu.userName);
+      expect(res.usernname).to.equal(tu.userName);
+      done();
+    })
+    .catch(done);
+  });
+
+  // if a timeout occurs here, print out error
+  it('fails with invalid format', (done) => {
+    const req = { authorization: DUMMY_STR };
+    jwtUtil.getTokenDetailsFromToken(req).then((res) =>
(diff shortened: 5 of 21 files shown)


annyhe:
@pallavi2209 @iamigo This is the foundation for the API logs: additions to many files will look similar to api/v1/helpers/verbs/doFind.js. Please leave feedback.

utils/apiLog.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
pallavi2209: 2017

utils/apiLog.js
```diff
@@ -0,0 +1,84 @@
...
+      method: req.method,
+    };
+    /* extract user, token and ipaddress and update log object */
+    jwtUtil.getTokenDetailsFromToken(req)
+    .then((resObj) => {
+      logObject.user = resObj.username;
+      logObject.token = resObj.tokenname;
+
+      // in-place modification of logObject
+      mapApiResultsToLogObject(resultObj, logObject);
```
pallavi2209: I think we should create logs even if token is not provided. In that case, this code should be outside then.

utils/apiLog.js
```diff
@@ -0,0 +1,84 @@
...
+function mapApiResultsToLogObject(resultObj, logObject) {
+  if (resultObj.reqStartTime) {
+    logObject.totalTime = `${Date.now() - resultObj.reqStartTime}ms`;
+  }
+
+  if (resultObj.dbTime) {
+    logObject.dbTime = `${resultObj.dbTime}ms`;
+  }
+
+  if (resultObj.recordCount) {
```
pallavi2209: Would be great to have some comments on what each parameter interprets.

tests/logging/jwtUtil.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/logging/jwtUtil.js
```
pallavi2209: Do you think we should have this file in an independent utils folder instead of logging folder?

tests/logging/jwtUtil.js
```diff
@@ -0,0 +1,84 @@
...
+  after(tu.forceDeleteUser);
+
+  it('works with valid Authorization from cookie', () => {
+    const cookies = {
+      Authorization: token,
+    };
+
+    jwtUtil.getTokenDetailsFromToken(cookies).then((res) => {
+      expect(res.tokenname).to.be.defined;
+      expect(res.usernname).to.be.defined;
```
pallavi2209: Can we test the values as well?

tests/logging/enableApiLog.js
```diff
@@ -0,0 +1,121 @@
...
+const tu = require('../testUtils');
+const u = require('./utils');
+const constants = require('../../api/v1/constants');
+const Aspect = tu.db.Aspect;
+const path = '/v1/samples/upsert/bulk';
+
+/* replicated api bulk upsert basic tests with enableApiLogs enabled
+ to execute api activity logging and test that it does not interfere with
+ the existing code. */
+describe('api: POST ' + path, () => {
```
pallavi2209: Do you think it will be better to have tests for all method types?

utils/apiLog.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * /utils/apiLog.js
```
pallavi2209: There is a file activityLog in utils. Do you think it is better to move apiLog to activityLog because api request is a one of the activity type (api, worker and realtime)? Feel free to modify function names if conflicting.

api/v1/helpers/verbs/utils.js
```diff
@@ -680,7 +681,7 @@ function forbidden(next, modelName) {
 // ----------------------------------------------------------------------------
 
 module.exports = {
-
+  logAPI,
```
pallavi2209: Not sure if this a good practice. Files not requiring logAPI will be importing the logAPI code for no reason.

utils/apiLog.js
annyhe: let's keep them separate for now: activityLog has generic functions for all loggers, while apiLog has specific API logging functions.

utils/apiLog.js
annyhe: if no token is provided, wouldn't the request err? In that case nothing would be logged, and an error would be thrown instead (403). 
Which endpoints would be token free? The /login and maybe /register?

tests/logging/jwtUtil.js
annyhe: We can move it to another folder at a later time. Currently there isn't any reason to move it.

tests/logging/enableApiLog.js
annyhe: Good point, I will add more tests for other endpoints, as API logging will cover not only POST, but PUT, PATCH, DELETE.

api/v1/helpers/verbs/utils.js
annyhe: only a few files would be free from logAPI: most files in api/v1/helpers/verbs with the name do*.js would include logAPI.

tests/logging/enableApiLog.js
annyhe: Actually there might be a better way of duplicating tests: I'll run the api tests with the enableAPILogs token set to true. This way we check for the same results, and the tests should still pass.

package.json
```diff
@@ -36,6 +36,7 @@
     "start-clock": "node --max_old_space_size=1500 --nouse-idle-notification clock/index.js",
     "start-worker": "node --max_old_space_size=1500 --nouse-idle-notification worker/jobProcessor.js",
     "test-api": "mocha -R dot --recursive tests/api",
+    "test-api-log": "ENABLE_API_LOGS=true mocha -R dot --recursive tests/api",
```
annyhe: testing the API logger by running the API tests with the config variable on. If any tests fail here while passing without  the API logger, we know there's something wrong with the logger implementation.

utils/apiLog.js
pallavi2209: So, if the tokens are disabled at some point, like now, we wont be able to get an idea of other parameters as well, like dbTime etc. I think logging and token use should be independent. If token is not provided, we can just say token=None.

utils/apiLog.js
```diff
@@ -0,0 +1,107 @@
...
+    // create api activity log object
+    const logObject = {
+      ipAddress: activityLogUtil.getIPAddrFromReq(req),
+      responseBytes: getSize(req.body),
+      uri: req.url,
+      method: req.method,
+    };
+
+    if (!featureToggles.isFeatureEnabled('enforceApiToken')) {
+      // log with no token or user info
```
annyhe: log regardless of whether api token enforcement.

annyhe:
@pallavi2209 @iamigo Ready for more comments :)

config/toggles.js
```diff
@@ -69,6 +69,10 @@ const longTermToggles = {
   enforceWritePermission: environmentVariableTrue(pe,
                                         'ENFORCE_WRITE_PERMISSION'),
 
+  // Enable api activity logging
+  enableApiLogs: environmentVariableTrue(pe,
```
iamigo: Can we use "ENABLE_API_ACTIVITY_LOGS" so it follows the same naming convention as the worker one?

package.json
iamigo: I like the blanket coverage by doing it this way, but it also means these tests are not counted by coveralls. Can you think of another way of achieving this? We've added a test utility function called "toggleOverride" which lets us turn a feature on or off in a test's before/after function (as long as the feature doesn't require re-initializing top-level app server index.js).

utils/apiLog.js
pallavi2209: Then maybe create a folder activityLogs and put all the logging files together there? Just trying to prevent having too many root level files (similar kind) in utils folder.

package.json
annyhe: @iamigo Another way of testing would be to toggleOverride enable api logs for all test api tests, but how would that differ from what is happening here?
As for how running this line not being counted by coveralls, as long as coveralls counts the original tests/api, we don't need coveralls to count it again.

iamigo: OK @annyhe

utils/apiLog.js
```diff
@@ -0,0 +1,108 @@
...
+    } else {
+      // if enforcing API token,
+      // extract user, token to update log object
+      jwtUtil.getTokenDetailsFromToken(req)
+      .then((resObj) => {
+        logObject.user = resObj.username;
+        logObject.token = resObj.tokenname;
+
+        // log with the token
+        combineAndLog(resultObj, logObject);
```
pallavi2209: Why check enforceApiToken? Since you are not catching any error from getTokenDetailsFromToken, why not just call combineAndLog outside then? Would be cleaner code and no dependency on enforceApiToken

utils/apiLog.js
```diff
@@ -0,0 +1,108 @@
...
+  activityLogUtil.printActivityLogString(logObject, 'api');
+}
+
+/**
+ * Combines input obj with values from the request object
+ * then logs
+ * @param {Object} req - the request object
+ * @param {Object} resultObj - Object with the rest of the fields to print
+ */
+function logAPI(req, resultObj) {
```
pallavi2209: Can we test this function?

utils/apiLog.js
annyhe: good point. will change

pallavi2209:
I don't see code for the API's not using verbs (like user tokens, authenticate, register, ssoconfig etc). Requested @annyhe to create another PR for that.

utils/apiLog.js
annyhe: Regarding the test, because there is an async function involved, sinon would need a passed-in callback to check that the .then contents have been called
https://www.sitepoint.com/sinon-tutorial-javascript-testing-mocks-spies-stubs/

utils/apiLog.js
annyhe: let's move it in another pull request

annyhe:
after speaking with @pallavi2209 there are missing areas in api that are not being logged. Will do a separate pull request for that (ie. api/controllers/**.js)

annyhe:
@iamigo Please review.

api/v1/helpers/verbs/doDelete.js
```diff
@@ -50,6 +53,8 @@ function doDelete(req, res, next, props) {
 
     // when a resource is deleted, delete all its associations too
     u.deleteAllAssociations(o, assocNames);
+    resultObj.retval = o.dataValues;
+    u.logAPI(req, resultObj);
```
iamigo: I don't like the mix of calling this function both as u.logAPI(...) and as logAPI(...) (line 36) in the same module--it's too confusing to the reader. Please standardize and call it one way across all API modules.

config/toggles.js
```diff
@@ -69,6 +69,10 @@ const longTermToggles = {
   enforceWritePermission: environmentVariableTrue(pe,
                                         'ENFORCE_WRITE_PERMISSION'),
 
+  // Enable api activity logging
+  enableApiWorkerLogs: environmentVariableTrue(pe,
```
iamigo: `enableApiActivityLogs`

api/v1/helpers/verbs/doDelete.js
annyhe: u.logAPI and logAPI were different functions. Renamed logAPI to logAuditAPI to avoid confusion.

annyhe:
@iamigo Ready for more comments :)

api/v1/helpers/verbs/doDelete.js
iamigo: So just to be clear, "logAuditAPI" is the old audit log stuff that we will remove once the new activity logging is in place?

annyhe: yes, logAuditAPI is the old audit log. The new API logger is called logAPI.

utils/jwtUtil.js
```diff
@@ -91,7 +91,9 @@ function getTokenDetailsFromTokenString(s) {
     if (s) {
       jwt.verify(s, env.tokenSecret, {}, (err, decodedData) => {
         if (err !== null || !decodedData) {
-          return reject(err);
+          return reject(new apiErrors.ForbiddenError({
```
annyhe: @iamigo I added the forbidden status code, to be consistent with the other return.

PR #205, opened 2017-01-19
Worker logs: sampleTimeout job

pallavi2209:


pallavi2209:
@iamigo Yes this is for sampleTimeout job, worker logging. But I assumed we are covering clock logging as a part of worker logging, because clock activities are executed in worker dynos. I didn't know there is different story for clock logging as well.
Let me update the PR title to avoid confusion.

worker/jobProcessor.js
```diff
@@ -76,9 +76,34 @@ jobQueue.process(jobType.BULKUPSERTSAMPLES, (job, done) => {
...
+      objToReturn.recordCount = dbRes.numberTimedOut;
+
+      // errorCount = number of samples that did not time out
+      objToReturn.errorCount = dbRes.numberEvaluated - dbRes.numberTimedOut;
+      objToReturn.workTime = dbEndTime - jobStartTime;
+      objToReturn.reqStartTime = reqStartTime;
+      done(null, objToReturn);
+    } else {
+      done();
+    }
```
shriramshankar: Lines 94 to 103 seems similar to lines 53 to 69, expect for a few lines. Do you think moving them to the activityLogUtils will help?

pallavi2209: For current calculations of dbTime, queueTime and workTime, looks like we can use common method, but recordCount and errorCount can vary from job to job. Let's move them actually, will improve readability.

pallavi2209:
@shriramshankar updated the PR

PR #212, opened 2017-01-23
Correct user token endpoints

pallavi2209:
GET /v1/users/{key}/tokens now correctly returns tokens of given user, instead of all tokens
GET /v1/users/{key}/tokens/{tokenName} now correctly returns given token, instead of first token

### DIFF
--- a/api/v1/controllers/userTokens.js
+++ b/api/v1/controllers/userTokens.js
@@ -26,7 +26,8 @@ const authUtils = require('../helpers/authUtils');
 function whereClauseForUser(userNameOrId) {
   const whr = {};
   if (u.looksLikeId(userNameOrId)) {
-    // need to use '$table.field$' for association fields
+    /* need to use '$table.field$' for association fields. User IDs in the url
+    are case-sensitve. */
     whr.where = { '$User.id$': {} };
     whr.where['$User.id$'] = userNameOrId;
   } else {
--- a/api/v1/controllers/userTokens.js
+++ b/api/v1/controllers/userTokens.js
@@ -25,10 +25,10 @@ const authUtils = require('../helpers/authUtils');
  */
 function whereClauseForUser(userNameOrId) {
   const whr = {};
-  if (u.looksLikeId()) {
+  if (u.looksLikeId(userNameOrId)) {
     // need to use '$table.field$' for association fields
     whr.where = { '$User.id$': {} };
-    whr.where['$User.id$'][cnstnts.SEQ_LIKE] = userNameOrId;
+    whr.where['$User.id$'] = userNameOrId;
   } else {
     whr.where = { '$User.name$': {} };
     whr.where['$User.name$'][cnstnts.SEQ_LIKE] = userNameOrId;
--- a/api/v1/controllers/userTokens.js
+++ b/api/v1/controllers/userTokens.js
@@ -9,7 +9,7 @@
 /**
  * api/v1/controllers/userTokens.js
  */
-'use strict';
+'use strict'; // eslint-disable-line strict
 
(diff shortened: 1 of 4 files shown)


api/v1/controllers/userTokens.js
```diff
@@ -165,6 +182,13 @@ module.exports = {
         const whr = whereClauseForUserAndTokenName(user, tokenName);
         helper.model.findOne(whr)
         .then((o) => {
+          if (!o) {
+            const err = new apiErrors.ResourceNotFoundError();
+            err.resource = helper.model.name;
+            err.key = user + ', ' + tokenName;
+            throw err;
+          }
+
```
pallavi2209: Let me add some tests for this as well.

pallavi2209:
@shriramshankar @iamigo PR open for comments.

api/v1/controllers/userTokens.js
```diff
@@ -18,25 +18,35 @@ const u = require('../helpers/verbs/utils');
...
-    {
-      'user.name': {},
-    },
-    {
-      'user.id': {},
-    },
-  ];
-  whr[cnstnts.SEQ_OR][0]['user.name'][cnstnts.SEQ_LIKE] = user;
-  whr[cnstnts.SEQ_OR][1]['user.id'][cnstnts.SEQ_LIKE] = user;
+  if (u.looksLikeId()) {
```
iamigo: `if (u.looksLikeId(userNameOrId)) {`

pallavi2209:
@iamigo Corrected the uuid check.

api/v1/controllers/userTokens.js
```diff
@@ -18,25 +18,35 @@ const u = require('../helpers/verbs/utils');
...
-      'user.id': {},
-    },
-  ];
-  whr[cnstnts.SEQ_OR][0]['user.name'][cnstnts.SEQ_LIKE] = user;
-  whr[cnstnts.SEQ_OR][1]['user.id'][cnstnts.SEQ_LIKE] = user;
+  if (u.looksLikeId(userNameOrId)) {
+    // need to use '$table.field$' for association fields
+    whr.where = { '$User.id$': {} };
+    whr.where['$User.id$'] = userNameOrId;
+  } else {
```
shriramshankar: aren't the where clauses defined something like this ?
where : {
    User: {
      id: "id" 
     }
}

pallavi2209: I don't think this is working given User is an association of token table. 
{ where: { User: { id: 'a568a893-0571-42b9-aa55-ca44e1308346' } } } gives me 'val.replace is not a function' error.

api/v1/controllers/userTokens.js
```diff
@@ -25,10 +25,10 @@ const authUtils = require('../helpers/authUtils');
  */
 function whereClauseForUser(userNameOrId) {
   const whr = {};
-  if (u.looksLikeId()) {
+  if (u.looksLikeId(userNameOrId)) {
     // need to use '$table.field$' for association fields
     whr.where = { '$User.id$': {} };
-    whr.where['$User.id$'][cnstnts.SEQ_LIKE] = userNameOrId;
+    whr.where['$User.id$'] = userNameOrId;
```
iamigo: Doing it this way instead of using `$ilike` means that we would no longer return a record with id `6da84a2a-f4b7-4ff1-b97b-56ce92501361` if we put `6da84a2a-F4B7-4ff1-b97b-56ce92501361` in the url because of uppercase/lowercase mismatch. I'm OK with that, and it's better for db performance, but then we should document that record ids in a url are case sensitive.

pallavi2209:
@iamigo Added documentation in swagger yaml.

PR #213, opened 2017-01-23
Right click to open perspectives in a new tab

pallavi2209:


pallavi2209:
This is how it looks now.
screen shot 2017-01-23 at 10 52 24 am

PR #215, opened 2017-01-24
Removing these node settings gives us more flexibility

iamigo:
... to change dyno sizes in heroku


pallavi2209:
Looks like the tests are not passing after merging with master because of some code style errors.

iamigo:
weird that the error was in this PR and not the other one, will fix in a minute

PR #220, opened 2017-01-26
api changes for aspect sorting by rank

shriramshankar:


tests/api/v1/aspects/get.js
```diff
@@ -133,7 +145,7 @@ describe(`api: GET ${path}`, () => {
       .set('Authorization', token)
       .expect(constants.httpStatus.OK)
       .expect((res) => {
-        expect(res.body.length).to.be.equal(TWO);
+        expect(res.body.length).to.be.equal(THREE);
```
iamigo: Need a "hierarchy" test which shows that the new field is included in the list of aspect fields returned for every sample.

shriramshankar: The hierarchy tests for the aspect rank were in the other PR.

iamigo: oh yeah! thx!

PR #234, opened 2017-02-02
Do not return token from /v1/authenticate api

pallavi2209:
Create /authenticate route instead, so that token is not exposed to user through API.

### DIFF
--- a/api/v1/controllers/authenticate.js
+++ b/api/v1/controllers/authenticate.js
@@ -42,21 +42,18 @@ module.exports = {
         return u.handleError(next, loginErr, resourceName);
       }
 
-      // Create a new token on login.
-      const createdToken = jwtUtil.createToken(user.name, user.name);
-
       req.logIn(user, (_err) => {
         resultObj.dbTime = new Date() - resultObj.reqStartTime;
         if (_err) {
           return u.handleError(next, _err, resourceName);
         }
 
-        u.logAPI(req, resultObj, createdToken);
-        return res.status(httpStatus.OK).json({
+        const retObj = {
           success: true,
           message: 'authentication succeeded',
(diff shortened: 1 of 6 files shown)


api/v1/controllers/authenticate.js
```diff
@@ -42,21 +42,18 @@ module.exports = {
...
         }
 
-        u.logAPI(req, resultObj, createdToken);
-        return res.status(httpStatus.OK).json({
+        const retObj = {
           success: true,
           message: 'authentication succeeded',
-          token: createdToken,
-        });
+        };
```
iamigo: Is there any wording that needs to be updated in our documentation or swagger docs around this?

pallavi2209: Yes, swagger needs to be updated. Let me add that.

iamigo: Thx!

Hey, aren't you supposed to be learning!? ;)

view/loadView.js
```diff
@@ -215,13 +215,7 @@ module.exports = function loadView(app, passport) {
...
       if (_req.user && _req.user.name) {
         const token = jwtUtil.createToken(_req.user.name, _req.user.name);
-        _res.cookie(
-          'Authorization',
-          token, {
-            secure: true,
-            httpOnly: true,
-          }
-        )
+        _res.cookie('Authorization', token);
```
pallavi2209: Removing {secure:true, httpOnly: true} (was added in #140 ) because token is not set in cookie and hence SSO user gets Forbidden when USE_ACCESS_TOKEN=true.
@annyhe Please confirm if this is okay.

pallavi2209:
@iamigo Updated the PR

PR #235, opened 2017-02-02
added worker id

harshkothari410:


config/activityLog.js
```diff
@@ -40,6 +40,7 @@ module.exports = {
       errorCount: 0,
       ipAddress: 'None',
       jobType: 'None',
+      jobId: 'None',
```
iamigo: Please put the fields in alphabetical order: jobId before jobType

jobQueue/jobWrapper.js
```diff
@@ -124,6 +124,10 @@ function logAndRemoveJobOnComplete(req, job) {
       logObject.jobType = job.type;
     }
 
+    if (job.id) {
```
iamigo: Right.

config/activityLog.js
harshkothari410: Okies

PR #238, opened 2017-02-03
subject tags should be passed as include or exclude filter

harshkothari410:


api/v1/apiErrors.js
```diff
@@ -71,6 +71,16 @@ apiErrors.create({
...
 });
 
+apiErrors.create({
+  code: 11101,
+  status: 400,
+  name: 'InvalidSubjectTagsParameterError',
+  parent: apiErrors.ValidationError,
+  fields: [],
+  defaultMessage: 'Subject Tags should be passed in query parameter as ' +
+  'include filter or exclude filter not the combination of both.',
```
iamigo:
```
  defaultMessage: 'Subject Tags should be passed in query parameter as ' +
  'an include filter or an exclude filter, but not the combination of both.',
```

api/v1/controllers/subjects.js
```diff
@@ -60,6 +60,29 @@ function checkDuplicates(tagsArr) {
 }
 
 /**
+ * Validates the correct filter of Subject Tags
+ * passed in query parameters
+ * @param {Array} subjectTagsArray Subject Tags Array
+ */
+function subjectTagsValidation(subjectTagsArray) {
```
iamigo: Would be more appropriate to use a verb for the function name here: "validateSubjectTags"

api/v1/swagger.yaml
```diff
@@ -4359,6 +4359,8 @@ paths:
            sign to indicate that the subject having tags with those names will
            be excluded. For example, ?subjectTags=-TAG3,-TAG4 will return the
            subject hierarchy without subjects having tags -  TAG3 and TAG4.
+           Tags should be passed as include or exclude filter not the combination
```
iamigo: Tags should be passed as an include filter or as an exclude filter but not the combination of both.

tests/api/v1/subjects/getHierarchyAspectAndTagsFilters.js
```diff
@@ -237,7 +237,47 @@ describe(`api: GET ${path}:`, () => {
...
+      .end((err /* , res */) => {
+        if (err) {
+          return done(err);
+        }
+
+        done();
+      });
+    });
   });
+
```
iamigo: I think you also need a test for this:

'?subjectTags=verycold,-ea,cold'

api/v1/controllers/subjects.js
```diff
@@ -62,22 +62,22 @@ function checkDuplicates(tagsArr) {
 /**
  * Validates the correct filter of Subject Tags
  * passed in query parameters
- * @param {Array} subjectTagsArray Subject Tags Array
+ * @param {Array} validateSubjectTags Subject Tags Array
  */
-function subjectTagsValidation(subjectTagsArray) {
+function subjectTagsValidation(validateSubjectTags) {
```
iamigo: I meant change the function name, not the arg name

harshkothari410: oops

harshkothari410:
@iamigo is there any change required? Let me know.

shriramshankar:
Can you load the perspectives and test it? The perpectives call the hierarchy endpoint and you might have to make changes there too.

shriramshankar:
@harshkothari410 Also, this check is for all the filters on the hierarchy endpoint. Sorry it looks like it was not clear from the story description

harshkothari410:
@shriramshankar I did not get you

shriramshankar:
@harshkothari410  the hierarchy endpoint takes the following 4 query parameters
1. subjectTags
2. aspectTags
3. aspect
4.sample

The validation that you have done just checks the subjectTags parameters. You need to to the check for the other 3 too. Also, the perspective layer (views/perspective) converts the perspectives created/updated by the user to the hierarchy endpoint. You probably will have to change this so that it passes these validations.

shriramshankar:
@harshkothari410 Can we talk about the changes you are making on hangout tomorrow?

harshkothari410:
@shriramshankar yes sure

harshkothari410:
@shriramshankar updated PR.

api/v1/controllers/subjects.js
```diff
@@ -60,6 +60,26 @@ function checkDuplicates(tagsArr) {
 }
 
 /**
+ * Validates the correct filter parameter
+ * passed in query parameters
+ * @param {Array} filterParams Filter Tags Array
+ */
+function validateFilterParams(filterParams) {
+  let subjectTagsCounter = 0;
```
iamigo: Instead of this...
```
+  let subjectTagsCounter = 0;
+
+  for (let i = 0; i < filterParams.length; i++) {
+    if (filterParams[i][ZERO] === '-') {
+      subjectTagsCounter++;
+    }
+  }
```

... why not this... ?
```
const EXCLUDE_SYMBOL = '-';
subjectTagsCounter = filterParams
  .filter((i) => i.startsWith(EXCLUDE_SYMBOL)).length;
```

harshkothari410: Yo made changes as per your comment @iamigo

PR #240, opened 2017-02-04
Story 1 for adding the sort by column

Prashanth-Billa:
I have made the changes, and wanted you to give me a feedback so that I can improve and get back if required.

Thank you

### DIFF
--- a/db/model/subject.js
+++ b/db/model/subject.js
@@ -101,6 +101,13 @@ module.exports = function subject(seq, dataTypes) {
       type: dataTypes.STRING(constants.fieldlen.longish),
       allowNull: true,
     },
+    sortBy: {
+      type: dataTypes.STRING(constants.fieldlen.longish),  
+      allowNull: true,
+      validate: {
+        is: constants.nameRegex,
+      },
+    },
     relatedLinks: {
       type: dataTypes.ARRAY(dataTypes.JSON),
       allowNull: true,
--- a/db/model/subject.js
+++ b/db/model/subject.js
@@ -101,13 +101,6 @@ module.exports = function subject(seq, dataTypes) {
       type: dataTypes.STRING(constants.fieldlen.longish),
       allowNull: true,
     },
-    sortBy: {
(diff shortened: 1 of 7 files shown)


db/model/subject.js
```diff
@@ -101,6 +101,13 @@ module.exports = function subject(seq, dataTypes) {
       type: dataTypes.STRING(constants.fieldlen.longish),
       allowNull: true,
     },
+    sortBy: {
+      type: dataTypes.STRING(constants.fieldlen.longish),
```
iamigo: On second thought, 4096 is too long for this use case... please add a constant to constants.fieldlen called "sortField" with length 254.

migrations/20170203233255-addSortByToSubject.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+module.exports = {
+  up: function (queryInterface, Sequelize) {
+    /*
```
iamigo: you need to add your up/down implementations in this file

tests/db/model/subject/update.js
```diff
@@ -218,6 +218,22 @@ describe('db: subject: update: ', () => {
       })
       .catch(done);
     });
```
iamigo: add a test where sortBy is empty and you update it and then confirm that it's populated
add a test where sortBy is not empty and you update it to set it to null and then confirm that it's null
add a test where sortBy is not empty and you update it to set it to empty string and then confirm that it's empty string

tests/db/model/subject/utils.js
```diff
@@ -21,6 +21,7 @@ const subjectPrototype = {
   helpUrl: 'http://www.bar.com',
   imageUrl: 'http://www.bar.com/foo.jpg',
   isPublished: true,
+  sortBy: ''
```
iamigo: Just curious, why are you setting this here?

Prashanth-Billa:
Hey Ian,

All the changes have been made and the codeline is updated. Request you to review and let me know any changes are required.

Thanks

PR #241, opened 2017-02-06
Add sortBy field to subject model

Prashanth-Billa:


db/constants.js
```diff
@@ -23,6 +24,7 @@ module.exports = {
   sampleNameSeparator: '|',
   defaultJsonArrayValue: [],
   defaultArrayValue: [],
+  defaultSortByValue: null,
```
iamigo: why add this? field will default to null if nothing is specified, right?

tests/db/model/subject/utils.js
```diff
@@ -20,7 +20,7 @@ const subjectPrototype = {
   helpEmail: 'foo@bar.com',
   helpUrl: 'http://www.bar.com',
   imageUrl: 'http://www.bar.com/foo.jpg',
-  isPublished: true,
+  isPublished: true
```
iamigo: our style guide requires a trailing comma here

tests/db/model/subject/utils.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
     const s = JSON.parse(JSON.stringify(subjectPrototype));
     s.name = name;
     s.parentId = parentId;
+    s.sortBy = null
```
iamigo: our style guide requires a terminating semi-colon here

iamigo:
951 passing (55s)
  35 pending
  1 failing
  1) db: subject: create:  Simple Subjects, i.e. no parents/children should fail, sort by too long:
     Error: should have failed since sort by is too long
      at Subject.create.then (tests/db/model/subject/create.js:178:14)
      at tryCatcher (node_modules/bluebird/js/release/util.js:16:23)

PR #243, opened 2017-02-09
DO NOT MERGE: proof-of-concept for the sample store: initialize from db

iamigo:
(This isn’t necessarily code I want to check in. This is just for proof-of-concept.)

### DIFF
--- a/utils/sampleStoreUtil.js
+++ b/utils/sampleStoreUtil.js
@@ -1,0 +1,100 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * utils/sampleStoreUtil.js
+ *
+ * Assumes that there is a Set in Redis with key "samsto:subjects" and a Hash
+ * in Redis for every subject which has samples, keyed like
+ * "samsto:subjects:LOWERCASE_OF_SUBJECT_ABSOLUTE_PATH". Each Hash contains
+ * the subject's samples: each field in the Hash is a lower-case aspect name,
+ * with the value being the stringified sample.
+ */
+'use strict'; // eslint-disable-line strict
+const r = require('../cache/redisCache').client;
+const Sample = require('../db').Sample;
+const PFX = 'samsto';
+const SEP = ':';
+const SUBJECT_SET = PFX + SEP + 'subjects';
+
+/**
+ * Get subject absolute path from sample name.
+ *
+ * @param {String} name - The sample name.
+ * @return {String} - The subject's absolute path.
+ */
+function getAbsolutePath(name) {
+  return name.split('|')[0];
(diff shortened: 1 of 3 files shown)


iamigo:
@pallavi2209 @shriramshankar does this init() function give you what you need to be able to work on your stories (rebuild hierarchy from db+redis, assess read/write concurrency)? If not, what can I add to this to make it more useful?

shriramshankar:
@iamigo I was just looking at it. The init function is all I need to re-create the hierarchy.

shriramshankar:
Have some tust1 meetings that I need to attend today. Will pull this branch down and start writing the hierarchy party after the trust1 stuff.

utils/sampleStoreUtil.js
```diff
@@ -0,0 +1,100 @@
...
+      const sample = i.get();
+      const subkey = SUBJECT_SET + SEP +
+        getAbsolutePath(sample.name).toLowerCase();
+      subjectKeys.add(subkey);
+      const samkey = sample.aspect.name.toLowerCase();
+      delete sample.aspect;
+      commands.push(['hset', subkey, samkey, JSON.stringify(sample)]);
+    });
+    commands.push(['sadd', SUBJECT_SET, Array.from(subjectKeys)]);
+    console.log('REDIS COMMANDS (BATCH):', commands);
```
shriramshankar: Do you need to do Array.from at line 83. ? I think the keys would become array indices (0,1....) instead of  "SUBJECT_SET + SEP+ subject absolute path". Correct?

pallavi2209: getAbsolutePath gets the subject absolute path only @shriramshankar

pallavi2209:
@iamigo The code looks great. Do you have a working app with samples populated using the init function?

utils/sampleStoreUtil.js
shriramshankar: Just tested this locally, we need to do Array.from.

pallavi2209: Oh, ignore my comment as well, I misunderstood your comment.

utils/sampleStoreUtil.js
```diff
@@ -0,0 +1,100 @@
...
+    const subjectKeys = new Set();
+    const commands = [];
+    console.log(`Found ${samples.length} samples to store in redis`);
+    samples.forEach((i) => {
+      const sample = i.get();
+      const subkey = SUBJECT_SET + SEP +
+        getAbsolutePath(sample.name).toLowerCase();
+      subjectKeys.add(subkey);
+      const samkey = sample.aspect.name.toLowerCase();
+      delete sample.aspect;
```
shriramshankar: We would also need some kind of an "one time" init function for the aspects. If we are caching aspects like we cache the perspectives, we would have to do a "get" on all the aspects to for them to be in Redis.

shriramshankar:
I was able to recreate the hierarchy endpoint (without filters )by working off of this branch. The TTBF for subjects/Salesforce/hierarchy endpoint with this implementation is ~200-250ms VS 1.45s when not using redis to store samples. I have used native promises for the implementation, my guess is, if we use blue bird or async it will be a bit faster than this.

PR #244, opened 2017-02-09
on get unnamed perspective with query, redirect to default 

annyhe:
perspective with query


view/perspective/app.js
```diff
@@ -572,7 +572,8 @@ function getPerspective(perspNameOrId) {
  * @param {Array} pnames - Array of perspective names
  */
 function whichPerspective(pnames) {
-  let h = window.location.href;
```
annyhe: window.location.href included the query. The page was not loading .../perspectives?rootSubject=NorthAmerica. 
window.location.pathname includes only the part after host and before the query. Therefore we need to add the query on at the window.location.href.

PR #247, opened 2017-02-10
API layer changes and test cases for the new Subject sortBy column

Prashanth-Billa:
First version


Prashanth-Billa:
Merge conflicts in pull request resolved

api/v1/swagger.yaml
```diff
@@ -6998,6 +7043,11 @@ definitions:
         maxLength: 2082
         description: >
           Authentication success message.
+      token:
```
iamigo: this shouldn't be here

migrations/20170203233255-addSortByToSubject.js
```diff
@@ -10,8 +10,9 @@ module.exports = {
     return qi.sequelize.transaction(() => qi.addColumn('Subjects', 'sortBy', {
       type: Sequelize.STRING(constants.fieldlen.sortField),
       allowNull: true,
+      defaultValue: '',
```
iamigo: you can't modify the migration--it's already been deployed
if you need more db changes, you have to create a new migration

tests/api/v1/subjects/put.js
```diff
@@ -560,4 +600,4 @@ describe('api: PUT subjects with tags', () => {
       done();
     });
   });
-});
+});
```
iamigo: must have trailing newline

migrations/20170203233255-addSortByToSubject.js
Prashanth-Billa: resolved

tests/api/v1/subjects/put.js
Prashanth-Billa: resolved

Prashanth-Billa:
Style issues resolved for all test files. create, update in subject model. get, put, post, patch in API db subject

iamigo:
@Prashanth-Billa test failures on Subject find?

migrations/20170213221219-newSortByColumnForSubject.js
```diff
@@ -0,0 +1,28 @@
...
+      type: Sequelize.STRING(constants.fieldlen.sortField),
+      allowNull: true,
+      defaultValue: '',
+      validate: {
+        is: /^[0-9a-z_-]*$/i,
+      },
+    }));
+  },
+
+  down: function (qi, Sequelize) {
```
iamigo: This migration code is wrong--it should be modifying the table definition from its previous state. The column *already* exists (from your last migration). What specifically are you changing from what the previous commit defined?

migrations/20170213221219-newSortByColumnForSubject.js
```diff
@@ -0,0 +1,28 @@
...
+    /*
+      Add reverting commands here.
+      Return a promise to correctly handle asynchronicity.
+    */
+    return qi.sequelize.transaction(() => qi.removeColumn('Subjects', 'sortBy'));
```
iamigo: same thing here--should revert to whatever was there previously

tests/api/v1/subjects/get.js
```diff
@@ -23,335 +23,381 @@ const ONE = 1;
...
-      vt.parentId = us.id;
-      return Subject.create(vt);
-    })
-    .then((createdVt) => {
-      vt.id = createdVt.id;
-      done();
-    })
-    .catch(done);
-  });
+    before((done) => {
```
iamigo: Tabs should be two spaces, never four.

tests/api/v1/subjects/get.js
```diff
@@ -23,335 +23,381 @@ const ONE = 1;
...
-      .end((err/* , res */) => {
-        if (err) {
-          done(err);
-        }
-
-        done();
+    after(u.forceDelete);
+    after(tu.forceDeleteUser);
+
+    describe('duplicate tags fail', () => {
```
iamigo: Please make sure all tab indents meet style guidelines.

iamigo:
Good job on code coverage!

migrations/20170213221219-newSortByColumnForSubject.js
Prashanth-Billa: I hope the changeColumn method would work. The UP and DOWN states now reflect the new and previous state correspondingly. Pls review. Thanks

iamigo: You only have to include the attribute which have changed here: defaultValue and the validate regex.

Prashanth-Billa:
ok, defaultValue and validate fields are only included.

api/v1/swagger.yaml
```diff
@@ -4000,6 +4016,13 @@ paths:
                 description: >
                   Max 60 characters; case insensitive; allows alpha-numeric characters,
                   underscore (_) and dash (-).
+              sortBy:
```
pallavi2209: Overall changes look good. I just felt that we can one more line in description on how this sortBy field is used. Something Like "Default empty. When set to some string, the subjects are sorted alphabetically on string value."
@Prashanth-Billa If you want, you can make this change in next PR, I will merge this now

Prashanth-Billa: sure. sounds good! Will include the description change as part of story 3 - view layer. Thank you

iamigo: Good suggestion. I don't think the core Refocus app would actually use this field to sort subjects in the admin console or even in api response... it's more about making it available as an option for the Lens developer to choose to use or not. Go back to the description in the user story and try to come up with some language to add to swagger based on that.

PR #251, opened 2017-02-16
Reject revoked tokens.

iamigo:


tests/enforceToken/revoke.js
```diff
@@ -0,0 +1,122 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/tokenReq/api/token/createtoken.js
```
pallavi2209: tests/enforceToken/revoke.js

utils/jwtUtil.js
```diff
@@ -56,25 +97,49 @@ function verifyToken(req, cb) {
...
-            return cb();
-          }
+            req.user = user; // set user in request
 
-          return handleInvalidToken(cb);
+            /*
+             * No need to check the token record if this is the default UI
+             * token.
+             */
+            if (decodedData.username === decodedData.tokenname) {
```
pallavi2209: So, are we restricting users to create tokenname = username? Or we are going to follow the strategy of having an extra field like, `isUIToken: true` when creating token from UI.

pallavi2209:
Tests failing?

utils/jwtUtil.js
iamigo: After talking all of that over with you the other day, I convinced myself that I couldn't make that change without invalidating all the existing tokens, but I don't think that's true. In any case, this logic still needs to be there to handle the existing tokens, even if we start encoding new tokens with the extra field, so I think we should leave this as is, and create a new story to encode that new field into new tokens. What do you think, @pallavi2209?

pallavi2209: With this code in place, we will not restrict a user who created token (with tokenName =username) and the token is revoked. We can merge this code for now. What we can do in another story is that we can keep the existing tokens as it is, add a field `isUIToken: true` to the tokens created on the fly from UI, and then add a check `if (decodedData.isUIToken && decodedData.isUIToken ==true) { cb() }` instead of check `if (decodedData.username === decodedData.tokenname) { cb() }`
Let me create a story for that.

tests/enforceToken/revoke.js
```diff
@@ -0,0 +1,122 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/tokenReq/api/token/revoke.js
```
pallavi2209: tests/enforceToken/revoke.js

shriramshankar:
Looks good.

PR #255, opened 2017-02-17
DO NOT MERGE: SampleStore: Sample fields are stored as hash: sample initialized from db.

shriramshankar:


shriramshankar:
@iamigo here is the new sampleStore utils with the samples stored in its own set. Each of the samples is stored as an hashmap with its fields as the key and value as the field's value. 
The hierarchy recreation is not done yet.I am off tomorrow and will have some time on Sunday to get it done.

utils/sampleStoreUtil.js
```diff
@@ -0,0 +1,132 @@
...
+      const sample = i.get();
+      const subkey = SUBJECT_SET + SEP +
+        getAbsolutePath(sample.name).toLowerCase();
+      subjectKeys.add(subkey);
+      const samkey = sample.aspect.name.toLowerCase();
+      delete sample.aspect;
+      commands.push(['sadd', subkey, samkey]);
+
+      // push each of the sample into its own hash map
+      const sampleHashKey = SAMPLE_SET + SEP + sample.name;
```
shriramshankar: this needs to be in lowercase. Will do in when I do the hierarchy

utils/sampleStoreUtil.js
```diff
@@ -0,0 +1,132 @@
...
+    const commands = [];
+    console.log(`Found ${samples.length} samples to store in redis`);
+    samples.forEach((i) => {
+      const sample = i.get();
+      const subkey = SUBJECT_SET + SEP +
+        getAbsolutePath(sample.name).toLowerCase();
+      subjectKeys.add(subkey);
+      const samkey = sample.aspect.name.toLowerCase();
+      delete sample.aspect;
+      commands.push(['sadd', subkey, samkey]);
```
pallavi2209: This is adding aspect name as member to a set of subject absolute path. Correct? We can add sample name instead of aspect name.

index.js
```diff
@@ -88,6 +88,13 @@ function start() { // eslint-disable-line max-statements
   const PORT = process.env.PORT || conf.port;
   app.set('port', PORT);
 
+  // ******************** USIING REDIS AS A DATA STORE ***********************
+
+  // INITIALIZING SAMPLES INTO REDIS
+  require('./utils/sampleStoreUtil.js').init();
```
pallavi2209: We can create hash maps of aspects as well, just like samples.

utils/sampleStoreUtil.js
shriramshankar: How does having sample name help VS having just aspect name?

index.js
shriramshankar: Yes we need a map for Aspects.

utils/sampleStoreUtil.js
pallavi2209: Umm..not better I guess, we just thought we get samples first when getting hierarchy, and then aspect. But I guess, both ways are similar.
Also, do you think storing this as hash with both aspect names and sample names will help getting both at same time (no then). Better than set? No better? Doesnt matte muchr?

utils/sampleStoreUtil.js
```diff
@@ -0,0 +1,132 @@
...
+              JSON.stringify(sample[field])]);
+          } else {
+            commands.push(['hset', sampleHashKey, field, sample[field]]);
+          }
+        }
+
+      });
+    });
+    commands.push(['sadd', SUBJECT_SET, Array.from(subjectKeys)]);
+    commands.push(['sadd', SAMPLE_SET, Array.from(sampleKeys)]);
```
pallavi2209: Why this set?

utils/sampleStoreUtil.js
shriramshankar: Lets just use one of  sample name or aspect name. It does not really matter to be but since we already have something for aspect name. Lets just use that.

utils/sampleStoreUtil.js
shriramshankar: The sets are used by the purge function for deleting the subject and sample hash.

utils/sampleStoreUtil.js
iamigo: OK--aspect name.

iamigo:
@shriramshankar close?

PR #256, opened 2017-02-17
Story 3 - View layer changes for the new Subject sortBy column

Prashanth-Billa:


annyhe:
@Prashanth-Billa This pull request contains more changes than the views layer. For views changes we usually deploy a demo app to check it out. I deployed a demo app at http://refocus-pr-256.herokuapp.com/. Will need to wait a few minutes for the app to be provisioned

Prashanth-Billa:
Sure. Will also add a detailed description for the column in a new commit now (Swagger API).

All commits except 'Story 3 - View layer changes for the new attribute 'sortBy' addition' have already been merged into the refocus main. (They were part of previous user stories). Thanks.

annyhe:
For the view files you only changed the config files, so I don't see a need for test files.

PR #259, opened 2017-02-21
Adjust the max sample name length…

iamigo:
to be the sum of subject abs path max plus aspect name max plus 1 (for the “|” pipe)

### DIFF
--- a/db/model/sample.js
+++ b/db/model/sample.js
@@ -30,6 +30,8 @@ const messageCodeLen = 5;
 const assoc = {};
 const EMPTY_STRING = '';
 const NO = 0;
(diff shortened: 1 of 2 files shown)


shriramshankar:
Looks good but I was wondering why do we need to do a, +1, instead of changing constants.fieldlen.normalName directly?

iamigo:
@shriramshankar I wanted to be able to accommodate max subject absolute path length plus max aspect name length plus one extra character for the pipe character that we add in between the subject and aspect.

shriramshankar:
@iamigo Got it. Thanks.

PR #261, opened 2017-02-23
Multiredis

iamigo:
Allows admin to pin each of the various redis use cases to a specific redis instance based on config.

Note: Leaving out docs for now, pending the creation of a doc on moving Refocus to production (separate user story).


config.js
```diff
@@ -114,6 +111,25 @@ module.exports = {
...
+        pe.REDIS_PUBSUB || pe.REDIS_URL || DEFAULT_LOCAL_REDIS_URL,
+      queue:
+        pe.REDIS_QUEUE || pe.REDIS_URL || DEFAULT_LOCAL_REDIS_URL,
+      realtimeLogging:
+        pe.REDIS_REALTIME_LOGGING || pe.REDIS_URL || DEFAULT_LOCAL_REDIS_URL,
+      sampleStore:
+        pe.REDIS_SAMPLE_STORE || pe.REDIS_URL || DEFAULT_LOCAL_REDIS_URL,
+      session:
+        pe.REDIS_SESSION || pe.REDIS_URL || DEFAULT_LOCAL_REDIS_URL,
+    },
```
shriramshankar: do we need a different redis for session, realtimeLogging and limiter?

iamigo: We don't *need* to... but once I was identifying all the different uses, I just went with making *all* of them configurable, giving admin the ability to leave stuff on the main redis instance or move anything off to a separate instance.

shriramshankar: Also, if an admin configures a separate redis instance for each of the above use cases nothing will remain on the main redis.  Are we ok with this? It does not matter to me, though.

cache/redisCache.js
```diff
@@ -7,19 +7,23 @@
...
 
 module.exports = {
-  client,
+  client: {
+    cache: redis.createClient(rconf.instanceUrl.cache),
+    limiter: redis.createClient(rconf.instanceUrl.limiter),
+    pub: redis.createClient(rconf.instanceUrl.pubsub),
+    realtimeLogging: redis.createClient(rconf.instanceUrl.realtimeLogging),
+    sampleStore: redis.createClient(rconf.instanceUrl.sampleStore),
+    sub,
```
shriramshankar: do we want to export the session redis client too?

iamigo: No, the redis session store constructor take a redis url, not a redis client

config.js
iamigo: I'm ok with giving choice to admin.

config.js
```diff
@@ -74,8 +75,28 @@ const deprioritizeJobsFrom = configUtil.csvToArray(pe.DEPRIORITIZE_JOBS_FROM);
...
 const JOB_QUEUE_TTL_SECONDS = pe.TTL_KUE_JOBS || DEFAULT_JOB_QUEUE_TTL_SECONDS;
 
-module.exports = {
+/*
+ * Assigns each of the different redis uses cases to a particular redis
+ * instance, if configured, or falls back to the primary redis instance.
+ */
+const redisUrls = {
+  cache: pe.REDIS_CACHE && pe[pe.REDIS_CACHE] ?
+    pe[pe.REDIS_CACHE] : (pe.REDIS_URL || DEFAULT_LOCAL_REDIS_URL),
```
pallavi2209: Can you give an example here, a bit hard to interpret.
So if we have pe.REDIS_URL (lets say x) set, the result will be x, else the result will be local redis url. But if pe.REDIS_CACHE is set (lets say y), do you expect the result to be y or x?

pallavi2209: Okay, got it, && has preference over ?
And to get configured cache working we need to have pe.REDIS_CACHE set (say x) and pe[x] set (say y)

PR #262, opened 2017-02-24
Populate samplestore from db

iamigo:


shriramshankar:
Thanks. I will do my changes branching off from this branch.

cache/sampleStore.js
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Populate the redis sample store from the db.
+ *
+ * @returns {Promise} which resolves to the list of redis batch responses, or
+ *  false if the feature is not enabled.
+ */
+function populate() {
+  if (!featureToggles.isFeatureEnabled(constants.featureName)) {
+    return Promise.resolve(false);
+  }
```
pallavi2209: Do you think we should include some console logs at different places to confirm what is happening in background, given we will be populating redis very rarely

cache/sampleStore.js
```diff
@@ -0,0 +1,252 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * ./cache/sampleStore.js
```
pallavi2209: some functions/variables in this file are used by most of other redis cache related files, but some are not (like eradicate/populate) when imported. Would it be better if we separate populate redis logic from redis config vars/functions ?

cache/sampleStore.js
```diff
@@ -0,0 +1,252 @@
...
+ * @returns {Object} the object with no nulls and stringified arrays.
+ */
+function removeNullsAndStringifyArrays(obj, arrayFields) {
+  Object.keys(obj).forEach((key) => {
+    if (arrayFields.includes(key)) {
+      obj[key] = JSON.stringify(obj[key]);
+    } else if (obj[key] === null) {
+      delete obj[key];
+    }
+  });
```
pallavi2209: If we dont want to keep null/empty arrays, then this should be changed to following:
```
 Object.keys(obj).forEach((key) => {
    // delete null fields and empty arrays, then stringify array fields
    if (obj[key] === null ||
     (Array.isArray(obj[key]) && obj[key].length === 0)) {
      delete obj[key];
    } else if (arrayFields.includes(key)) {
      obj[key] = JSON.stringify(obj[key]);
    }
  });
```

tests/cache/sampleStore.js
```diff
@@ -0,0 +1,144 @@
...
+      value: '5',
+      relatedLinks: [
+        { name: 'Salesforce', value: 'http://www.salesforce.com' },
+      ]
+    }))
+    .then(() => done())
+    .catch(done);
+  });
+
+  after(u.forceDelete);
```
pallavi2209: flush redis as well:
`    .then(() => redisClient.flushallAsync())`

cache/sampleStore.js
```diff
@@ -0,0 +1,252 @@
...
+
+    // Batch of commands to create each individal subject set...
+    const subjectCmds = Object.keys(subjectSets)
+      .map((key) => ['sadd', key, subjectSets[key]]);
+    batchPromises.push(redisClient.batch(subjectCmds).execAsync());
+
+    // Batch of commands to create each individal sample hash...
+    const sampleCmds = Object.keys(sampleHashes)
+      .map((key) => ['hmset', key, sampleHashes[key]]);
+    batchPromises.push(redisClient.batch(sampleCmds).execAsync());
```
pallavi2209: Just to confirm, since we are not using sorted sets, we will have to sort samples in API layer (default by name)?

cache/sampleStore.js
iamigo: If it's initiated via API call (Anny's story), then it will be logged via API activity logs, but I don't see any harm in logging it. I'll add that.

cache/sampleStore.js
iamigo: I like that. I'll separate that stuff out.

cache/sampleStore.js
```diff
@@ -0,0 +1,252 @@
...
+    sample: ['relatedLinks'],
+  },
+  indexKey: {
+    aspect: PFX + SEP + 'aspects',
+    sample: PFX + SEP + 'samples',
+    subject: PFX + SEP + 'subjects',
+  },
+  objectType: { aspect: 'aspect', sample: 'sample', subject: 'subject' },
+  prefix: PFX,
+  separator: SEP,
```
iamigo: Not sure what you mean... which commands?

cache/sampleStore.js
iamigo: Good point. Let's go ahead and use sorted sets--then we only have to sort in API layer if a custom sort order was requested.

iamigo:
Thx for all the well-thought-out comments, @pallavi2209!

cache/sampleStore.js
pallavi2209: Like this:
```
  commands: {
    hgetall: 'hgetall',
    hget: 'hget',
    smembers: 'smembers',
....
  },
```

cache/sampleStore.js
```diff
@@ -0,0 +1,103 @@
...
+      'criticalRange',
+      'warningRange',
+      'infoRange',
+      'okRange',
+    ],
+    sample: ['relatedLinks'],
+  },
+  indexKey: {
+    aspect: PFX + SEP + 'aspects',
+    sample: PFX + SEP + 'samples',
```
shriramshankar: didn't we decide on a different prefix for aspects, something like "refocache" ?

cache/sampleStore.js
iamigo: not sure why

cache/sampleStore.js
iamigo: i thought that sample store was better because it's easier to explain the different between this and our "normal" API response cache
...
easy enough to change it now...
what do you think?

cache/sampleStore.js
pallavi2209: Nevermind, I just thought we had sequelize commands defined in constants file, like $iLike etc, so we can define these strings too, but not a big deal.

cache/sampleStore.js
shriramshankar: I thought all the /aspects endpoints are going to hit this redis aspect store. So this will also act like the normal API cache.  
Not sure which way to go here.

tests/cache/sampleStore.js
```diff
@@ -0,0 +1,148 @@
...
+const samstoinit = require('../../cache/sampleStoreInit');
+const redisClient = require('../../cache/redisCache').client.sampleStore;
+const expect = require('chai').expect;
+const tu = require('../testUtils');
+const u = require('./utils');
+const Aspect = tu.db.Aspect;
+const Subject = tu.db.Subject;
+const Sample = tu.db.Sample;
+
+describe('sampleStore (feature off):', () => {
```
pallavi2209: This test fails if redis feature is enabled in environment. Maybe we should add before after like this:
  before((done) => {
    tu.toggleOverride(sampleStore.constants.featureName, false);
    done();
  });

  after((done) => {
    tu.toggleOverride(sampleStore.constants.featureName, true);
    done();
  });

cache/sampleStoreInit.js
```diff
@@ -0,0 +1,178 @@
...
+      if (subjectSets.hasOwnProperty(subKey)) {
+        subjectSets[subKey].push(aspKey);
+      } else {
+        subjectSets[subKey] = [aspKey];
+      }
+
+      // For creating each individual aspect hash...
+      if (!aspectHashes.hasOwnProperty(aspKey)) {
+        aspectHashes[aspKey] = samsto.cleanAspect(s.aspect);
+      }
```
shriramshankar: Looks like the "aspect hashes" are created out of aspects from the sample object. I think we had decided to create the "aspects hashes" direct out the the aspect table.

iamigo: OK, I'll change that.

iamigo: Updated @shriramshankar

iamigo:
updated test to set/reset initial feature enablement state @pallavi2209

cache/sampleStoreInit.js
```diff
@@ -0,0 +1,197 @@
...
+        nameParts[0]); // eslint-disable-line no-magic-numbers
+
+      // Track each of these in the master indexes for each object type.
+      sampleIdx.add(samKey);
+      subjectIdx.add(subKey);
+
+      // For creating each individual subject set...
+      if (subjectSets.hasOwnProperty(subKey)) {
+        subjectSets[subKey].push(aspKey);
+      } else {
```
shriramshankar: just noticed this, while I was running my tests. Considering that we can easily convert an aspect to an aspectKey but not the other way round, the aspect name should be pushed into the subjectSet instead of aspectKey. Correct?

iamigo: We can store the aspect's redis key, the aspect name, the sample name, the sample's redis key... they're all interchangeable.
The most common operation for this will be the read to get the hierarchy, since the write will typically only happen when the sample is initially created (or if the aspect name changes), so if we wanted to optimize for reconstructing the hierarchy, wouldn't you want it to store the sample's redis key?

shriramshankar: We need to do some string manipulation if we store one of these anyways. Since we already have a toKey function I thought we can use it. If you want to leave it the way it is, I can write something for the hierarchy.

iamigo: There's string manipulation on one or both side of the equation no matter what so let's get it right for the hierarchy use case. With no filtering, would make sense to use the sample's redis key. But with filtering, it might make most sense to use aspect name. Which is what you said initially. I'll make that change now.

shriramshankar: ok

annyhe:
is there a reason why only one travis test ran? or is this a minor optimization?

iamigo:
@annyhe I think that was just a byproduct of yesterday's travis failures

PR #263, opened 2017-02-24
tests show GET /v1/tokens/tokenID returns createdBy and user name

annyhe:


pallavi2209:
Do we need to check only GET method, or all the token endpoints?

annyhe:
@pallavi2209 The story is only for the GET method

PR #264, opened 2017-02-27
Redis: Sample: get endpoints without filters

pallavi2209:


pallavi2209:
Base branched from #262. Should be merged after that is merged.

api/v1/controllers/samples.js
```diff
@@ -62,7 +150,35 @@ module.exports = {
...
+        const sampleRes = cleanAddAspectToSample(
+          responses[ZERO], responses[ONE], res.method
+        );
+
+        res.status(httpStatus.OK).json(sampleRes);
+      });
+    } else {
+      doGet(req, res, next, helper);
+    }
   },
```
shriramshankar: the redis related changes need to go in its own file, under cache/nouns/samples

cache/sampleStore.js
```diff
@@ -81,10 +103,12 @@ function eradicate() {
  */
 function removeNullsAndStringifyArrays(obj, arrayFields) {
   Object.keys(obj).forEach((key) => {
-    if (arrayFields.includes(key)) {
-      obj[key] = JSON.stringify(obj[key]);
-    } else if (obj[key] === null) {
+    // delete null fields and empty arrays, then stringify array fields
+    if (obj[key] === null ||
+     (Array.isArray(obj[key]) && obj[key].length === 0)) {
```
shriramshankar: deleting empty arrays would be breaking the current API contract. Anything that is not a null is allowed.

iamigo: Just pushed up a commit for this

cache/sampleStore.js
```diff
@@ -38,6 +38,11 @@ const constants = {
   objectType: { aspect: 'aspect', sample: 'sample', subject: 'subject' },
   prefix: PFX,
   separator: SEP,
+  commands: {
+    hgetall: 'hgetall',
+    hget: 'hget',
+    smembers: 'smembers',
+  },
 };
```
shriramshankar: Unless you are creating your own wrapper for the redis commands, there is little value exporting them.

api/v1/controllers/samples.js
pallavi2209: nouns or verbs? In verbs, add a check for toggle and sample model?

cache/sampleStore.js
pallavi2209: I guess it should be fine. I am mostly worried about typos. http://rc3.org/2006/10/11/use-contants-instead-of-string-literals/

api/v1/controllers/samples.js
shriramshankar: I am using nouns for now, for the lack of a better word to use. Maybe verb" is better but the more changes I am doing now I am leaning towards "models". 
@iamigo Any suggestions?

iamigo: "models" works for me

pallavi2209: Got it!

cache/models/samples.js
```diff
@@ -0,0 +1,155 @@
+/**
```
iamigo: Lots of API-level stuff in here... I guess I was imagining this more as a replacement for sequelize commands, so the API layer would fork either to execute sequelize cmds or come here for redis cmds... but I guess that would require a larger refactoring or our API layer.
Thoughts?

pallavi2209: I can try keeping the api level stuff like logging, extracting info from req object and response in api layer only. And then just do the redis transactions here and return the result object from here.

iamigo: Thx!

cache/models/samples.js
```diff
@@ -0,0 +1,155 @@
...
+    // get aspect
+    commands.push([
+      'hgetall',
+      sampleStore.toKey(constants.objectType.aspect, aspectName),
+    ]);
+
+    redisClient.batch(commands).execAsync()
+    .then((responses) => {
+      resultObj.dbTime = new Date() - resultObj.reqStartTime; // log db time
+
```
shriramshankar: dbTime is kind of confusing here. Wondering if we want to do some logging just for the redis layer?

cache/sampleStore.js
```diff
@@ -49,6 +49,23 @@ function toKey(type, name) {
...
 
 /**
+ * Convert array strings to json from redis object
+ * @param  {Object} obj - Object to convert
+ * @param  {Object} arrayFields - List of array fields which were stringified.
+ * @returns {Object} - Converted object
+ */
+function arrayStringsToJson(obj, arrayFields) {
+  Object.keys(obj).forEach((key) => {
+    // Some array fields might have already been parsed in previous calls
```
shriramshankar: iterating over the array instead of object keys is probably much faster.

cache/models/samples.js
```diff
@@ -0,0 +1,155 @@
...
+    .then((sampleAndAspects) => {
+      resultObj.dbTime = new Date() - resultObj.reqStartTime; // log db time
+      const samples = sampleAndAspects[ZERO];
+      const aspects = sampleAndAspects[ONE];
+
+      samples.forEach((sampleObj) => {
+        const sampleAspect = aspects.find((aspect) =>
+          aspect.name === sampleObj.name.split('|')[ONE]
+        );
+
```
shriramshankar: this take O(n^2) correct?

cache/models/samples.js
```diff
@@ -0,0 +1,155 @@
...
+      // get all aspect names
+      return redisClient.smembersAsync(constants.indexKey.aspect);
+    })
+    .then((allAspectKeys) => {
+      // add to commands to get aspect
+      allAspectKeys.forEach((aspectName) => {
+        aspectCmds.push(
+          ['hgetall', aspectName.toLowerCase()]
+        );
```
shriramshankar: aspectName will already be in lowercase  since you are getting it from redis

cache/models/samples.js
iamigo: I think I prefer to still call our retrieval from redis "dbTime" rather than introduce a new field.

cache/models/samples.js
pallavi2209: Yes. Can we do better?
Other approach was going to redis for each sample to get aspect. Given aspects numbers are way less than samples, I voted to get all aspects at once and save trips to redis.

pallavi2209:
Thanks for the comments. Planning to work on comments once Ian's changes are merged.

cache/models/samples.js
shriramshankar: Talked to Ian about something similar yesterday.  I make redis gets for each of the samples and aspects, not sure which is better.

pallavi2209: I guess for sample endpoints, it would not matter much because its not used that often as compared to hierarchy endpoint.

tests/cache/models/redisTestUtil.js
```diff
@@ -0,0 +1,106 @@
...
+    tu.forceDelete(tu.db.Sample, testStartTime)
+    .then(() => tu.forceDelete(tu.db.Aspect, testStartTime))
+    .then(() => tu.forceDelete(tu.db.Subject, testStartTime))
+    .then(() => tu.forceDelete(tu.db.Profile, testStartTime))
+    .then(() => tu.forceDelete(tu.db.User, testStartTime))
+    .then(() => redisClient.flushallAsync())
+    .then(() => done())
+    .catch(done);
+  },
+};
```
shriramshankar: I will use this too. Very helpful.

iamigo: me too, once it's merged :)

cache/sampleStore.js
```diff
@@ -49,6 +49,23 @@ function toKey(type, name) {
 } // toKey
 
 /**
+ * Convert array strings to json from redis object
+ * @param  {Object} obj - Object to convert
+ * @param  {Object} arrayFields - List of array fields which were stringified.
+ * @returns {Object} - Converted object
+ */
+function arrayStringsToJson(obj, arrayFields) {
```
shriramshankar: I can use this too but in my next PR.

pallavi2209:
@shriramshankar @iamigo updated PR.

pallavi2209:
Travis builds passed. Merging this now.

PR #265, opened 2017-02-28
fixed aspecttags filter issue

harshkothari410:


harshkothari410:
Where should I write test for this @iamigo ? I am confused.

iamigo:
Good question, @harshkothari410. I don't have a good answer for this, since we have not prioritized any selenium-like testing.

harshkothari410:
I guess this thing is using Hierarchy end points and it is covered. So let me know if test is necessary or not because I cannot think about the place where I can write test for this. If we dont want test then I will go forward for another bug. And you guys can merge it. @iamigo @shriramshankar

PR #266, opened 2017-02-28
Persist samplestore to db

iamigo:
Define env var PERSIST_REDIS_SAMPLE_STORE_MILLISECONDS.
Define a new clock job using PERSIST_REDIS_SAMPLE_STORE_MILLISECONDS. If ENABLE_REDIS_SAMPLE_STORE env var is true, this new job will truncate the db samples table and insert all the samples from the redis sample store into the table.


cache/sampleStorePersist.js
```diff
@@ -0,0 +1,55 @@
...
+    return Promise.resolve(false);
+  }
+
+  return Sample.destroy({ truncate: true, force: true })
+  .then(() => redisClient.smembersAsync(constants.indexKey.sample))
+  .then((keys) => keys.map((key) => ['hgetall', key]))
+  .then((cmds) => redisClient.batch(cmds).execAsync())
+  .then((res) => {
+    const samplesToCreate = res.map((sample) => {
+      sample.relatedLinks = JSON.parse(sample.relatedLinks);
```
pallavi2209: Maybe use arrayStringsToJson function (in #264) so that if we happen to add new array field to samples in future, we wont have to change here.

tests/cache/persist.js
```diff
@@ -0,0 +1,130 @@
...
+  after((done) => {
+    u.forceDelete(done)
+    .then(() => redisClient.flushallAsync())
+    .then(() => tu.toggleOverride(sampleStore.constants.featureName,
+      initialFeatureState))
+    .then(() => done())
+    .catch(done);
+  });
+
+  it('ok', (done) => {
```
pallavi2209: Do you think this will be better test: After populating redis, delete/change a sample from postgres db, then persist and see if we get all samples back. I am assuming that is the purpose of db persist?

iamigo: @pallavi2209 i updated the test as suggested

PR #268, opened 2017-03-01
changes to assemble the complete subject hierarchy from the cache layer

shriramshankar:
We need to merge PR  #262  and #264 before this. I have checked in @pallavi2209's redisTestUtil.js file too for the tests to pass.  I have also done some refracting for the cache layer and the DB layer to be able to use the same set of filters.

### DIFF
--- a/cache/models/subject.js
+++ b/cache/models/subject.js
@@ -1,0 +1,228 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * cache/model/subject.js
+ */
+'use strict'; // eslint-disable-line strict
+
+const sampleStore = require('../sampleStore');
+const constants = sampleStore.constants;
+const redisClient = require('../redisCache').client.sampleStore;
+const u = require('../../utils/filters');
+const ONE = 1;
+const TWO = 2;
+
+// using the bluebird promise implementation instead of the native promise
+const Promise = require('bluebird');
+
+/*
+ * All the query params that can be expected in the hierarchy endpoint are
+ * defined as a key. The values of the query parameters are later added to them
+ * by the set filters function
+ */
+const filters = {
+  aspect: {},
+  subjectTags: {},
+  aspectTags: {},
+  status: {},
+};
+
+/**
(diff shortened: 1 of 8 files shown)


tests/cache/models/subjects/getHierarchy.js
```diff
@@ -0,0 +1,211 @@
...
+      .set('Authorization', token)
+      .expect(constants.httpStatus.OK)
+      .expect((res) => {
+        expect(res.body.children[0].children[0].samples).to.be
+          .an('array');
+        console.log(res.body);
+        expect(res.body.children[0].children[0].samples).to.have.lengthOf(1);
+        expect(res.body.children[0].children[0].samples[0].aspect.rank).
+          to.equal(10);
+      })
```
shriramshankar: need to talk about this. returns '10' instead of 10.

cache/models/subject.js
```diff
@@ -0,0 +1,228 @@
...
+
+/**
+ * Given a subject with its samples, aspect, aspectTags and sampleStatus filters
+ * are applied to the samples and the filtered samples are attached back to the
+ * subject
+ * @param  {Object} subj - Subject object on which the filters are applied
+ * @returns {Object} - filtered subject
+ */
+function filterSamples(subj) {
+  const filteredSamples = [];
```
shriramshankar: @iamigo @pallavi2209  I moved the aspect filters along with the other filters. Applying just the aspect includes filter while getting it from redis (using "hmget aspectnames")  will work but adding the excludes filter will nest the promises and break the promise chain. Any thoughts about this?

cache/models/subject.js
```diff
@@ -0,0 +1,228 @@
...
+  return redisClient.smembersAsync(subjectKey).then((aspectNames) => {
+    // console.log(1);
+    const sampleAspectPromise = [];
+    aspectNames.forEach((aspect) => {
+      const sampleName = res.absolutePath + '|' + aspect;
+      sampleAspectPromise.push(getSampleFromRedis(sampleName));
+      sampleAspectPromise.push(getAspectFromRedis(aspect));
+    });
+
+    return Promise.all(sampleAspectPromise);
```
pallavi2209: We can do this in one redis call using commands array as we discussed.

shriramshankar:
@iamigo or @pallavi2209 Can one of you merge this PR? I think I skipped a patch that I should not have while rebasing and my workspace is kind of messed up now. I will create another PR with the changes in the comments.

pallavi2209:
I am fine merging this if @iamigo approves.

PR #272, opened 2017-03-03
Sample store: Redis: add filters to get samples endpoints

pallavi2209:


api/v1/controllers/samples.js
```diff
@@ -53,7 +53,9 @@ module.exports = {
   findSamples(req, res, next) {
```
iamigo: So theoretically we should be able to run *all* the api sample tests with the sampleStore feature flag enabled and everything should work?

pallavi2209: Yes, let me actually add all those here as well. More tests never hurt.

cache/models/samples.js
```diff
@@ -47,92 +127,246 @@ function cleanAddAspectToSample(sampleObj, aspectObj, method) {
...
+    const sampAspArr = sampleName.split('|');
     const commands = [];
 
-    // get sample
+    if (sampAspArr.length < TWO) {
+      throw new redisErrors.ResourceNotFoundError({
+        explanation: 'Incorrect sample name.',
+      });
+    }
+
```
shriramshankar: looks like a validation that can be done in the API layer.

tests/cache/models/samples/get.js
```diff
@@ -109,4 +243,73 @@ describe(`api::redisEnabled::GET ${path}`, () => {
...
+    .expect(constants.httpStatus.BAD_REQUEST)
+    .end((err, res) => {
+      if (err) {
+        return done(err);
+      }
+
+      return done();
+    });
+  });
+});
```
shriramshankar: new line at EOF

cache/models/samples.js
pallavi2209: I thought we are keeping only API related things in api layer, rest of the logic goes in redis layer. Plus, splitting in api layer would mean passing sample name and aspect name or split again in redis layer.

pallavi2209:

cache/models/samples.js
shriramshankar: I think we are moving the validations too. @iamigo Will your sample validation story take care of this in the API layer?

iamigo: The only validation piece I was going to be adding in the API layer is this: 
Check for uniqueness of relatedLinks names within a sample. Do this on sample POST/PATCH/PUT when relatedLinks not empty. Reject with validation error if a relatedLink name is reused within the same sample.

cache/models/samples.js
```diff
@@ -47,92 +127,250 @@ function cleanAddAspectToSample(sampleObj, aspectObj, method) {
...
+  if (!opts.order) {
+    resArr = applyLimitAndOffset(opts, sampKeysArr);
+  }
+
+  // apply wildcard expr on name, if specified
+  if (opts.filter && opts.filter.name) {
+    const filteredKeys = filterByFieldWildCardExpr(
+      resArr, 'name', opts.filter.name
+    );
+    console.log(filteredKeys);
```
shriramshankar: console logs

cache/models/samples.js
```diff
@@ -16,12 +16,92 @@ const u = require('../../api/v1/helpers/verbs/utils');
...
+    }
+
+    if (strA > strB) {
+      return isDescending ? MINUS_ONE : ONE;
+    }
+
+    return ZERO;
+  });
+}
+
```
shriramshankar: this is nicely done.

pallavi2209:
@shriramshankar Removed console log. Updated the PR.

PR #277, opened 2017-03-06
Redis: Sample upsert and bulk upsert endpoints

pallavi2209:


api/v1/controllers/samples.js
```diff
@@ -159,25 +159,46 @@ module.exports = {
...
 
-      // loop through remove values to delete property
-      if (helper.fieldsToExclude) {
-        u.removeFieldsFromResponse(helper.fieldsToExclude, o.dataValues);
-      }
+      redisModelSample.upsertSampleInRedis(
+        sampleQueryBody, resultObj, res.method
+      )
+      .then((sampleRes) => {
+        resultObj.dbTime = new Date() - resultObj.reqStartTime;
```
iamigo: lines 170-180 and lines 189-200 are almost the same... I don't want you to spend more than a couple of minutes thinking about this, but if it's easy to do, can the duplicate-ish code be merged?

api/v1/controllers/samples.js
```diff
@@ -196,25 +217,43 @@ module.exports = {
...
 
         const j = jobWrapper.createJob(jobType.BULKUPSERTSAMPLES,
           wrappedBulkUpsertData, req);
       } else {
-        helper.model.bulkUpsertByName(value,
-          userName);
+        redisModelSample.bulkUpsertSampleInRedis(value);
       }
-    });
+    } else {
```
iamigo: Same comment here as above... if we're able to easily collapse the near duplicate code into one code block, it's less surface area to hide a bug and fewer lines to test and fewer lines of code to maintain

cache/models/samples.js
```diff
@@ -250,6 +267,142 @@ function applyFiltersOnSampObjs(sampObjArray, opts) {
...
+ * else creates a new one.
+ * @param  {Object} sampleQueryBodyObj - Query Body Object for a sample
+ * @returns {Array} - Array of redis commands responses
+ */
+function upsertOneSample(sampleQueryBodyObj) {
+  const sampleName = sampleQueryBodyObj.name;
+  const subjAspArr = sampleName.toLowerCase().split('|');
+  if (subjAspArr.length < TWO) {
+    throw new redisErrors.ResourceNotFoundError({
+      explanation: 'Incorrect sample name.',
```
iamigo: If we're throwing ResourceNotFoundError the explanation should be something like 'no sample was found with this name' but here it feels more like an "Invalid..." error because a sample name without a pipe character can never be valid, can never be "found".

cache/models/samples.js
```diff
@@ -250,6 +267,142 @@ function applyFiltersOnSampObjs(sampObjArray, opts) {
...
+  commands.push([
+    'hgetall',
+    sampleStore.toKey(constants.objectType.aspect, aspectName),
+  ]);
+  commands.push([
+    'hgetall',
+    sampleStore.toKey(constants.objectType.sample, sampleName),
+  ]);
+
+  return redisClient.batch(commands).execAsync()
```
iamigo: I think here we will end up sending all three commands, but if we're throwing an error if the first comes back false, then why would we want to issue the other commands? And same for second?

cache/models/samples.js
```diff
@@ -375,4 +528,49 @@ module.exports = {
...
   },
+
+  /**
+   * Upsert sample in Redis.
+   * @param  {Object} qbObj - Query body object
+   * @param  {Object} logObject - Log object
+   * @param  {String} method - Type of request method
+   * @returns {Promise} - Resolves to upserted sample
+   */
+  upsertSampleInRedis(qbObj, logObject, method) {
```
iamigo: I think I prefer naming these functions *without* the "...InRedis" suffix--the fact that they're in *this* "cache" module and not the "db" module says that already, and without the suffix it sets up an equivalence/interchangeability between the db functions and the redis functions which I like

cache/sampleStore.js
```diff
@@ -68,7 +68,7 @@ function getNameFromKey(key) {
  */
 function arrayStringsToJson(obj, arrayFields) {
   arrayFields.forEach((field) => {
-    if (obj[field] && !Array.isArray(obj[field])) {
+    if (obj && obj[field] && !Array.isArray(obj[field])) {
```
iamigo: good catch, thx!

cache/sampleStoreInit.js
```diff
@@ -46,7 +47,7 @@ function eradicate() {
  * @returns {Promise} which resolves to the list of redis batch responses.
  */
 function populateAspects() {
-  return Aspect.findAll()
+  return Aspect.findAll({ where: { isPublished: true } })
```
iamigo: thx!

cache/sampleStoreInit.js
```diff
@@ -63,6 +64,25 @@ function populateAspects() {
...
+    const cmds = [];
+    subjects.forEach((s) => {
+      const key = samsto.toKey(constants.objectType.subject, s.absolutePath);
+      cmds.push(['sadd', constants.indexKey.subject, key]);
+    });
+
+    return redisClient.batch(cmds).execAsync();
+  })
+  .catch(console.error); // eslint-disable-line no-console
+} // populateAspects
```
iamigo: // populateSubjects

cache/sampleStoreInit.js
```diff
@@ -137,7 +156,7 @@ function populate() {
   const msg = 'Populating redis sample store from db';
   console.log(msg); // eslint-disable-line no-console
 
-  const promises = [populateSamples(), populateAspects()];
+  const promises = [populateSamples(), populateAspects(), populateSubjects()];
```
iamigo: This doesn't matter at an implementation/execution level here, but I think subjects and aspects should be listed in this array *before* samples--it (very subtly) helps communicate that subjects and aspects have to exist for a sample to exist

cache/models/samples.js
```diff
@@ -250,6 +267,142 @@ function applyFiltersOnSampObjs(sampObjArray, opts) {
...
+    const status = sampleUtils.computeStatus(aspectObj, value);
+
+    if (!sampObj || (sampObj && (sampObj[sampFields.PRVS_STATUS] !== status))) {
+      const prevStatus = sampObj ? sampObj[sampFields.STATUS] :
+                            dbConstants.statuses.Invalid;
+      props.push(sampFields.PRVS_STATUS, prevStatus);
+      props.push(sampFields.STS_CHNGED_AT, new Date().toString());
+      props.push(sampFields.STATUS, status);
+    }
+  }
```
pallavi2209: @shriramshankar @iamigo Please take a look at this logic to confirm that we update/create sample fields correctly.

iamigo: looks right to me

shriramshankar: Any reason why you are using the props array? You can update the qbObj with the new fields and pass it to the hmset command.

pallavi2209: @shriramshankar No reason, just realized we can pass an object too. Lets do it then. Thanks! :)

api/v1/controllers/samples.js
```diff
@@ -196,25 +217,43 @@ module.exports = {
...
-        wrappedBulkUpsertData.userName = userName;
         wrappedBulkUpsertData.reqStartTime = reqStartTime;
+        wrappedBulkUpsertData.isRedisEnabled = true;
 
         const j = jobWrapper.createJob(jobType.BULKUPSERTSAMPLES,
           wrappedBulkUpsertData, req);
       } else {
-        helper.model.bulkUpsertByName(value,
-          userName);
+        redisModelSample.bulkUpsertSampleInRedis(value);
```
shriramshankar: Can we change this to return the list of samples, to be used by the real-time publisher?

pallavi2209:
@shriramshankar @iamigo Updated the PR.

pallavi2209:
@iamigo @shriramshankar updated the PR. Some changes since you last reviewed:
1. Returning samples from bulk upsert method now, Shriram's comment
2. Separated redis batch commands based on Ian's comment
3. Removed props and updated queryObject to upsert, based on Shrirams comments
4. Added tests for bulk upsert job Queue
5. Some refactoring based on comments

PR #279, opened 2017-03-08
Redis: Sample timeout

pallavi2209:


cache/sampleStoreTimeout.js
```diff
@@ -0,0 +1,94 @@
+/**
```
iamigo: Just curious, in your testing, did you get a sense of how long this takes with the sampleStore feature on vs. off (with a dataset close to our current prod dataset)?

cache/sampleStoreTimeout.js
```diff
@@ -0,0 +1,94 @@
...
+      redisClient.smembersAsync(sampleStore.constants.indexKey.sample)
+      .then((allSamples) => {
+        const commands = [];
+        const aspectType = sampleStore.constants.objectType.aspect;
+
+        allSamples.forEach((sampKey) => {
+          const aspectName = sampKey.split('|')[ONE];
+          commands.push(['hgetall', sampKey]); // get sample
+          commands.push(
+            ['hgetall', sampleStore.toKey(aspectType, aspectName)]
```
iamigo: I think this is going to be sending tons of duplicate requests for the same aspect. How about adding each sample's aspect to a Set first, to consolidate it down to a list of unique aspects?

cache/sampleStoreTimeout.js
```diff
@@ -0,0 +1,94 @@
...
+          );
+        });
+
+        return redisClient.batch(commands).execAsync();
+      })
+      .then((redisResponses) => {
+        const samples = [];
+        const aspects = [];
```
iamigo: ... then down here load the aspects into a hash instead of an array so you can look up each one directly

cache/sampleStoreTimeout.js
```diff
@@ -0,0 +1,94 @@
...
+              objToUpdate,
+            ]);
+            numberTimedOut++;
+          }
+        }
+
+        numberEvaluated = samples.length;
+        return redisClient.batch(sampCmds).execAsync();
+      })
+      .then(() => resolve({ numberEvaluated, numberTimedOut }))
```
iamigo: I'm not sure you need the "resolve()" here, can't you just return the object, e.g.

`.then(() => ({ numberEvaluated, numberTimedOut }))`

cache/sampleStoreTimeout.js
pallavi2209: Right, let me try and optimize this given it will be executing constantly.

cache/sampleStoreTimeout.js
pallavi2209: Let me optimize and measure times then.

cache/sampleStoreTimeout.js
pallavi2209: Even if the code is included in new Promise(resolve, reject)...? Have not tested it but never done that before.

iamigo: Line 38: `return new Promise((resolve, reject) => {`

but your call to redisClient.smembersAsync is already returning a promise, so just return that, i.e.

return redisClient.smembersAsync(sampleStore.constants.indexKey.sample) ...

cache/sampleStoreTimeout.js
```diff
@@ -34,61 +94,46 @@ module.exports = {
...
+      return redisClient.batch(commands).execAsync();
+    })
+    .then((redisResponses) => {
+      const sampCmds = getSampleTimeoutCommands(
+      redisResponses, samplesCount, curr);
+      numberEvaluated = samplesCount;
+      numberTimedOut = sampCmds.length;
+      return redisClient.batch(sampCmds).execAsync();
+    })
+    .then(() => {
```
iamigo:
```
 +    .then(() => {
 +      const res = { numberEvaluated, numberTimedOut };
 +      return res;
 +    })
```
can just be

```+    .then(() => ({ numberEvaluated, numberTimedOut }))```

File: cache/sampleStoreTimeout.js [hunk 6]
```diff
@@ -34,61 +94,46 @@ module.exports = {
...
+      const res = { numberEvaluated, numberTimedOut };
+      return res;
+    })
+    .catch((err) => {
```
iamigo: if we're not doing anything but just throwing the same err, then we shouldn't need that catch block at all here

pallavi2209:
@iamigo @shriramshankar updated the PR. Without bulk upsert happening and all samples already timeout, it take ~1sec with redis feature on and ~2 sec with redis feature off (~11k samples being evaluated). Still need to get results with bulk upserts happening. When doing that, found a bug in bulk upserts, for which I will be creating another PR.

cache/sampleStoreTimeout.js
```diff
@@ -0,0 +1,141 @@
...
+  for (let num = 0; num < samples.length; num++) {
+    const samp = samples[num];
+    const aspName = samp.name.split('|')[ONE];
+
+    if (!aspName) {
+      throw new redisErrors.ResourceNotFoundError({
+        explanation: 'Aspect not found.',
+      });
+    }
+
```
shriramshankar: line 40 to 43 will never be executed. Line 37 will throw an "array out of bounds error" if aspect is not found.

pallavi2209: When I just tried this in my console: `'abc'.split('|')[1]` and it returns undefined. Do you mean something else? I can still have a check though.

PR #284, opened 2017-03-09
added validation to handle the hierarchy error

shriramshankar:


db/model/subject.js
```diff
@@ -345,8 +349,32 @@ module.exports = function subject(seq, dataTypes) {
...
+        /*
+         * If the depth of the subject does not match the hierarchyLevel
+         * throw an error
+         */
+        if (subjectDepth !== inst.hierarchyLevel) {
+          throw new InvalidHierarchyLevel('The updated depth of ' +
+            `${inst.absolutePath}: ${subjectDepth}, does not match the ` +
+             `updated hierarchyLevel: ${inst.hierarchyLevel}`);
+        }
+
```
shriramshankar: @iamigo @pallavi2209 This check fails in Travis wherever I put this. I also tried to write a sequelize raw query to throw an error when a root subject has an entry in the Subjectsancestors table. This validation does not run locally. Any other suggestions on how to handle it? Also, I have a validation on the API layer to prevent any changes to the hierarchyLevel field, with the API level validation in place, do we need the DB validation too?

iamigo: I'm fine with just using the "read only fields" logic you added in the API layer.

In fact, can you add a user story to extend that paradigm to other models which have read-only fields? e.g. sample "status", "previousStatus", "statusLastChangedAt"...

shriramshankar: Ok. I have made one final change to see if the validation will run in travis. If it does not run I will take it out from the db layer.

shriramshankar:
@iamigo PR updated with just the API level validation.

db/model/subject.js
```diff
@@ -210,10 +209,15 @@ module.exports = function subject(seq, dataTypes) {
       },
     },
     hooks: {
+
       /**
        * BeforeCreate hook
        * set up appropriate parentAbsolutePath, absolutePath and ParentId
-       *
+       * If parentAsbilutePath or parentId is provided while creating a subject,
```
iamigo: sp parentAsbolutePath

db/model/subject.js
```diff
@@ -488,7 +491,13 @@ module.exports = function subject(seq, dataTypes) {
        *  change, otherwise returns a Promise which resolves to undefined, or
        * rejects if an error was encountered
        */
-      beforeUpdate(inst, opts) {
+      beforeUpdate(inst /* ,  opts */) { // eslint-disable-line max-statements
+
+        /*
+         * If a subject is getting unpublished, check to see if its childeren are
```
iamigo: sp children

PR #285, opened 2017-03-09
Redis: upsert: Correct promises flow

pallavi2209:
Previously: When checking for subject and aspect to create/update sample, we were throwing error if not found, but control did not exit out of promise (which I expected). This leads to creation of samples even if subject/aspect is not present.
Now, with Promise.all(), if any promise fails, no further execution and error is returned.

Added additional bulk upsert tests to test the responses from bulk upsert method.


cache/models/samples.js
```diff
@@ -368,70 +389,54 @@ function upsertOneSample(sampleQueryBodyObj, method) {
...
+
+    // sampleQueryBodyObj updated with fields
+    createSampHsetCommand(sampleQueryBodyObj, sample, aspectObj);
+
+    // add aspect name to subject set, add sample key to sample set,
+    // create/update hash of sample
+    return redisClient.batch([
+      ['sadd', subjKey, aspectName],
+      ['sadd', constants.indexKey.sample, sampleKey],
+      ['hmset', sampleKey, sampleQueryBodyObj],
```
pallavi2209: Just to optimize, I think we should execute only hmset if sample is already present, let me make that change as well.

cache/models/samples.js
```diff
@@ -586,7 +596,7 @@ module.exports = {
    */
   bulkUpsertSample(sampleQueryBody) {
     const promises = sampleQueryBody.map(
-      (sampleReq) => upsertOneSample(sampleReq)
+      (sampleReq) => upsertOneSample(sampleReq, null, true)
     );
```
shriramshankar: While you are making these changes, can you add subjectAbsolutePath to the samples returned? I need it to attach the subject to the sample object.

shriramshankar: Sorry. I can extract it from the sample name. I do not need it as a seperate field.

pallavi2209: I needed to pass isBulk to customize response based on caller method (null is request method to create api links which we don not want in bulk upsert). I did this to match existing results, but I feel we can improve on that. In case of bulk upsert, we are sending this object {isFailed: true, Explanation: [errobject] }, and in simple upsert we just send err object.
I think we should change this behaviour and return error object directly in both cases and check error in caller function (bulk upsert) by using `instanceof Error `
If we do this, we will be changing in non-redis implementation as well (which I didnt wanna touch with this PR)

pallavi2209:
@iamigo @shriramshankar Done with changes now, sorry for assigning too soon, later found out few more things worth changing change.
I am able to test bulk upsert with time out now.
With Redis feature off: Timeout take 2 to 18 minutes for 812 samples to timeout. Log Example:

> Mar 09 19:43:21 test-refocus-pallavi app/clock.1:  info: activity=worker dbTime=13816ms errorCount=13194 ipAddress=None jobId=65758 jobType=SAMPLE_TIMEOUT queueTime=29ms recordCount=821 token=None totalTime=13834ms user=None workTime=13818ms 

With Redis feature on: Timeout takes <1 to 3 minutes for 812 samples to timeout. Log Example:

> Mar 09 18:28:17 test-refocus-pallavi app/clock.1:  info: activity=worker dbTime=2294ms errorCount=12754 ipAddress=None jobId=65587 jobType=SAMPLE_TIMEOUT queueTime=14ms recordCount=821 token=None totalTime=2319ms user=None workTime=2296ms

Its just that I dont understand why then number of errorCount is different, which was true even if all the samples are timedout already. There was a difference of exactly 440 samples in both cases.

iamigo:
yeah it's a weird use of the "errorCount" field
but the timings are in line with expectations and very good

cache/models/samples.js
shriramshankar: While returning the samples, is there a way to indicate if the sample was created or updated?

shriramshankar:
@annyhe Looks like Ian already reviewed this PR. We can merge this PR now and assign any future PR for you to review.

tests/cache/models/samples/upsertBulk.js
```diff
@@ -114,6 +148,27 @@ describe('api::redisEnabled::POST::bulkUpsert ' + path, () => {
...
+      {
+        name: `${tu.namePrefix}NOT_EXIST|${tu.namePrefix}Aspect1`,
+        value: '2',
+      }, {
+        name: `${tu.namePrefix}Subject|${tu.namePrefix}Aspect2`,
+        value: '4',
+      },
+    ])
+    .then((response) => {
+      expect(response[0].isFailed).to.be.true;
```
annyhe: did you mean expect(response[1].isFailed).to.be.false?

tests/cache/models/samples/upsertBulk.js
```diff
@@ -136,6 +191,23 @@ describe('api::redisEnabled::POST::bulkUpsert ' + path, () => {
...
+      {
+        name: `${tu.namePrefix}NOT_EXIST|${tu.namePrefix}Aspect1`,
+        value: '2',
+      }, {
+        name: `${tu.namePrefix}Subject|${tu.namePrefix}NOT_EXIST`,
+        value: '4',
+      },
+    ])
+    .then((response) => {
+      expect(response[1].isFailed).to.be.true;
```
annyhe: did you want to check the explanation field?

cache/models/samples.js
annyhe: does this mean you want to remove the isFailed field from bulk upsert?

api/v1/controllers/samples.js
```diff
@@ -88,7 +88,7 @@ module.exports = {
       const sampleName = req.swagger.params.key.value.toLowerCase();
 
       redisModelSample.getSample(
-        sampleName, resultObj, res.method, req.swagger.params
+        sampleName, resultObj, req.method, req.swagger.params
```
annyhe: what is the reason behind changing the response.method to request.method? Are they identical?

cache/models/samples.js
```diff
@@ -368,70 +393,59 @@ function upsertOneSample(sampleQueryBodyObj, method) {
...
+  })
+  .then(() => redisClient.hgetallAsync(
+      sampleStore.toKey(constants.objectType.sample, sampleName)
+  ))
+  .then((updatedSamp) =>
+   cleanAddAspectToSample(updatedSamp, aspectObj, method)
+  )
+  .catch((err) => {
+    if (isBulk) {
+      return err;
```
annyhe: why are we returning the error for bulk upsert, instead of throwing it?

PR #290, opened 2017-03-15
Sample events to api

shriramshankar:


api/v1/controllers/samples.js
```diff
@@ -226,10 +234,23 @@ module.exports = {
 
         const j = jobWrapper.createJob(jobType.BULKUPSERTSAMPLES,
           wrappedBulkUpsertData, req);
-      } else if (featureToggles.isFeatureEnabled(constants.featureName)) {
-        redisModelSample.bulkUpsertSample(value);
       } else {
+        const bulkUpsertPromise =
+        featureToggles.isFeatureEnabled(constants.featureName) ?
```
annyhe: what is the reason for the change to either bulkUsert through the redisModel, or through the api?

cache/sampleStoreTimeout.js
```diff
@@ -27,10 +28,13 @@ const ONE = 1;
  * @param  {Array} samples - Samples to be timedout
  * @param  {Hash} aspects - Aspects hash, name-> object
  * @param  {Date} curr - Current datetime
- * @returns {Array} Commands needed to timeout samples
+ * @returns {Object} - with with sampCmds and timedOutSamples properties.
```
annyhe: duplicate with.
also timedOutSamples sound like samples that already timed out. Perhpas use samplesToTimeOut?

tests/realtime/redisPublisher.js
```diff
@@ -0,0 +1,119 @@
...
+    .then((s) => {
+      sampId = s.id;
+      return Subject.create(subjectSA);
+    })
+    .then(() => done())
+    .catch(done);
+  });
+  after(u.forceDelete);
+  describe('publishSample function tests: ', () => {
+    it('with EventType argument: sample should be published with subject  ' +
```
annyhe: so regardless of the EventType argument, the returned results are the same?

tests/logging/enableWorkerLog.js
```diff
@@ -215,7 +215,8 @@ describe('enableWorkerLog: db: sample: timeout: ', () => {
     mockUpdatedAt.setMinutes(updatedAt.getMinutes() + fiveMinutes);
     Sample.doTimeout(mockUpdatedAt)
     .then((res) => {
-      expect(res).to.eql({ numberEvaluated: 4, numberTimedOut: 2 });
+      expect(res).to.contains({ numberEvaluated: 4, numberTimedOut: 2 });
```
annyhe: what is the reason for changing equals to contains? has the res changed?

db/model/sample.js
```diff
@@ -254,16 +255,32 @@ module.exports = function sample(seq, dataTypes) {
...
+
+              /*
+               * NOTE: a separate copy of the sample needed to be made (by
+               * deep cloning the required fields) to let the garbage collector
+               * cleanup the timedOutSamples array.
+               */
+              const timedOutSample = {
+                value: constants.statuses.Timeout,
+                status: constants.statuses.Timeout,
+                name: s.name.split()[0],
```
annyhe: is there a reason we are using s.name.split()[0] instead of s.name?

api/v1/controllers/samples.js
shriramshankar: There is no change to either of them here. The only change is the use of the ternary operator.

db/model/sample.js
shriramshankar: Yes, it creates a copy of the string instead. There is a note about it a few lines above.

tests/logging/enableWorkerLog.js
shriramshankar: res has changed

tests/realtime/redisPublisher.js
shriramshankar: It is the type of the event that is published. There is more information about it in publisher.js

realtime/redisPublisher.js
```diff
@@ -0,0 +1,134 @@
...
+      sample.subject = sub.get();
+
+      // attach absolutePath field to the sample
+      sample.absolutePath = subName;
+
+      /*
+       *pass the sample instance to the publishObject function only if the
+       *aspect and subject are published
+       */
+      if (sample.aspect.isPublished && sub.isPublished) {
```
pallavi2209: We can have this check before attaching the subject to sample. Correct?

realtime/redisPublisher.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * @param  {Object} inst - Model instance to be published
+ * @param  {String} event  - Type of the event that is being published
+ * @param  {[Array]} changedKeys - An array containing the fields of the model
+ * that were changed
+ * @param  {[Array]} ignoreAttributes An array containing the fields of the
+ * model that should be ignored
+ * @returns {Object} - object that was published
+ */
+function publishObject(inst, event, changedKeys, ignoreAttributes) {
```
pallavi2209: From where are we getting ignoreAttributes? Is this paramter for future use?

worker/jobProcessor.js
```diff
@@ -101,6 +110,13 @@ jobQueue.process(jobType.SAMPLE_TIMEOUT, (job, done) => {
   const dbStartTime = Date.now();
   sampleTimeoutJob.execute()
   .then((dbRes) => {
+    // send the timeoutsample to the client by publishing it to redis channel
+    if (dbRes && dbRes.timedOutSamples) {
+      dbRes.timedOutSamples.forEach((sample) => {
+        publisher.publishSample(sample, subHelper.model, sampleEvent.upd);
+      });
```
pallavi2209: Are we replicating this logic in clock/scheduledJobs/sampleTimeoutJob.js and here? Any reason why?

realtime/redisPublisher.js
shriramshankar: Yes for possible future use cases.

realtime/redisPublisher.js
```diff
@@ -0,0 +1,134 @@
...
+ * The sample object needs to be attached its subject object and it also needs
+ * a absolutePath field added to it before the sample is published to the redis
+ * channel.
+ * @param  {Object} sampleInst - The sample instance to be published
+ * @param  {Model} model - The subject model used to get the related subject
+ * instance
+ * @param  {String} event  - Type of the event that is being published
+ * @returns {Promise} - which resolves to a sample object
+ */
+function publishSample(sampleInst, model, event) {
```
shriramshankar: We should avoid requiring files as much as possible. Requiring the subject model from the realtime module to db module might  help with not having to deal with circular dependencies in the future.

worker/jobProcessor.js
shriramshankar: the clock process can run independent of the workers or can enqueue the job to the workers. We need it in both the places to handle these use cases.

cache/sampleStoreTimeout.js
shriramshankar: Sure, will change it.

realtime/redisPublisher.js
shriramshankar: Yes we can.

worker/jobProcessor.js
pallavi2209: So, sending 2 events for a sample timeout when worker process is enabled is fine?

shriramshankar: No, you are right. I have to change it.

shriramshankar:
@annyhe @pallavi2209 Updated the PR with the changes.

tests/realtime/redisPublisher.js
```diff
@@ -0,0 +1,119 @@
...
+        done();
+      })
+      .catch(done);
+    });
+
+    it('without EventType argument: sample should be published with subject ' +
+      ' object and asbolutePath field', (done) => {
+      Sample.findById(sampId)
+      .then((sam) => publisher.publishSample(sam, Subject))
+      .then((pubObj) => {
```
annyhe: are you intentionally leaving out the eventType in publishSample?

shriramshankar: yes, this test is for that use case.

PR #291, opened 2017-03-16
RedisStore: Refactoring of sample endpoints - get, find and upsert

pallavi2209:
Refactored get and find sample, removed apilinks from common method. Set links for find samples. Refactored upsert sample to not to pass req.method.

### DIFF
--- a/api/v1/controllers/samples.js
+++ b/api/v1/controllers/samples.js
@@ -53,9 +53,7 @@ module.exports = {
   findSamples(req, res, next) {
     if (featureToggles.isFeatureEnabled(constants.featureName)) {
       const resultObj = { reqStartTime: new Date() }; // for logging
-      redisModelSample.findSamples(
-        resultObj, req.method, req.swagger.params
-      )
+      redisModelSample.findSamples(req, res, resultObj)
       .then((response) => {
         // loop through remove values to delete property
         if (helper.fieldsToExclude) {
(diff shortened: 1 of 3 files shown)


api/v1/controllers/samples.js
```diff
@@ -83,27 +81,7 @@ module.exports = {
...
-        }
-
-        u.logAPI(req, resultObj, sampleRes); // audit log
-        res.status(httpStatus.OK).json(sampleRes);
-      })
-      .catch((err) => u.handleError(next, err, helper.modelName));
-    } else {
-      doGet(req, res, next, helper);
-    }
+    doGet(req, res, next, helper);
```
annyhe: yay less code :)

api/v1/controllers/samples.js
```diff
@@ -172,9 +150,7 @@ module.exports = {
           u.checkDuplicateRLinks(sampleQueryBody.relatedLinks);
         }
 
-        upsertSamplePromise = redisModelSample.upsertSample(
-          sampleQueryBody, resultObj, res.method
-        );
+        upsertSamplePromise = redisModelSample.upsertSample(sampleQueryBody);
       } else {
```
annyhe: are we still logging in redisModelSample.upsertSample?

cache/models/samples.js
```diff
@@ -777,7 +766,6 @@ module.exports = {
 
     return redisClient.batch(commands).execAsync()
     .then((responses) => {
-      logObject.dbTime = new Date() - logObject.reqStartTime; // log db time
       const sample = responses[ZERO];
```
annyhe: did we remove logging for this endpoint?

api/v1/controllers/samples.js
pallavi2209: Logging before returning response. Check the code after the changes.

cache/models/samples.js
pallavi2209: Logging in doGet verb instead.

api/v1/helpers/verbs/doGet.js
```diff
@@ -65,17 +68,18 @@ function doGet(req, res, next, props) {
...
-      // loop through remove values to delete property
-      if (props.fieldsToExclude) {
-        u.removeFieldsFromResponse(props.fieldsToExclude, o.dataValues);
-      }
+    getPromise.then((o) => {
+      resultObj.dbTime = new Date() - resultObj.reqStartTime;
 
-      // o is a sequelize obj, get dataValues obj
-      u.logAPI(req, resultObj, o.dataValues);
+      u.logAPI(req, resultObj, o);
```
annyhe: what is o? is the logged object the dataValues object, ie. the API response?

pallavi2209: o can be sequelize object or redis object, not datavalues object.

annyhe: o is used by the logging utility to return the approximate the size of returned information. If o is the sequelize object or redis object, the approximation may be way off. 
We may need to modify the getSize(object) function in utils/apiLog.js to parse the object.

pallavi2209: I extract the dataValues from Sequelize object in apiLog function. Please  look into details to confirm.

PR #293, opened 2017-03-17
redisstore: Delete endpoints for related links and tags for samples and aspects

pallavi2209:
added tests for aspect tag and related links delete endpoints
delete ssample related links endpoint


cache/models/samples.js
```diff
@@ -507,6 +507,65 @@ module.exports = {
...
+          explanation: 'Sample not found.',
+        });
+      }
+
+      currSampObj = sampObj;
+      const aspectName = sampleName.split('|')[ONE];
+      return redisOps.getHashPromise(aspectType, aspectName);
+    })
+    .then((aspObj) => {
+      if (!aspObj) {
```
annyhe: why are we checking for aspect only? do we also need to check whether the subject exists?

tests/cache/models/aspects/deleteRelatedLinks.js
```diff
@@ -0,0 +1,128 @@
...
+    .expect(constants.httpStatus.OK)
+    .end((err, res) => {
+      if (err) {
+        done(err);
+      }
+
+      expect(res.body.relatedLinks).to.have.length(ZERO);
+      redisOps.getHashPromise(objectType.aspect, n.name)
+      .then((aspect) => {
+        expect(JSON.parse(aspect.relatedLinks)).to.have.length(0);
```
annyhe: 0 as in ZERO?

tests/cache/models/aspects/deleteTags.js
```diff
@@ -0,0 +1,141 @@
...
+        const tags = JSON.parse(aspect.tags);
+        expect(tags).to.have.length(1);
+        expect(tags).to.have.members(['tag1']);
+        done();
+      });
+    });
+  });
+
+  it('error if tag not found', (done) => {
+    api.delete(oneDeletePath.replace('{key}', aspId).replace('{akey}', 'x'))
```
annyhe: no error here in the test. Did you want to check for an error?

cache/models/samples.js
pallavi2209: Sample exists only is subject exits. Thats what we check when creating a sample.

tests/cache/models/aspects/deleteRelatedLinks.js
pallavi2209: Yes. Let me make this change with the changes based on Shriram's comments as well, when he reviews.

tests/cache/models/aspects/deleteTags.js
pallavi2209: I just copied this test from api and checked the redis entry. I dont think we throw an error if tag does not exist. Is that correct @shriramshankar ? If that is true, I can change the description of test.

shriramshankar: No errors thrown if tag does not exist.

shriramshankar:
There are issues with the implementation of "post sample", sampleTimeout and  probably "sample delete" too. Probably there is something we can learn from fixing these issues that could be applicable in these changes. I want to make sure that these issues are addressed before I review this PR.

pallavisingh9:
Okay, let me double check those implementations. Let me know if you find something wrong, I can work on it.

pallavi2209:
@annyhe Are we waiting on something here to review/approve?

PR #292, opened 2017-03-17
added permset in the cache layer - upsert/ bulk upsert endpoints

shriramshankar:

PR #294, opened 2017-03-18
move sample realtime events for put/patch/delete to api layer

shriramshankar:


shriramshankar:
@annyhe @iamigo PR #292 needs to be merged before this one

realtime/redisPublisher.js
```diff
@@ -92,32 +92,49 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
...
-  const subName = sample.name.split('|')[0];
-  const options = {};
-  options.where = { absolutePath: subName };
-  return model.findOne(options)
+  const nameParts = sample.name.split('|');
+  const subName = nameParts[0];
+  const aspName = nameParts[1];
+  const subOpts = {
+    where: {
+      absolutePath: subName,
```
iamigo: case insensitive?

realtime/redisPublisher.js
```diff
@@ -92,32 +92,49 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
...
+    },
+  };
+  const aspOpts = {
+    where: {
+      name: aspName,
```
iamigo: case insensitive?

shriramshankar: aspectName and subjectName are always from the "sample.name" field from the db and the "sample.name" field is formed from its subject and its aspect in the db. So the case always matches when we do the find here.

tests/realtime/redisPublisher.js
```diff
@@ -81,6 +81,29 @@ describe('redis Publisher', () => {
       })
       .catch(done);
     });
+
+    it('when tried to publish sample without aspect,'+
```
iamigo: any other tests we can add here to make sure that we aren't breaking any of the existing realtime event publishing?

shriramshankar: The sample object the is finally passed to the real-time emitter should have the following fields/objects to not to break anything
1. absolutePath - as a field copied over from subject absolutePath
2. subject object always attached - having  tags field
3. aspect object always attached having tags field
I have added tests to check for these.

api/v1/helpers/verbs/doPut.js
```diff
@@ -75,6 +77,14 @@ function doPut(req, res, next, props) {
   putPromise.then((o) => {
     resultObj.dbTime = new Date() - resultObj.reqStartTime;
     u.logAPI(req, resultObj, o);
+
+    // publish the update event to the redis channel
+    if (props.publishEvents) {
+      publisher.publishSample(o, props.associatedModels.subject,
+
```
annyhe: is there a reason we are adding an empty line here?

db/model/sample.js
```diff
@@ -320,52 +322,6 @@ module.exports = function sample(seq, dataTypes) {
       }, // hooks.beforeCreate
 
       /**
-       * Publishes the created sample to redis channel *including* the values
-       * from its aspect association.
-       *
-       * @param {Sample} inst - The newly-created instance
-       */
-      afterCreate(inst /* , opts */) {
-        let samp;
```
annyhe: can we test for whether a sample is published on realtime create and delete?

shriramshankar: There is another story for this.

shriramshankar:
@iamigo @annyhe  Changes done and PR updated.

shriramshankar:
@annyhe  Can you merge this PR?

PR #296, opened 2017-03-22
using a subject-only hierarchy when cache is enabled

shriramshankar:


api/v1/helpers/nouns/subjects.js
```diff
@@ -159,6 +159,7 @@ module.exports = {
   fieldScopeMap: {
     samples: 'withSamples',
     hierarchy: 'hierarchy',
+    subjectHierarchy: 'subjectHierarchy',
   },
```
shriramshankar: If we have a better name then "subjectHierarchy" , I can change it.

iamigo: The only suggestion I could come up with is "subjectOnlyHierarchy", but I leave that up to you.

PR #299, opened 2017-03-22
read-only fields for Sample/Aspects/Subjects through the PUT/PATCH/POST api endpoints

pallavi2209:
Read only fields:
- Aspect: id, isDeleted
- Sample: id, isDeleted, status, previousStatus, statusChangedAt, createdAt, updatedAt
- Subject: absolutePath, childCount, id, isDeleted

Added tests.

### DIFF
--- a/api/v1/controllers/utils.js
+++ b/api/v1/controllers/utils.js
@@ -70,8 +70,7 @@ function noReadOnlyFieldsInReq(req, props) {
   if (props.readOnlyFields) {
     props.readOnlyFields.forEach((field) => {
       // if request body is an array, check each object in array.
-      if (typeof requestBody !== 'undefined' &&
-            requestBody && requestBody.constructor === Array) {
+      if (Array.isArray(requestBody)) {
         requestBody.forEach((reqObj) => {
           checkReadOnlyFieldInObj(field, reqObj);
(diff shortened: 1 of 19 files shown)


api/v1/controllers/utils.js
```diff
@@ -44,6 +45,44 @@ function hasDuplicates(tagsArr) {
...
+ * @param  {Object} props - The module containing the properties of the
+ *  resource type.
+ */
+function noReadOnlyFieldsInReq(req, props) {
+  const requestBody = req.body;
+  if (props.readOnlyFields) {
+    props.readOnlyFields.forEach((field) => {
+      // if request body is an array, check each object in array.
+      if (typeof requestBody !== 'undefined' &&
+            requestBody && requestBody.constructor === Array) {
```
shriramshankar: Array.isArray does something similar, correct?

pallavi2209: Yes, it is and I think I didn't use it because is only supported in ES5 and both have similar performance. But I guess it does not matter because we are already using ES6 features and I agree Array.isArray is concise and readable.
Let me change this.

pallavi2209:
@shriramshankar Updated the PR

PR #300, opened 2017-03-22
Add vertical scrollbars on modal dropdown

annyhe:
- also remove the search icon from the input field


annyhe:
Review app is at http://secure-island-62458.herokuapp.com/perspectives/A_SR_Grid
To test:
- Click the Create Perspective, or 
- edit a perspective by clicking on a pencil button, then 
- open a dropdown in the modal. if the dropdown is long enough, it should have a vertical scrollbar, and be scrollable

iamigo:
just curious about why you removed the search icon

iamigo:
what username/pw do you use for that review app?

pallavi2209:
@iamigo I think your staging username/pwd will work

iamigo:
the right half of the scrollbar appears to be cut off

pallavi2209:
@annyhe So the search will not work in these dropdowns now? I think it used to work before. Feels it is inconvenient for users. Also, not sure why perspectives are not loading any data.

annyhe:
the search will still work. 
I took out the search icon because it was not placed in the correct position, and the user already knows to start searching, so it was redundant. 
The changes in this PR does not affect any data loading, and the data should be the same from staging as of yesterday.
@iamigo Where does the scrollbar appear to be cut off? Can you add a screenshot?

iamigo:
I couldn't screen capture because it only renders when in focus... but i can show you on hangout

pallavi2209:
So, is it better if we say "Search a subject" instead of "Select a subject" like we have "Search perspective" on main display?
I don't see the search icon currently in prod as well, so I guess that was not the problem, you just had to change the class property to fix scrolling. Correct?

annyhe:
@pallavi2209 Yes I changed the css property so the dropdown in modal will show up, and the dropdown will scroll.
@iamigo How do you adjust the chrome settings to induce the half-hidden vertical scrollbar? I am having trouble reproducing it.

annyhe:
@iamigo There's no simple way to fix the half-hidden vertical scrollbar so it looks consistent on both the monitor and on mac screen with preference scroll: When scrolling.
Let's merge this and move on.

PR #301, opened 2017-03-23
Add “tags” to list of fields available for GET /v1/subjects?fields=…

iamigo:


tests/api/v1/subjects/get.js
```diff
@@ -395,10 +395,25 @@ describe(`api: GET ${path}`, () => {
...
+    .set('Authorization', token)
+    .expect(constants.httpStatus.OK)
+    .end((err, res) => {
+      if (err) {
+        done(err);
+      }
+
+      expect(res.body[ZERO]).to.have.property('tags');
+      expect(res.body[ZERO]).to.have.property('name');
+      expect(res.body[ZERO]).to.have.property('isPublished');
```
annyhe: can the test check whether the tags have the expected values? ie. setup subject with tags: ['aaa', 'bbb'] and expect res.body.tags to equal ['aaa', 'bbb']

iamigo: well, it _can_, but those kinds of tests tend to be in the db layer... this is just testing that the api layer is doing the right thing

annyhe:
@iamigo there seems to be a flapping test. Re-run the travis tests again or force push to trigger the tests, or merge via command line

PR #305, opened 2017-03-25
Store token in session instead of cookies, just stored session id in cookies

harshkothari410:
- Token is stored in Session
- session id is stored in cookies
- user info in session as well
- removed all occurrences of getCookie authorization 
- fixed test


utils/jwtUtil.js
```diff
@@ -89,12 +89,7 @@ function checkTokenRecord(t) {
  * @param  {Function} cb - callback function
  */
 function verifyToken(req, cb) {
-  const authorization = req.headers.authorization;
-  let token;
-
-  if (authorization) {
-    token = req.headers.authorization;
-  }
+  const token = req.session.token || req.headers.authorization;
```
iamigo: Yes!

harshkothari410:
I created Review App @iamigo : http://refocus-pr-305.herokuapp.com/

harshkothari410:
@iamigo without SSO I successfully run it. You can also check it out. With SSO I do not know how to run.

iamigo:
FYI @harshkothari410 docs on setting up SSO (but you don't need to do it in your review app): https://salesforce.github.io/refocus/docs/80-SSOSetup.html

view/loadView.js
```diff
@@ -306,9 +306,9 @@ module.exports = function loadView(app, passport) {
 
         // return token
         const token = jwtUtil.createToken(req.user.name, req.user.name);
+        req.session.token = token;
```
pallavi2209: So now if we don't need to send plain token back, can we remove /authenticate route and make use of /v1/authenticate instead?

pallavi2209:
We are storing session in cookie for SSO flow as well. Please check /sso/saml route in loadView. We will have to change that as well.

view/loadView.js
harshkothari410: I guess we can do this without any problem. Can I try to remove that ? @pallavi2209 @iamigo

pallavi2209: Fine with me.

view/admin/actions/index.js
```diff
@@ -152,7 +151,6 @@ function requestWithOutPayload(verb, url) {
   return new Promise((resolve, reject) => {
     superagent(verb, API_URL + url)
     .set('Content-Type', 'application/json')
-    .set('Authorization', u.getCookie('Authorization'))
```
pallavi2209: Do we still need getCookie and setCookie functions?

harshkothari410: Yes I guess because its general function for cookies not just for authorization cookies.

PR #306, opened 2017-03-26
Get perspectives filters

iamigo:
Add support for GET /v1/perspectives filter by name and rootSubject
e.g. GET /v1/perspectives?name=Abc*

### DIFF
--- a/api/v1/swagger.yaml
+++ b/api/v1/swagger.yaml
@@ -2015,6 +2015,20 @@ paths:
           $ref: "#/parameters/limitParam"
         -
           $ref: "#/parameters/offsetParam"
+        -
+          name: name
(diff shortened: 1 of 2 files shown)


annyhe:
The failing travis tests are about samples. We are also getting lots of 'Unhandled rejection SequelizeForeignKeyConstraintError: insert or update on table "Subjectsancestors" violates foreign key constraint "Subjectsancestors_SubjectId_fkey' when running api tests. Do you know what causes this @iamigo @shriramshankar?

shriramshankar:
Not sure what would cause "Unhandled rejection SequelizeForeignKeyConstraintError: ...". I think there was a story to fix this.

PR #307, opened 2017-03-27
add re-parent subject tests, for PUT and PATCH

annyhe:
- refactored beforeUpdate in db/model/subject.js; previously the cause of parenting errors on un-publish
- renamed test files
- add tests as follows:
(0) PATCH or PUT /v1/subjects/{key} unpublishing leaf subject with parentId or parentAbsolutePath field should re-parent the subject as if isPublished: true 
(1) PATCH or PUT /v1/subjects/{key} unpublishing leaf subject without parentId nor parentAbsolutePath field should set the subject as top level subject
  (a) PATCH or PUT /v1/subjects/{key} with parentId and parentAbsolutePath fails if parentAbsolutePath does not refer to the same subject record as specified by parentId
  (b) PATCH or PUT /v1/subjects/{key} with parentId BUT NO parentAbsolutePath sets the parent based on the parentId
  (c) PATCH or PUT /v1/subjects/{key} with parentAbsolutePath BUT NO parentId sets the parent based on the parentAbsolutePath
  (d) PATCH /v1/subjects/{key} with NEITHER parentId NOR parentAbsolutePath does not reparent the subject
  (d.1) PUT /v1/subjects/{key} with NEITHER parentId NOR parentAbsolutePath makes a subject root
  (e) PATCH or PUT /v1/subjects/{key} with parentId === id fails
  (e) PATCH or PUT /v1/subjects/{key} with parentAbsolutePath === absolutePath fails
  (f) PATCH or PUT /v1/subjects/{key} with parentId unchanged, parentAbsolutePath pointing to a different parent fails
  (g) PATCH or PUT /v1/subjects/{key} with parentAbsolutePath unchanged, parentId pointing to a different parent fails


api/v1/controllers/subjects.js
```diff
@@ -318,7 +326,37 @@ module.exports = {
    */
   patchSubject(req, res, next) {
     validateRequest(req);
-    doPatch(req, res, next, helper);
+    const queryBody = req.swagger.params.queryBody.value;
+    const { parentAbsolutePath, parentId } = queryBody;
+
+    // if both parentAbsolutePath and parentId,
+    // check they point to the same subject
+    if (queryBody.parentId && parentAbsolutePath) {
```
iamigo: why queryBody.parentId here and not just parentId?

api/v1/controllers/subjects.js
```diff
@@ -373,7 +411,37 @@ module.exports = {
    */
   putSubject(req, res, next) {
     validateRequest(req);
-    doPut(req, res, next, helper);
+    const queryBody = req.swagger.params.queryBody.value;
+    const { parentAbsolutePath, parentId } = queryBody;
+
+    // if both parentAbsolutePath and parentId,
+    // check they point to the same subject
+    if (queryBody.parentId && parentAbsolutePath) {
```
iamigo: same comment as above

api/v1/controllers/subjects.js
```diff
@@ -373,7 +411,37 @@ module.exports = {
...
+      helper.model.findOne({
```
iamigo: Any way to have put and patch both use the same block of code?

db/dbErrors.js
```diff
@@ -71,6 +71,25 @@ dbErrors.create({
   explanation: 'If a subject specifies that it has a parent, the parent ' +
    'subject must already exist.',
 });
+dbErrors.create({
+  code: 10115,
+  status: 400,
+  name: 'ParentSubjectNotMatch',
+  parent: dbErrors.ValidationError,
+  range: [],
+  defaultMessage: 'The parent\'s id does not match specified parentId.',
```
iamigo: `The parentAbsolutePath and parentId do not match.`

db/dbErrors.js
```diff
@@ -71,6 +71,25 @@ dbErrors.create({
...
+  explanation: 'If a subject specifies both a parentAbsolutePath and a parentId, the parent ' +
+   'specified by both fields need to have the same id.',
+});
+dbErrors.create({
+  code: 10115,
+  status: 400,
+  name: 'ParentCannotBeOwnSubject',
```
iamigo: `IllegalSelfParenting`

db/dbErrors.js
```diff
@@ -71,6 +71,25 @@ dbErrors.create({
...
+  parent: dbErrors.ValidationError,
+  range: [],
+  defaultMessage: 'The parent\'s id cannot be identical to the subject id. ' +
```
iamigo: `A subject may not be its own parent.`

db/model/subject.js
```diff
@@ -539,37 +542,122 @@ module.exports = function subject(seq, dataTypes) {
...
+          function checkParent() {
+            // by default, set params to use parentAbsolutePath
+            setValues(true);
+
+            // when parentAbsolutePath is supplied
+            if (inst.changed('parentAbsolutePath')) {
+              if (inst.parentAbsolutePath) {
+
+                // find subject whose absolutePath === parentAbsolutePath
+                // if it does not exist, throuw 404 subject not found
```
iamigo: throw

tests/db/model/subject/updateWithParentIdAndParentAbsolutePath
```diff
@@ -0,0 +1,284 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
iamigo: 2017

tests/db/model/subject/updateWithParentIdAndParentAbsolutePath
```diff
@@ -0,0 +1,284 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/db/model/subject/updateChild.js
```
iamigo: mismatch of file name?

annyhe:
@iamigo Updated as per comments

db/model/subject.js
```diff
@@ -539,37 +542,122 @@ module.exports = function subject(seq, dataTypes) {
...
+                  .then((parent) => {
+                    // if parentId changed too, check for consistency
+                    if (inst.changed('parentId') && inst.parentId) {
+
+                      if (parent && parent.id !== value1) {
+                        // parentAbsolutePath does not match parentId
+                        throw new ParentSubjectNotMatch({
+                          message: value + ' does not match ' + value1,
+                        });
+                      } else if (!parent) {
```
pallavi2209: Can this condition be outside  if (inst.changed('parentId') && inst.parentId) check?

db/model/subject.js
```diff
@@ -539,37 +542,122 @@ module.exports = function subject(seq, dataTypes) {
...
+                  })
+                  .catch((err) => {
+                    reject(err);
+                  });
+                });
+              } else if (!inst.changed('parentId')) {
+
+                // body contains parentAbsolutePath with empty value
+                // set parentAbsolutePath to null
+                return new seq.Promise((resolve, reject) => {
```
pallavi2209: return new seq.Promise.resolve(); Does this work?

db/model/subject.js
```diff
@@ -539,37 +542,122 @@ module.exports = function subject(seq, dataTypes) {
...
+                  resolve();
+                });
+              }
+            } {
+              // only the parentId field changed. It may be empty
+              // find parent subject
+              if (inst.parentId === inst.id) {
```
pallavi2209: How do you know parent if changed here?

db/model/subject.js
```diff
@@ -539,37 +542,122 @@ module.exports = function subject(seq, dataTypes) {
...
           return new seq.Promise((resolve, reject) => {
-            Subject.scope({ method: [param, value] }).find()
+            checkParent()
             .then((parent) => {
               if (parent) {
                 inst.setDataValue('absolutePath',
                   parent.absolutePath + '.' + inst.name);
                 inst.setDataValue(key, value);
-                inst.setDataValue(key1, parent.getDataValue(param1));
+                inst.setDataValue(key1, parent[param1]);
```
pallavi2209: This code must be working, but it was very confusing for me to understand. I feel this code can be improved a lot in terms of readability when using specific variables instead of setValues function.

api/v1/controllers/subjects.js
```diff
@@ -31,11 +31,66 @@ const u = require('../helpers/verbs/utils');
 const httpStatus = require('../constants').httpStatus;
 const apiErrors = require('../apiErrors');
 const redisSubjectModel = require('../../../cache/models/subject');
+
+const ParentSubjectNotFound = require('../../../db/dbErrors')
```
pallavi2209: dbErrors in api layer? Should be create this errors in apiErrors file instead?

db/dbErrors.js
```diff
@@ -71,6 +71,25 @@ dbErrors.create({
   explanation: 'If a subject specifies that it has a parent, the parent ' +
    'subject must already exist.',
 });
+dbErrors.create({
+  code: 10115,
+  status: 400,
+  name: 'ParentSubjectNotMatch',
+  parent: dbErrors.ValidationError,
+  range: [],
```
annyhe: I included it due to it being in the previous error. I will take it out as it doesn't seem necessary.

tests/api/v1/subjects/patchWithParent.js
```diff
@@ -0,0 +1,313 @@
...
+    api.patch(`${path}/${i1}`)
+    .set('Authorization', token)
+    .send({
+      name: `${tu.namePrefix}Canada`,
+      parentId: i0a,
+      parentAbsolutePath: n0.name, // unchanged parent
+    })
+    .expect(constants.httpStatus.BAD_REQUEST)
+    .end((err  , res ) => {
+      console.log(err, res.error)
```
pallavi2209: Why console log?

tests/api/v1/subjects/patchWithParent.js
```diff
@@ -0,0 +1,313 @@
...
+const ONE = 1;
+const TWO = 2;
+
+describe(`api: PATCH ${path} with parents`, () => {
+  let token;
+  const n0 = { name: `${tu.namePrefix}Canada`, isPublished: true };
+  const n1 = { name: `${tu.namePrefix}Ontario`, isPublished: true };
+  const n2 = { name: `${tu.namePrefix}Manitoba`, isPublished: true };
+  const p0 = { name: `${tu.namePrefix}NA`, isPublished: true };
+  const p1 = { name:
```
pallavi2209: A tip: Its better to use logical names  instead of 'n0' 'n1's etc, so that they are more readable and a new user can understand the code in just a look. Or add some comments/examples explaining the hierarchy in your mind

tests/api/v1/subjects/patchWithParent.js
```diff
@@ -0,0 +1,313 @@
...
+      if (err) {
+        done(err);
+      }
+
+      expect(res.body.errors[0].type).to.equal('ParentSubjectNotMatch');
+      done();
+    });
+  });
+
+  it('fail when trying to update with parentAbsolutePath pointing to different subject than parentId', (done) => {
```
pallavi2209: Duplicate test?

tests/api/v1/subjects/patchWithParent.js
```diff
@@ -0,0 +1,313 @@
...
+      name: p1.name,
+      parentId: i0a,
+    })
+    .expect(constants.httpStatus.OK)
+    .end((err, res ) => {
+      if (err) {
+        done(err);
+      }
+
+      expect(res.body.parentAbsolutePath).to.equal(n0a.name);
```
pallavi2209: Can we check for parent ids also for similar tests?

tests/api/v1/subjects/putWithParent.js
```diff
@@ -0,0 +1,302 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/api/v1/subjects/putWithParent.js
```
pallavi2209: Same comments as patchWithParent file

tests/db/model/subject/updateParent.js
```diff
@@ -0,0 +1,403 @@
...
+
+  /*
+   * Visual depiction below:
+   * Before                            After
+   * O<= root parent1 O<= new parent2  * O<= root parent1 O<= new parent2
+   * |               /|                *                  |
+   * O<= child1 ===>/ O<= child1       *                  O<= child1
+   * |                |                *                  |
+   * O<= child2       O<= child2       *                  O<= child2
+   */
```
pallavi2209: +1 on effort here! :)

annyhe:
on db beforeUpdate, the code is getting complicated. All we added is to validate the parentAbsolutePath and parentId fields, if they have changed.

Here is the simplified version of the requested change. Feedback is welcome.
// PAP is shorthand for parentAbsolutePath, 
// PID is shorthand for parentId
initialize PAP_changed, PID_changed: check whether the fields are in inst.changed() <- an array
If both are false,  there's no update to do. Return

initialize PAP_empty, PID_empty: check whether the body fields are empty
If both are true, decrement the parent's childCount AND update the subject to root subject, 

initialize PAP_exist, PID_exist: check whether the subjects they point to exist.
if either does not, throw a ParentNotFound error
if both exist and their id's don't match, throw ParentNotMatch error.
if PAP === absolutePath || PID === id, throw cannot self parent error.

// if no errors are thrown yet
re-parent the subject 

// end

Feedback is welcome @pallavi2209 @iamigo

tests/api/v1/subjects/patchWithParent.js
annyhe: I will take that out. Thanks for catching it.

iamigo:
Sounds reasonable, Anny

annyhe:
@shriramshankar updated pull request, please review

tests/api/v1/subjects/patchWithParent.js
annyhe: good point. will take out.

api/v1/controllers/subjects.js
annyhe: replaced dbErrors with apiErrors

tests/api/v1/subjects/patchWithParent.js
annyhe: checks added

tests/api/v1/subjects/putWithParent.js
annyhe: done.

tests/api/v1/subjects/patchWithParent.js
annyhe: The hierarchy is pretty clear from the beforeEach setup.We can refactor the tests next time. This pull request is already getting large.

tests/db/model/subject/updateParent.js
annyhe: that was an old test :)

shriramshankar:
Will take a look. I want to test this too, to make sure that we do not have the hierarchy issue again.

tests/api/v1/subjects/putWithParent.js
```diff
@@ -0,0 +1,308 @@
...
+    .expect(constants.httpStatus.OK)
+    .end((err  , res ) => {
+      if (err) {
+        done(err);
+      }
+
+      expect(res.body.absolutePath).to.equal(NAME);
+      done();
+    });
+  });
```
shriramshankar: Any test to check that unpublishing a subject using a "put" does not cause reparenting?

annyhe: great point. will add test

api/v1/swagger.yaml
```diff
@@ -4221,6 +4226,11 @@ paths:
                 pattern: ^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$
                 description: >
                   The id of the subject's parent.
+              parentAbsolutePath:
+                type: string
+                maxLength: 254
```
pallavi2209: Sorry I missed this before, but the field length is 4096 (longish) specified in db model. Please confirm that this maxLength is what we want.

tests/api/v1/subjects/patchWithParent.js
pallavi2209: yeah it was a tip for future only :)
BeforeEach is clear but when we refer the variables in tests later, they are non-intuitive. I just wanted to stress that it is always a good practice to give logical names to variables. Thx!

tests/api/v1/subjects/putWithParent.js
annyhe: Test is failing:

  it.only('on un-publish, does not re-parent', (done) => {
    api.put(`${path}/${i1}`)
    .set('Authorization', token)
    .send({
      name: `${tu.namePrefix}Canada`,
      isPublished: false,
    })
    .end((err  , res ) => {
      if (err) {
        done(err);
      }

      console.log(res.body)
      expect(res.body.parentId).to.equal(i0); // parentId field is undefined
      expect(res.body.parentAbsolutePath).to.equal(n0.name); // '' instead of non-empty value: shows subject is re-parented
      done();
    });
  });

I am looking at the culprit files in db. Will push another commit when that is fixed

shriramshankar: May be try putting with all the fields like the request from the admin?

api/v1/swagger.yaml
annyhe: ok will do

tests/api/v1/subjects/putWithParent.js
annyhe: actually the re-parent behavior is correct, except we want a parentId field to be returned (it is also not returning on PUT with published: true and no parentId or parentAbsolutePath fields).

annyhe: To clarify:
  (0) PUT /v1/subjects/{key} unpublishing leaf subject with parentId or parentAbsolutePath field should re-parent the subject as if isPublished: true <------ this is not the current behavior. There's a bug. Will push another commit when that is fixed
  (1) PUT /v1/subjects/{key} unpublishing leaf subject without parentId nor parentAbsolutePath field should set the subject as top level subject: <------- this is the current behavior.

annyhe: I am getting a promises error:
updateParent(...).then is not a function.
But it clearly is a full-fledged sequelize object:
...     '$schema': null,
     '$schemaDelimiter': '',
     raw: true,
     attributes: 
      [ 'absolutePath',
        'childCount',
        'description',
        'geolocation',
        'helpEmail',
        'helpUrl',
        'id',
        'imageUrl',
        'isDeleted',
        'isPublished',
        'name',
        'parentAbsolutePath',
        'relatedLinks',
        'tags',
        'sortBy',
        'createdAt',
        'updatedAt',
        'deletedAt',
        'hierarchyLevel',
        'parentId',
        'createdBy' ] },
  hasPrimaryKeys: true,
  __eagerlyLoadedAssociations: [],
  isNewRecord: false }

Will troubleshoot tomorrow.

annyhe: resolved with return new seq.Promise((resolve, reject) => resolve(inst));

annyhe:
@shriramshankar @iamigo @pallavi2209 added the un-publish tests on api and db level. Feedback is welcome

api/v1/helpers/verbs/utils.js
```diff
@@ -567,6 +567,8 @@ function cleanAndStripNulls(obj) {
       // to pass swagger's schema validation
       if (key === 'parentAbsolutePath' && !o[key]) {
         o[key] = '';
+      } else if (key === 'parentId' && !o[key]) {
+        o[key] = null;
```
annyhe: previously on PUT with isPublished and name fields only, the parentId field was not being returned. This allows parentId to be returned even though it has a null value

db/model/subject.js
```diff
@@ -516,9 +523,9 @@ module.exports = function subject(seq, dataTypes) {
          * unpublished too. If any of the children are published, throw a
          * validation error
          */
-        if (inst.getDataValue('isPublished') === false) {
-          return new seq.Promise((resolve, reject) => {
-            inst.getChildren()
+        function checkPublished() {
```
annyhe: The entire beforeUpdate was refactored. It is best to view the entire file to see the changes.
The un-publish bug was caused by checking for isPublished===false returning the instance without checking for name or parent updates. The refactored version does:

checkIsPublished() <-- throws error if unpublishing a parent, else returns the instance
.then((inst) => updateParent(inst))
.then((inst) => updateName(inst))

This way all three checks are completed in the expected order.

db/helpers/subjectUtils.js
```diff
@@ -0,0 +1,98 @@
...
+ * @param {String} parentId
+ * @param {String} parentAbsolutePath
+ */
+function throwNotMatchError(parentId, parentAbsolutePath) {
+  throw new ParentSubjectNotMatch({
+    message: 'parentId and parentAbsolutePath refer to different subjects.' +
+    ' Found: ' + parentId + ', ' + parentAbsolutePath,
+  });
+}
+
```
shriramshankar: Any reason this error is defined here, rather than the dbError file?

db/model/subject.js
```diff
@@ -532,94 +539,100 @@ module.exports = function subject(seq, dataTypes) {
...
+          // pip is shorthand for parentId
+          const papChanged = inst.changed('parentAbsolutePath');
+          const pidChanged = inst.changed('parentId');
+
+          // initialize papEmpty, pidEmpty: check whether the
+          // fields are '' or null or falsey
+          const papEmpty = inst.parentAbsolutePath == null ||
+            inst.parentAbsolutePath == false;
+          const pidEmpty = inst.parentId == null || inst.parentId == false;
+
```
shriramshankar: do something like !inst.parentAbsolutePath or use ===.

api/v1/swagger.yaml
pallavi2209: Confirmed?

db/helpers/subjectUtils.js
annyhe: the error is not defined here, it is thrown here

db/model/subject.js
annyhe: good point. I will change to use == null

shriramshankar:
@annyhe Can you also test to make sure that the real-time events for subjects are not broken because of these changes?

annyhe:
@shriramshankar I tested real time events in staging, and it works. However bulk upsert is broken, ie. there's no data intake. Will troubleshoot by trying to replicate it locally

annyhe:
@shriramshankar on testing re-parent with correct and non-empty parentAbsolutePath, the test succeeds on local, but fails on staging environment. Do you know what can cause this?
Local db is: postgres (PostgreSQL) 9.6.2, which matches the heroku version

PR #308, opened 2017-03-27
Hard delete samples

shriramshankar:


db/model/sample.js
```diff
@@ -370,11 +347,14 @@ module.exports = function sample(seq, dataTypes) {
 
     }, // hooks
     indexes: [
-      { unique: true, fields: ['aspectId', 'subjectId', 'isDeleted'] },
       {
-        name: 'SampleStatusDeletedAt',
+        name: 'AspectIdSubjectId',
+        unique: true,
+        fields: ['aspectId', 'subjectId'],
```
iamigo: where is this used?

db/model/sample.js
```diff
@@ -370,11 +347,14 @@ module.exports = function sample(seq, dataTypes) {
...
+      },
+      {
+        name: 'SampleStatus',
         fields: [
-          'deletedAt',
           'status',
```
iamigo: where is this used?

tests/api/v1/aspects/utils.js
```diff
@@ -39,10 +39,10 @@ module.exports = {
   subjectToCreate,
 
   forceDelete(done) {
-    tu.forceDelete(tu.db.Aspect, testStartTime)
+    tu.forceDelete(tu.db.Sample, testStartTime)
```
iamigo: good catch!

db/model/sample.js
shriramshankar: AspectIdSubjectId and SampleStatus are old indices. AspectIdSubjectId was named aspectid_subjectid_isDeleted and SampleStatus was named SampleStatusIdDeleted. They were using the isDeleted and deletedAt fields, and I have removed these fields from the indices.

shriramshankar:
@iamigo @annyhe fixed the flapping test.

api/v1/helpers/verbs/doDelete.js
```diff
@@ -42,11 +43,22 @@ function doDelete(req, res, next, props) {
...
-      .then((o) => o.destroy());
+      .then((o) => {
+        /*
+         * An empty array is returned when destroy is called on an instance
+         * of a model that has hard delete turned on. We still want to return
+         * the deleted instance in all the cases. So, before the instance
+         * is destroyed it is saved and for models with hard deleted turned on
+         * this saved instance is returned
+         */
+        obj = o;
```
shriramshankar: I have not set the isDeleted field here, considering that the models with hard deleted turned on do not have the "isDeleted" field, does it make sense to set it here? @iamigo

iamigo: I don't think you should add it here, but let's make sure the swagger definition of the response schema is right

shriramshankar: Ok. Nothing to change in swagger too, there is no information about this field.

shriramshankar:
@annyhe Can you merge this PR if you are done with your review?

tests/db/model/sample/delete.js
```diff
@@ -0,0 +1,83 @@
...
+      expect(samp.dataValues).to.not.have.any.keys('isDeleted', 'deletedAt');
+      done();
+    })
+    .catch(done);
+  });
+  it('samples should not be found after destory is called', (done) => {
+    Sample.findById(sample.id)
+    .then((samp) => samp.destroy())
+    .then((o) => {
+      expect(o).to.have.length(0);
```
annyhe: is o an object? if so, why are we checking for o.length?

PR #309, opened 2017-03-28
passing in the right object to the log api function

shriramshankar:


api/v1/helpers/verbs/doGet.js
```diff
@@ -62,8 +62,8 @@ function doGet(req, res, next, props) {
         resultObj.dbTime = new Date() - resultObj.reqStartTime;
         const dbObj = JSON.parse(reply);
 
-        // dbObj is a sequelize obj, get dataValues obj
-        u.logAPI(req, resultObj, dbObj.dataValues);
+        // dbObj is a sequelize obj
+        u.logAPI(req, resultObj, dbObj);
```
iamigo: Any way to add a test for this?

shriramshankar:
@iamigo test added to cover this case.

PR #313, opened 2017-03-29
protect against bad entry in REPLICAS config var

harshkothari410:


db/utils.js
```diff
@@ -63,12 +63,15 @@ function getReadOnlyDBConfig(readReplicas) {
     readConfig = [];
     readReplicas.forEach((replicaUrl) => {
       const dbConfObj = dbConfigObjectFromDbURL(replicaUrl);
-      readConfig.push({
-        host: dbConfObj.host,
-        port: dbConfObj.port,
-        username: dbConfObj.user,
-        password: dbConfObj.password,
-      });
+      if (dbConfObj.host && dbConfObj.port &&
```
iamigo: if (dbConfObj && dbConfObj.host...

iamigo: pls. add something in tests/db/utils.js

harshkothari410: Yeah let me test first on review app to verify changes then I will update my PR

shriramshankar: Can you test this locally or in a review app to see if this is the only place to do the check.? Might have to change getReadReplicas in /config/configUtil.js.

harshkothari410: Yeah I will check in Review App.

harshkothari410:
Can you verify the behaviour ? @iamigo @shriramshankar I updated PR and I tested on review app.

config/configUtil.js
```diff
@@ -73,17 +73,20 @@ function csvToArray(str) {
...
         replicas.push(pe[replica]);
       }
     });
   }
 
+  if (!replicas.length) {
+    replicas = undefined;
+  }
+
   return replicas;
```
shriramshankar: If you just say replicas = undefined at line 76 and do not make anything other changes, it should take care i think.

harshkothari410: If you defined as undefined you can't push to that.

shriramshankar: What I am trying to say is, leave the old code as it is and Just add replicas = undefined at line 76. Replicas is assigned to [] at line 79 in the old code.

harshkothari410: So here is the problem with that case
Let's assume we have only one value in REPLICAS = test and there is no env variable name test. Then it will return null array instead of undefined and that will cause an error. So for handling that case I did this. only one parameter and that is bad one.

shriramshankar: Nice, got it.

shriramshankar: Do you want to add a test for this in tests/config/configUtil.js?

harshkothari410: added test @shriramshankar

tests/db/utils.js
```diff
@@ -12,7 +12,7 @@
 const expect = require('chai').expect;
 const u = require('../../db/utils');
 
-describe('db utils', () => {
+describe.only('db utils', () => {
```
iamigo: Must take out "only" before committing!

harshkothari410:
Yeah did it already @iamigo

PR #320, opened 2017-04-04
API: reject sample POST PUT or PATCH request, if request body contains name field.

annyhe:
a sample PATCH or POST or PUT should be rejected if there is a "name" field in the request body.
POST upsert and bulk upsert require there to be a "name" field in the request body for each sample.

### DIFF
--- a/api/v1/controllers/samples.js
+++ b/api/v1/controllers/samples.js
@@ -12,7 +12,7 @@
 'use strict'; // eslint-disable-line strict
 
 const featureToggles = require('feature-toggles');
-
+const apiErrors = require('../apiErrors');
 const helper = require('../helpers/nouns/samples');
 const subHelper = require('../helpers/nouns/subjects');
 const doDelete = require('../helpers/verbs/doDelete');
@@ -28,6 +28,7 @@ const sampleStoreConstants = sampleStore.constants;
 const redisModelSample = require('../../../cache/models/samples');
 const utils = require('./utils');
 const publisher = u.publisher;
+
 module.exports = {
 
   /**
@@ -98,7 +99,7 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   patchSample(req, res, next) {
-    utils.noReadOnlyFieldsInReq(req, helper);
+    utils.noReadOnlyFieldsInReq(req, helper.readOnlyFields);
     doPatch(req, res, next, helper);
   },
 
@@ -112,7 +113,7 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   postSample(req, res, next) {
-    utils.noReadOnlyFieldsInReq(req, helper);
+    utils.noReadOnlyFieldsInReq(req, helper.readOnlyFields);
     doPost(req, res, next, helper);
   },
 
@@ -127,7 +128,7 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   putSample(req, res, next) {
-    utils.noReadOnlyFieldsInReq(req, helper);
+    utils.noReadOnlyFieldsInReq(req, helper.readOnlyFields);
     doPut(req, res, next, helper);
   },
 
@@ -143,11 +144,14 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   upsertSample(req, res, next) {
-    utils.noReadOnlyFieldsInReq(req, helper);
+
+    // make the name post-able
+    const readOnlyFields = helper.readOnlyFields.filter((field) => field !== 'name');
+    utils.noReadOnlyFieldsInReq(req, readOnlyFields);
     const resultObj = { reqStartTime: new Date() };
     const sampleQueryBody = req.swagger.params.queryBody.value;
 
-    u.getUserNameFromToken(req,
+    return u.getUserNameFromToken(req,
       featureToggles.isFeatureEnabled('enforceWritePermission'))
     .then((userName) => {
       if (sampleQueryBody.relatedLinks) {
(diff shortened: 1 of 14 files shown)


api/v1/controllers/utils.js
```diff
@@ -46,6 +46,19 @@ function hasDuplicates(tagsArr) {
 }
 
 /**
+ * Check if name field is in the object.
+ * Throws validation error if object contains name field.
+ * @param  {Object} obj - Request object
+ */
+function rejectIfNameInBody(obj) {
```
iamigo: I think this function is only useful in the sample model. For every other model, you would just use the "noReadOnlyFieldsInReq" function. This one, presumably, is special because the name field is read-only for most things but it's NOT read-only for **upserts**.
So... maybe pull this fn back into the samples controller instead of utils, since it should never be used by any other model?

tests/api/v1/samples/upsert.js
```diff
@@ -63,6 +63,26 @@ describe(`api: POST ${path}`, () => {
   afterEach(u.forceDelete);
   after(tu.forceDeleteUser);
 
+  it('name field is required', (done) => {
```
iamigo: Can you add a couple more tests here (and for bulk upsert, too):
- upsert a sample whose name refers to an aspect which doesn't exist
- upsert a sample whose name refers to an aspect which is not published
- upsert a sample whose name refers to a subject which doesn't exist
- upsert a sample whose name refers to a subject which is not published

annyhe: ok, I'll write these tests to expect fail due to NOT_FOUND

annyhe:
@iamigo @shriramshankar added the suggested tests. Existing tests covered most of the non-existent subject/aspect, and un published aspect cases.

api/v1/controllers/samples.js
```diff
@@ -128,6 +144,7 @@ module.exports = {
    */
   putSample(req, res, next) {
     utils.noReadOnlyFieldsInReq(req, helper);
+    rejectIfNameInBody(req.body);
     doPut(req, res, next, helper);
```
shriramshankar: what resuing in noReadOnlyFieldsinReq ? The "noReadOnlyFieldsinReq" was created for to check for any read-only fields in requests.

annyhe: bulkUpsert and upsert also use noReadOnlyFieldsInReq. Adding the name field would reject upserts and bulk upserts

shriramshankar: you should be able to pass in the fields that you want to check to the "noReadIOnlyFieldsInReq" function. So, for upsert and bulkUpsert do not pass the name filed to the function.

annyhe:
@shriramshankar updated with changes. Please review.

api/v1/controllers/samples.js
```diff
@@ -143,11 +144,14 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   upsertSample(req, res, next) {
-    utils.noReadOnlyFieldsInReq(req, helper);
+
+    // make the name post-able
+    const readOnlyFields = helper.readOnlyFields.filter((field) => field !== 'name');
```
shriramshankar: Can we just say something like delete helper.readOnlyFields.name and pass the rest?

api/v1/controllers/samples.js
```diff
@@ -197,7 +201,10 @@ module.exports = {
    *  bulk upsert request has been received.
    */
   bulkUpsertSample(req, res/* , next */) {
-    utils.noReadOnlyFieldsInReq(req, helper);
+
+    // make the name post-able
+    const readOnlyFields = helper.readOnlyFields.filter((field) => field !== 'name');
+    utils.noReadOnlyFieldsInReq(req, readOnlyFields);
     const resultObj = { reqStartTime: new Date() };
```
shriramshankar: @iamigo MIssed this while reviewing the previous PR for readOnlyFields. Do we really want to go through the bulk upsert array before calling the "bulkupsert" function?

api/v1/controllers/samples.js
annyhe: readOnlyFields is an array. Delete doesn't work on arrays.

api/v1/controllers/samples.js
iamigo: This one here is just individual sample upsert, so keep it, but we should get rid of line 207. We still need to check for that stuff before doing each individual upsert, but we should not be iterating over the bulk upsert array here.

annyhe: I am confused: are we going to remove line 207 in this pull request? If so, we'll need to add checks to both 
        redisModelSample.bulkUpsertSample(value, userName)  and 
        helper.model.bulkUpsertByName(value, userName).

Even if we remove line 207, we still need to iterate over the bulk upsert array somewhere in the code, correct? 
Where were we moving the noReadOnlyFieldsInReq to?

iamigo: We only want to iterate over the array of samples AFTER sending back the "OK" response--otherwise we are holding up the web dyno.

shriramshankar: @annyhe @iamigo,  for now,  do we want to take out line 207 and merge the PR. This comment was not related to this story. Maybe we can create another story to handle it.

iamigo: I agree with @shriramshankar -- let's remove line 207 and merge this now, then follow up in separate story

annyhe:
merging is blocked due to slight decrease in coverage: 0.05%. @iamigo can you go ahead and merge this?

iamigo:
not blocked, just informational

tests/api/v1/samples/upsertBulk.js
```diff
@@ -297,58 +297,6 @@ describe('api: POST ' + path, () => {
     });
   });
 
-  it('upsert with readOnly field status should fail', (done) => {
```
iamigo: maybe just ".skip" these now since we'll need these tests again once we do the story you just created?

annyhe:
@iamigo update to skip the tests

iamigo:
OK, so maybe just stash the tests for whoever works on the next story

PR #321, opened 2017-04-04
write permissions for sample PUT, POST, PATCH and DELETE

shriramshankar:


cache/models/samples.js
```diff
@@ -586,15 +616,31 @@ module.exports = {
    * response. Note: Message body and message code will be updated if provided,
    * else no fields for message body/message code in redis object.
    * @param  {Object} params - Request parameters
+   * @param {String} userName - The user performing the write operation
    * @returns {Promise} - Resolves to a sample object
    */
-  patchSample(params) {
+  patchSample(params, userName) {
     const sampleName = params.key.value;
+    const aspectName = sampleName.split('|')[ONE];
```
iamigo: Is it possible to arrive at this point in the code with an invalid sampleName, e.g. a sampleName with no "|" character, or other ?

cache/models/samples.js
```diff
@@ -737,15 +784,31 @@ module.exports = {
    * if needed. Delete message body and message code fields from hash
    * if not provided because they will be null. Then update sample.
    * @param  {Object} params - Request parameters
+   * @param {String} userName - The user performing the write operation
    * @returns {Promise} - Resolves to a sample object
    */
-  putSample(params) {
+  putSample(params, userName) {
     const sampleName = params.key.value;
+    const aspectName = sampleName.split('|')[ONE];
```
iamigo: same question as above

cache/sampleStore.js
```diff
@@ -167,7 +167,8 @@ function isSampleWritable(aspectModel, aspectName, userName) {
   const options = {};
   options.where = { name: { $iLike: aspectName } };
   return aspectModel.findOne(options)
```
iamigo: Are we going to the db every time here?

shriramshankar: Only when the write permission is enabled.

cache/models/samples.js
shriramshankar: Yes,  it is possible. I will add the validation.

cache/sampleStore.js
shriramshankar: @iamigo Updated PR and will create a story for caching the related users.

cache/models/samples.js
```diff
@@ -523,14 +538,37 @@ module.exports = {
...
+    if (sampAspArr.length < TWO) {
+      throw new redisErrors.ResourceNotFoundError({
+        explanation: 'Incorrect sample name.',
+      });
+    }
+
+    const aspectName = sampAspArr[ONE];
     let currSampObj;
     let aspectObj;
+    const hasWritePerm = featureToggles
```
iamigo: This same line is repeated multiple times--make it a fn?

shriramshankar:
PR updated. Tested these changes manually also, to make sure they work.

pallavi2209:
Sorry, I missed the review request. Looking at it now.

pallavi2209:
The changes look good. Just one thing to confirm. We do not enforce write permission on creating sample. Correct?

iamigo:
> We do not enforce write permission on creating sample. Correct?

We are supposed to be enforcing based on Aspect's writers. @shriramshankar

shriramshankar:
@iamigo @pallavi2209 PR updated. Perms for POST sample needed to be done for both db and cache layer.

shriramshankar:
@iamigo The changes for upsert/bulkUpsert went in some time back and the tests for that went in along with the changes.   

FYI: few minutes I made a few changes to correct some typos.

iamigo:
@shriramshankar test failure?

  1) redis: aspect: create:  "after each" hook: forceDelete for "when name changes, the sampleStore refelct this change":
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

shriramshankar:
@iamigo Resolved the conflict. The test failure was resolved once I restarted the build again.

PR #326, opened 2017-04-10
move data to db or cache based on the change of state of enableRedisStore flag

shriramshankar:
move data to db or cache based on the change of state of enableRedisSampleStore flag

### DIFF
--- a/cache/sampleStoreInit.js
+++ b/cache/sampleStoreInit.js
@@ -18,13 +18,13 @@ const Subject = require('../db').Subject;
 const featureToggles = require('feature-toggles');
 const redisClient = require('./redisCache').client.sampleStore;
 const samsto = require('./sampleStore');
-const winston = require('winston');
+const log = require('winston');
 const samstoPersist = require('./sampleStorePersist');
 const constants = samsto.constants;
 
 /**
- * Deletes the key that does the previous value of the "enableRedisSampleStore"
- * flag.
+ * Deletes the previousStatuskey (that stores the previous value of the
+ * "enableRedisSampleStore" flag).
  * @returns {Promise} - which resolves to 0 when the key is deleted.
  */
 function deletePreviousStatus() {
@@ -59,7 +59,9 @@ function eradicate() {
       Promise.resolve(true);
     }));
   return deletePreviousStatus()
-    .then(() => Promise.all(promises));
+    .then(() => Promise.all(promises))
+    .then(() => log.info('Sample Store eradicated from cache :D'))
+    .then(() => true);
 } // eradicate
 
 /**
@@ -74,6 +76,8 @@ function populateAspects() {
     },
   })
   .then((aspects) => {
+    const msg = `Starting to load ${aspects.length} aspects to cache :|`;
+    log.info(msg);
     const aspectIdx = [];
     const cmds = [];
     aspects.forEach((a) => {
@@ -83,7 +87,9 @@ function populateAspects() {
     });
 
     cmds.push(['sadd', constants.indexKey.aspect, aspectIdx]);
-    return redisClient.batch(cmds).execAsync();
+    return redisClient.batch(cmds).execAsync()
+      .then(() => log.info('Done loading aspects to cache :D'))
+      .then(() => true);
   })
   .catch(console.error); // eslint-disable-line no-console
 } // populateAspects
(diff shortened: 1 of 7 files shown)


shriramshankar:
@iamigo tested this on staging, there are a few things we need to talk about before calling it done.

shriramshankar:
@iamigo @annyhe Can we not merge this until I talk to Ian about something. ?

shriramshankar:
List of thing to talk about
1. sample store is keyed off of sample name, assuming that the same name is unique in the database. Looks like sample names are not unique. ~5K difference between SELECT COUNT(DISTINCT name) FROM samples and SELECT COUNT(name) FROM samples. 
2. eradicate function works fine, but for large datasets  from prod "keys samto:su*" returns ~1700 always.

shriramshankar:
@iamigo @annyhe This is ready to be reviewed. Once the data issue is fixed on the sample table, we have to make sure that persistToDb function does its job right and not miss the sample for PHX.SP3 as shown in the image below.

image

cache/sampleStoreInit.js
```diff
@@ -160,63 +193,76 @@ function populateSamples() {
...
+       */
+      if (currentStatus) {
+        log.info('"enableRedisSampleStore" flag was switched to true,' +
+          ' so populating the cache from db');
+        return populate();
+      }
+
+      log.info('"enableRedisSampleStore" flag was switched to false' +
+          ' so persisting to db from cache. The cache will be eradicated ' +
+          'after the samples are persisted to db');
```
shriramshankar: I have added quite a bit of logging. These get logged when running the tests too, so if we think that it is too much, I can remove some.

cache/sampleStorePersist.js
```diff
@@ -17,21 +17,38 @@ const Sample = require('../db').Sample;
...
+  })
+  .then((keys) => keys.map((key) => ['hgetall', key]))
+  .then((cmds) => redisClient.batch(cmds).execAsync())
+  .then((res) => {
+    const samplesToCreate = res.map((sample) => {
+      sample.relatedLinks = JSON.parse(sample.relatedLinks);
+      return sample;
+    });
+    return Sample.bulkCreate(samplesToCreate);
+  })
```
shriramshankar: If fixing the data issue on samples does not fix the PHX.SP3 samples problem, we need to change this to Sample.bulkUpsertByName (but this takes about 5min to 10mins running locally). Considering that none of the hooks are called for bulkCreate, I am surprised that it even works for us.

cache/sampleStoreInit.js
```diff
@@ -18,10 +18,29 @@ const Subject = require('../db').Subject;
 const featureToggles = require('feature-toggles');
 const redisClient = require('./redisCache').client.sampleStore;
 const samsto = require('./sampleStore');
-const winston = require('winston');
+const log = require('winston');
+const samstoPersist = require('./sampleStorePersist');
 const constants = samsto.constants;
 
 /**
+ * Deletes the key that does the previous value of the "enableRedisSampleStore"
```
annyhe: did you mean: delete the key whose previous value was enableRedisSampleStore?

cache/sampleStoreInit.js
```diff
@@ -39,7 +58,10 @@ function eradicate() {
       console.error(err); // eslint-disable-line no-console
       Promise.resolve(true);
```
annyhe: do we want to do Promise.resolve(true) or return true?

tests/cache/sampleStoreFlagFlipFlop.js
```diff
@@ -0,0 +1,235 @@
...
+
+  it('flag from false to false: cache should not be populated', (done) => {
+    samstoinit.eradicate()
+    .then(() => tu.toggleOverride(sampleStore.constants.featureName, false))
+    .then(() => samstoinit.init())
+    .then((res) => expect(res).to.be.false)
+    .then(() =>
+      redisClient.smembersAsync(sampleStore.constants.indexKey.aspect))
+    .then((res) => {
+      expect(res).to.have.length(0);
```
annyhe: you can return the promise here, instead of doing another .then
return  redisClient.smembersAsync(sampleStore.constants.indexKey.sample))
})
.then((res) => {
expect...
return ...
})

shriramshankar:
@annyhe @iamigo Updated the PR

PR #328, opened 2017-04-11
Add subject id aspect Id to to sample store results

annyhe:

PR #329, opened 2017-04-11
Adding explanation to validation errors in lenses

pallavi2209:


api/v1/controllers/lenses.js
```diff
@@ -39,7 +39,9 @@ function updateLensDetails(seqObj) {
     if (seqObj.hasOwnProperty('sourceName')) {
       seqObj.name = seqObj.sourceName;
     } else {
-      throw new apiErrors.ValidationError();
+      throw new apiErrors.ValidationError({
+        explanation: 'name or source name is required in lens json.',
```
iamigo: I think name is always required in lens.json, but sourceName is not a valid option for lens.json. Users should never be posting sourceName--that's for us internally, i.e. if a user overrides the name when posting the lens, then we store the name from lens.json in the "sourceName" column in the db.

pallavi2209: Ok, let me check this, add tests and update description.

pallavi2209:
@iamigo updated the PR, but travis failing because of some react module, which is resolved by @annyhe in #328

iamigo:
still waiting on @annyhe review?

PR #332, opened 2017-04-12
change aspect time fields  to use ISO format

annyhe:


annyhe:
@iamigo The fields for operations on cache are identical to the API operations, except the order of two fields are reversed. 
I will file a bug to keep track of this.

PR #334, opened 2017-04-14
add to swagger so PUT perspectives/key updates the filterType

shriramshankar:
this allows the user to change the INCLUDE and EXCLUDE filter from the UI


annyhe:
Shriram put the pull request through

PR #337, opened 2017-04-14
bug: PATCH perspective without filter array throws invalid perspective error

annyhe:
- this PR fixes the bug
- includes test

### DIFF
--- a/api/v1/helpers/verbs/doPatch.js
+++ b/api/v1/helpers/verbs/doPatch.js
@@ -59,6 +59,10 @@ function doPatch(req, res, next, props) {
         // does give the needed effect; so adding it here!!!.
         if (props.modelName === 'Sample') {
           o.changed('value', true);
+        } else if (props.modelName === 'Perspective') {
+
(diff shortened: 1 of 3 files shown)


api/v1/helpers/verbs/doPatch.js
```diff
@@ -68,6 +69,11 @@ function doPatch(req, res, next, props) {
 
   patchPromise
   .then((retVal) => {
+    if (props.modelName === 'Perspective') {
+      helper.validateFilterAndThrowError(retVal.get());
+
+    }
+
```
shriramshankar: Does this need to happen here or in line 46  ?

annyhe: the line needs to be there, since it needs the updated perspective, not only the request body. Ie. if aspectTags = [], a request to PATCH with aspectTagFilter='INCLUDE' should fail.

annyhe:
@shriramshankar amended to move the check before the save.

PR #339, opened 2017-04-17
Check enforce token toggle when checking for write permission

pallavi2209:
@shriramshankar The bulk upsert request was getting rejected because of invalid token when checking for write permission, and hence we were not able to push samples from Focus in Staging. We need to resolve getUserNameFromToken  promise to true if token is not enforced.


pallavi2209:
Closing this and creating a story to figure out what should be done.

PR #340, opened 2017-04-18
Cascade delete samples when subject is deleted or unpublished

iamigo:


shriramshankar:
Are we doing migrations and deleting the samples in cache as a part of a different story?

iamigo:
> migrations

Will be a separate PR.

> deleting the samples in cache

Will add that to this PR today.

db/model/subject.js
```diff
@@ -431,6 +425,11 @@ module.exports = function subject(seq, dataTypes) {
         } else if (inst.previous('isPublished')) {
           // Treat unpublishing a subject as a "delete" event.
           common.publishChange(inst, eventName.del);
+
+          if (featureToggles.isFeatureEnabled(sampleStoreFeature)) {
+            subjectUtils.removeFromRedis(inst.absolutePath);
+          }
```
shriramshankar: Do we need this here too? I think the call to removeFromRedis at 354 will take care of this case.

iamigo: That is true--just running tests now and will update in a moment

PR #342, opened 2017-04-20
datamigration: delete samples of soft deleted and unpublised subjects

shriramshankar:


shriramshankar:
We need all the subjects (published = true|false, isDeleted = 0| not 0), we just do not need "samples" attached to the subjects that are either soft deleted or unpublished.

pallavi2209:
Oh its an OR, got it. Thanks!

PR #343, opened 2017-04-24
without filters, get hierarchy url contains no queries

annyhe:
from the perspective app, the url to load hierarchy would contain no query string, if all the perspective filters are empty.
- renamed a file
- add tests for query builder function getFilterQuery

### DIFF
--- a/view/perspective/utils.js
+++ b/view/perspective/utils.js
@@ -6,11 +6,61 @@
  * https://opensource.org/licenses/BSD-3-Clause
  */
 /**
- * view/perspective/configCreatePerspective.js
+ * view/perspective/utils.js
  *
  * JSON config for CreatePerspective.js
  * Includes all config for resources that need special treatment
  */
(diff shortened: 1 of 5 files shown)


view/perspective/utils.js
```diff
@@ -6,11 +6,61 @@
...
+/**
+ * Generate the filter string for the hierarchy API GET.
+ *
+ * @param {Object} p - The perspective object
+ * @returns {String} - The query string created generated based on the
+ *  perspective filters
+ */
+function getFilterQuery(p) {
+  let q = '';;
+
```
annyhe: i'll remove the extra semicolon

shriramshankar:
@annyhe it looks good. Can you confirm this thing works for the existing perspectives?

pallavi2209:
@iamigo Waiting for the build to pass. We can merge these changes then.

PR #347, opened 2017-04-29
Connection pooling

iamigo:


iamigo:
Tested on refocus-devbox by sending 5k jobs in 10s (i.e. a bulk upsert job every 2ms), and waiting to see how long it takes to finish processing all the enqueued work. Repeated 5 times with no conn pool overrides (i.e. sequelize default pool max=5, min=0), 4 times with conn pool max=12, min=8.

With no conn pool overrides: 23s, 15s, 76s (and some POST requests timed out), 24s, 25s.
With larger conn pool: 17s, 25s, 20s, 21s.

We'll need to tune in prod based on our actual load.

iamigo:
@pallavi2209 thx for asking about feature flag--we always need to be thinking like that. That being said, I did have a feature toggle for this, but when I wrote the test I realized that sequelize is already creating a pool, so there was no reason to that, so now the DB_CONNECTION_POOL_* env vars just override those defaults.

PR #348, opened 2017-04-30
Add kue stats logging

iamigo:


clock/scheduledJobs/kueStatsActivityLogs.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Generate the logging object. (Separating this out and exporting for easier
+ * testing.)
+ *
+ * @returns {Object} logging object
+ */
+function generateLogObject() {
+  const obj = JSON.parse(JSON.stringify(k));
+  jobQueue.card('active', (err, n) => {
+    if (!err) {
```
shriramshankar: Is "card" an api exposed by the kue library? Looks like they also have something like this https://github.com/Automattic/kue#queue-maintenance too to get the counts.

iamigo: the "inactiveCount" and "activeCount" etc. functions all call "card" ... their "card" function is exported, but I guess since it is not doc'd maybe it's considered not part of their public API... I'll change this to use the documented functions

iamigo:
any other blockers on this @shriramshankar @harshkothari410 ?

PR #341, opened 2017-04-20
Cache aspect writers

shriramshankar:

PR #349, opened 2017-05-01
changes to eradicate keys prefixed with smsto:subaspmap

shriramshankar:


cache/sampleStoreInit.js
```diff
@@ -41,15 +64,20 @@ function getPreviousStatus() {
...
  * @returns {Promise} upon completion.
  */
 function eradicate() {
   const promises = Object.getOwnPropertyNames(constants.indexKey)
     .map((s) => redisClient.smembersAsync(constants.indexKey[s])
     .then((keys) => {
+      if (constants.indexKey[s] === constants.indexKey.sample) {
+        // this is done to delete keys prefixed with "samsto:subaspmap:"
+        keys.push(...getSubAspMapKeys(keys));
+      }
```
shriramshankar: this is needed to because we do not have a master list of "subaspmap".  Maybe we can start tracking the "subaspmap" keys in a master list too? 

I think we can just get away by not tracking the mapping of the subjectAbsolutePath to the aspects.  

Any thoughts?

shriramshankar:
This needs to be merged before #341

cache/sampleStoreInit.js
iamigo: Yeah, maybe add another master list of these keys. But this is good for now, to not block enabling the feature.

PR #353, opened 2017-05-03
first commit for adding createdBy fields to resources.

annyhe:
- aspect and lenses tests, when request includes tokens

### DIFF
--- a/api/v1/helpers/verbs/doPost.js
+++ b/api/v1/helpers/verbs/doPost.js
@@ -52,10 +52,19 @@ function doPost(req, res, next, props) {
     } else {
       postPromise = authUtils.getUser(req)
       .then((user) => {
-        toPost.createdBy = user.id;
+        if (user) {
+          toPost.createdBy = user.id;
+        }
+
         return props.model.create(toPost);
       })
-      .catch((err) => u.handleError(next, err, props.modelName));
+      .catch((err) => {
+        if (err.status === 403) {
+          return props.model.create(toPost);
+        } else {
+          u.handleError(next, err, props.modelName);
+        }
+      });
     }
   }
 
--- a/api/v1/helpers/verbs/doPost.js
+++ b/api/v1/helpers/verbs/doPost.js
@@ -18,6 +18,7 @@ const httpStatus = require('../../constants').httpStatus;
 const constants = require('../../../../cache/sampleStore').constants;
 const redisModelSample = require('../../../../cache/models/samples');
 const featureToggles = require('feature-toggles');
+const authUtils = require('../authUtils');
 
 /**
  * Creates a new record and sends it back in the json response with status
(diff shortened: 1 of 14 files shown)


api/v1/helpers/verbs/doPost.js
```diff
@@ -45,9 +46,17 @@ function doPost(req, res, next, props) {
...
-    postPromise = featureToggles.isFeatureEnabled('enforceWritePermission') &&
-      props.modelName === 'Sample' ? u.createSample(req, props) :
-      props.model.create(toPost);
+    if (featureToggles.isFeatureEnabled('enforceWritePermission') &&
+      props.modelName === 'Sample') {
+      postPromise = u.createSample(req, props);
+    } else {
+      postPromise = authUtils.getUser(req)
+      .then((user) => {
+        toPost.createdBy = user.id;
```
iamigo: will throw error when no user found -- should just leave createdBy empty when no user record found

api/v1/swagger.yaml
```diff
@@ -6950,6 +6951,7 @@ definitions:
       #   description: >
       #    The number of direct children subjects under this subject.
       createdBy:
+        type: string
```
iamigo: Are you sure? Can createdBy be type string AND have a ref to UsersResponse?

api/v1/helpers/verbs/utils.js
```diff
@@ -565,7 +565,7 @@ function cleanAndStripNulls(obj) {
 
       // if undefined, parentAbsolutePath needs to be set to empty string,
       // to pass swagger's schema validation
-      if (key === 'parentAbsolutePath' && !o[key]) {
+      if ((key === 'parentAbsolutePath' || key === 'createdBy') && !o[key]) {
```
iamigo: Why this exception to our convention of removing null/empty attributes?

annyhe: don't we always want to have a createdBy field in the response, even if the value is ''?

pallavi2209:
@annyhe Can you please elaborate what are you trying to do in this PR? Do you intend createdBy to be a user object or string. Do these changes need to be behind a toggle? Why are some tests skipped? etc

api/v1/helpers/verbs/utils.js
iamigo: I think we should follow our normal pattern of omitting attributes with null or empty values, unless we have a compelling reason

annyhe:
The aim of the story is to add 
{ createdBy: userId, user: { name: userName }} to the response object. 
These changes will be behind a toggle. Will update the PR with more focused tests.

api/v1/helpers/verbs/utils.js
annyhe: To clarify, do we want to have a createdBy field when there is a token provided, otherwise not return a createdAt field? That seems a bit inconsistent. Don't we always want to return a createdAt field, similar to how we always want a parentAbsolutePath, or parentId?

annyhe:
How do we add feature toggle in db models? Default scope need to be changed to return the user association. The api tests fail with the user association when there is no userId provided.

PR #356, opened 2017-05-04
delete samples in cache and in db when subject absolutepath changes

shriramshankar:


shriramshankar:
the renameKeys function in redisOps.js is not used any more but I guess we can keep it , if we need it for something else in the future.

PR #358, opened 2017-05-04
Two minor adjustments to the sample store init and persist

iamigo:
(1) When populating the sample store from the db populate subjects and aspects FIRST, THEN samples.
(2) When persisting the sample store from redis to db, skip samples if missing aspectId/subjectId


tests/cache/sampleStoreFlagFlipFlop.js
```diff
@@ -23,7 +23,7 @@ const Sample = tu.db.Sample;
 const initialFeatureState = featureToggles
   .isFeatureEnabled(sampleStore.constants.featureName);
 
-describe('sampleStore (feature off):', () => {
+describe.only('sampleStore (feature off):', () => {
```
pallavi2209: Do we need this only?

PR #360, opened 2017-05-04
added UI token if the request is made from UI

harshkothari410:


harshkothari410:
I have tested locally in Heroku env but let me test on review app

harshkothari410:
I tested in review app as well. But you can check once in logs. @iamigo

harshkothari410:
Please review it @iamigo and @pallavi2209 so that it will be closed before I go

PR #362, opened 2017-05-06
check for read-only sample fields when doing bulk upsert

shriramshankar:


worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -38,9 +39,11 @@ module.exports = (job, done) => {
 
   let bulkUpsertPromise;
   if (featureToggles.isFeatureEnabled('enableRedisSampleStore')) {
```
iamigo: Maybe this is a little cleaner... ?

```
const sampleModel = featureToggles.isFeatureEnabled('enableRedisSampleStore')) ?
    cacheSampleModel : helper.model;
sampleModel.bulkUpsertByName(samples, userName, readOnlyFields);
```

tests/cache/models/samples/upsertBulk.js
```diff
@@ -290,6 +290,62 @@ describe('api::redisEnabled::POST::bulkUpsert ' + path, () => {
...
+    .end((err, res) => {
+      if (err) {
+        return done(err);
+      }
+      expect(res.body.status).to.equal('OK');
+      return done();
+    });
+  });
+
+  it('samples with read only fields in them should not be upserted', (done) => {
```
iamigo: Do we reject the whole batch if even one sample has a read-only field?

api/v1/controllers/samples.js
```diff
@@ -216,14 +216,14 @@ module.exports = {
         wrappedBulkUpsertData.upsertData = value;
         wrappedBulkUpsertData.userName = userName;
         wrappedBulkUpsertData.reqStartTime = reqStartTime;
-
+        wrappedBulkUpsertData.readOnlyFields = helper.readOnlyFields;
         const j = jobWrapper.createJob(jobType.BULKUPSERTSAMPLES,
           wrappedBulkUpsertData, req);
       } else {
         const bulkUpsertPromise =
         featureToggles.isFeatureEnabled(sampleStoreConstants.featureName) ?
```
iamigo: Maybe this is a little cleaner... ?

```
const sampleModel = featureToggles.isFeatureEnabled(sampleStoreConstants.featureName) ?
    redisModelSample : helper.model;
sampleModel.bulkUpsertByName(value, userName, readOnlyFields);
```

tests/cache/models/samples/upsertBulk.js
shriramshankar: No, just rejecting the ones with the read-only field in them. I can update the test to show that.

tests/cache/models/samples/upsertBulk.js
```diff
@@ -290,6 +286,67 @@ describe('api::redisEnabled::POST::bulkUpsert ' + path, () => {
...
+       */
+      setTimeout(() => {
+        api.get('/v1/samples?name=' +
+        `${tu.namePrefix}Subject|${tu.namePrefix}Aspect*`)
+        .end((err, res) => {
+          if (err) {
+            done(err);
+          }
+          expect(res.body).to.have.length(2);
+          expect(res.body[0].value).to.not.equal('10');
```
annyhe: what is the value of the res.body[0].value?

annyhe: Also, should we check for errorCount?

shriramshankar: There is no errorCount field on the GET samples endpoint. The value of res.body[0].value is set by the test that was run before this.

PR #364, opened 2017-05-10
pass the right set of readonly fields to the workers

shriramshankar:


shriramshankar:
found a mistake in the read-only field story that I did. Corrected it here.

PR #366, opened 2017-05-11
return errors, if any, when a bulkUpsert job is done processing

shriramshankar:


shriramshankar:
@iamigo @annyhe These are changes to the bulkUpsert workers, to make it return the errors if any. Tests will be added when I make the rest of the changes in the next PR.

annyhe:
@shriramshankar would be nice to have tests of the bulk upsert, to see the error array that is returned

shriramshankar:
@annyhe yes, will add this test too in the next PR. Thank you.

PR #367, opened 2017-05-11
return createdBy field on POST.  with samples tests

annyhe:
- for post /samples WITH and WITHOUT cache
- for upsert /samples WITHOUT cache


api/v1/controllers/samples.js
```diff
@@ -150,13 +151,25 @@ module.exports = {
     const resultObj = { reqStartTime: new Date() };
     const sampleQueryBody = req.swagger.params.queryBody.value;
 
-    return u.getUserNameFromToken(req,
-      featureToggles.isFeatureEnabled('enforceWritePermission'))
-    .then((userName) => {
+    return authUtils.getUser(req)
+    .then((user) => {
```
shriramshankar: What happens when req is sent without the authtoken?

shriramshankar:
@annyhe Since, we are supporting the sample in db and cache as well, the populateSamples method in sampleStoreInit needs to be changed to get the user list populated.

api/v1/controllers/samples.js
```diff
@@ -150,13 +151,25 @@ module.exports = {
...
         u.checkDuplicateRLinks(sampleQueryBody.relatedLinks);
       }
 
+      if (featureToggles.isFeatureEnabled('returnCreatedBy') && user) {
+        sampleQueryBody.createdBy = user.id;
+      }
+
+      let userName = '';
+      if (!featureToggles.isFeatureEnabled('enforceWritePermission')) {
+        userName = true;
```
shriramshankar: we will be removing the "enforceWritePermission" probably the sprint starting 5/15. I think your changes will be much cleaner if you do it without checking for this flag.

api/v1/helpers/verbs/doPost.js
```diff
@@ -42,9 +42,21 @@ function doPost(req, res, next, props) {
...
-    postPromise = u.getUserNameFromToken(req,
-      featureToggles.isFeatureEnabled('enforceWritePermission'))
-      .then((user) => redisModelSample.postSample(req.swagger.params, user));
+    postPromise = authUtils.getUser(req)
+    .then((user) => {
+      const userObject = user && featureToggles.isFeatureEnabled('returnCreatedBy') ?
+        { name: user.name, id: user.id, email: user.email } : false;
+      return redisModelSample.postSample(req.swagger.params, userObject);
+    })
+    .catch((err) => {
```
shriramshankar: When there is an error, whether if is with status 403 or not, I am wondering if it makes sense to just reject the request.

api/v1/controllers/samples.js
annyhe: if request is sent without authentication token, request will still be granted. The createdBy field will be stripped out because of its null value, so the response will NOT contain createdBy, user fields.

api/v1/helpers/verbs/doPost.js
annyhe: if token is not enforced, we should still grant the request.

tests/cache/models/samples/upsertBulkWithCreatedBy.js
```diff
@@ -0,0 +1,114 @@
...
+const Subject = tu.db.Subject;
+const path = '/v1/samples/upsert/bulk';
+const URL1 = 'https://samples.com';
+const relatedLinks = [
+  { name: 'link1', url: URL1 },
+  { name: 'link2', url: URL1 },
+];
+
+describe.only('api::redisEnabled::POST::bulkUpsert ' + path, () => {
+  let token;
```
shriramshankar: "only" here.

annyhe:
The reported decrease in coverage by coveralls does not make any sense: https://coveralls.io/builds/11530907
The files whose coverage reportedly decreased are not affected by this pull request.

annyhe:
removed .only and updated with minor lint changes. Please review @shriramshankar @pallavi2209

PR #368, opened 2017-05-12
Get hierarchy worker

jgraff2:


jgraff2:
@iamigo @shriramshankar 
I've made some changes to the tests and they are passing now.
Coverage is down in controllers/subjects.js because of error handling code that can't be covered in automated tests. I have tested it manually. Is this ok?

jobQueue/jobWrapper.js
```diff
@@ -229,8 +229,8 @@ function createPromisifiedJob(jobName, data, req) {
  */
 function createJob(jobName, data, req) {
   const jobPriority = calculateJobPriority(jobName, data, req);
-  const job = jobQueue.create(jobName, data)
-  .ttl(TIME_TO_LIVE)
+  const job = jobQueue.create(jobName, data);
+  job.ttl(TIME_TO_LIVE)
```
iamigo: why breaking this up into two separate statements?

jgraff2: This was causing errors for me. I wasn't able to figure out why, but this fixed the problem.

I've been running it locally. Can you tell me how to do that?

api/v1/controllers/subjects.js
```diff
@@ -249,32 +252,67 @@ module.exports = {
...
+      reqStartTime: Date.now(),
+      params: params,
+    };
+
+    if (featureToggles.isFeatureEnabled('enableWorkerProcess')
+    && featureToggles.isFeatureEnabled('enqueueHierarchy')) {
+      jobWrapper.createJob(jobType.GET_HIERARCHY, resultObj, req)
+      .ttl(WORKER_TTL)
+      .on('complete', (resultObj) => {
+        u.logAPI(req, resultObj, resultObj.retval);
```
shriramshankar: the jobWrapper.createJob also creates a on(''complete) listener. Wondering if we can do this without having to create 2 listeners for the same event. ? If there is no way around this, we want to make sure both the listeners are called.

api/v1/controllers/subjects.js
```diff
@@ -33,6 +34,9 @@ const apiErrors = require('../apiErrors');
 const redisSubjectModel = require('../../../cache/models/subject');
 const sampleStoreFeature =
                   require('../../../cache/sampleStore').constants.featureName;
+const jobType = require('../../../jobQueue/setup').jobType;
+const jobWrapper = require('../../../jobQueue/jobWrapper');
+const WORKER_TTL = 30000;
 const ZERO = 0;
```
shriramshankar: Can we add this ttl to the setup file instead? When we move some of the other API operations to the workers, all of them can use the same variable.

tests/cache/jobQueue/getHierarchy.js
```diff
@@ -0,0 +1,347 @@
...
+const Subject = tu.db.Subject;
+const path = '/v1/subjects/{key}/hierarchy';
+const logger = require('../../../utils/activityLog').logger;
+
+describe(`api: GET using worker process ${path}`, () => {
+
+  before(() => {
+    tu.toggleOverride('enableWorkerProcess', true);
+    tu.toggleOverride('enqueueHierarchy', true);
+    require('../../../worker/jobProcessor');
```
shriramshankar: I think requiring this will call all the worker process. Can you just call the bulkUpsert worker here by doing (jobQueue.process(jobType.BULKUPSERTSAMPLES, bulkUpsertSamplesJob) and probably do the same in the other tests too?

api/v1/controllers/subjects.js
jgraff2: They are both called. If the listener in jobWrapper isn't called, the logging test will fail.
I think it's better to do this here because the listener in jobWrapper is generic and this one is specific to getHierarchy jobs.

tests/cache/jobQueue/getHierarchy.js
jgraff2: I assume you mean jobType.GET_HIERARCHY?
I wrote this before I merged the change that split out the job processors so that never occurred to me.
Doesn't seem like it should matter much either way, but I think it's easier to understand this way since the idea is to simulate a normal worker process.

api/v1/controllers/subjects.js
jgraff2: My understanding was that the 30 sec timeout is specific to the getHierarchy job. Would it be the same for future API jobs?
@iamigo

tests/cache/jobQueue/getHierarchy.js
shriramshankar: Yes, I meant jobType.GET_HIERARCHY. Your tests should be starting only the worker process it needs and having it this way gives an impression that the "GET_HIERARCHY" job type needs all the workers to get the job done.

jgraff2: ok - updated.

api/v1/controllers/subjects.js
iamigo: 30 second timeout is universal in heroku, applies to everything.

> The countdown for this 30 second timeout begins after the entire request (all request headers and, if applicable, the request body) has been sent from the router to the dyno. The request must then be processed in the dyno by your application, and a response delivered back to the router, within 30 seconds to avoid the timeout.

jobQueue/setup.js
```diff
@@ -31,6 +31,7 @@ module.exports = {
     BULKUPSERTSAMPLES: 'bulkUpsertSamples',
     PERSIST_SAMPLE_STORE: 'PERSIST_SAMPLE_STORE',
     SAMPLE_TIMEOUT: 'SAMPLE_TIMEOUT',
+    GET_HIERARCHY: 'GET_HIERARCHY',
```
iamigo: pls. insert this in alphabetical order... when there are more of them it will make it easier to read

worker/jobProcessor.js
```diff
@@ -20,10 +20,12 @@ const jobQueue = require('../jobQueue/jobWrapper').jobQueue;
...
 const persistSampleStoreJob = require('./jobs/persistSampleStoreJob');
+const getHierarchyJob = require('./jobs/getHierarchyJob');
 const workerStarted = 'Worker Process Started';
 
 console.log(workerStarted); // eslint-disable-line no-console
 
 jobQueue.process(jobType.BULKUPSERTSAMPLES, bulkUpsertSamplesJob);
 jobQueue.process(jobType.SAMPLE_TIMEOUT, sampleTimeoutJob);
 jobQueue.process(jobType.PERSIST_SAMPLE_STORE, persistSampleStoreJob);
+jobQueue.process(jobType.GET_HIERARCHY, getHierarchyJob);
```
iamigo: pls. put these in alphabetical order, too

api/v1/controllers/subjects.js
jgraff2: ok so in that case should we have a 30 second timeout setting to be used in any jobs that send back a response, and the current default timeout to be used in asynchronous jobs?
Or can we just do 30 seconds for everything? Do bulkUpsert jobs need longer than that?

shriramshankar: BulkUpsert jobs have a much longer timeout. We can use 30 seconds for the GET requests that are handled by the workers.

jgraff2: ok

shriramshankar:
@jgraff2 Looks like there is a conflict in one of the files. We can merge it once the conflict is fixed.

iamigo:
@jgraff2

jgraff2:
@shriramshankar @iamigo 
Sorry for not giving an update - It's not quite ready to be merged.
I made a change for Shriram's request to move the ttl to the settings file, but when I tested it the timeout wasn't working at all. I eventually figured out that it was the recent change to the job order that broke it. I think it's something in kue and I want to look into it further but I put that on hold for now because I spent two days on this and I want to make sure I finish the other work this sprint.

It's just the timeout that's affected, which is a minor issue - If you want to get this merged sooner than later I could create a separate bug and work on that next sprint.

shriramshankar:
@jgraff2 @iamigo I am fine with merging this. Since we are setting the timeout to 30s it does not really matter if we are sending the 503 from the dyno or Heroku is sending it from the routers.

jobQueue/setup.js
```diff
@@ -40,6 +40,7 @@ if (featureToggles.isFeatureEnabled('instrumentKue')) {
 module.exports = {
   jobConcurrency: {
     BULKUPSERTSAMPLES: conf.bulkUpsertSampleJobConcurrency,
+    GET_HIERARCHY: 1,
```
iamigo: Would you mind setting up a conf.getHierarchyJobConcurrency variable like the one for bulk upsert? It would be nice to be able to tune this from env. variables if needed.

jgraff2:
@iamigo @shriramshankar 
I've finished investigating the timeout issue. Changing the order of the jobs exposed the issue but there's more to it than that. The workaround we discussed yesterday doesn't work reliably. I don't think the timeout ever worked reliably for any of the job types. 

I think we should go ahead and merge this as is and address the timeout issue separately. I've resolved the conflicts.

iamigo:
`I don't think the timeout ever worked reliably for any of the job types` -- this could *also* be the root of our queue slowdown problem!!!

jgraff2:
I had that thought too. Let me know if you want to discuss so I can explain what's happening in more depth.

jobQueue/setup.js
jgraff2: ok will do.

PR #372, opened 2017-05-17
Fix subject hierarchy level on put

annyhe:


annyhe:
I tested through the UI too. This PR fixes the unintentionally setting hierarchyLevel to 1 error.

PR #374, opened 2017-05-19
Fix case sensitive bulk upsert non cache

annyhe:
- problem: user is able to update the sample name by bulk upsert
- solution: remove the sample name from the update sample object in db model.

### DIFF
--- a/tests/api/v1/samples/upsert.js
+++ b/tests/api/v1/samples/upsert.js
@@ -208,6 +208,26 @@ describe(`api: POST ${path}`, () => {
       });
     });
 
+    it('case-incorrect name becomes a combination of subject absolutePath' +
+      ' and aspect name', (done) => {
+      const sampleName = `${subject.absolutePath}|${aspect.name}`;
+      api.post(path)
+      .set('Authorization', token)
+      .send({
+        name: sampleName.toLowerCase(),
+        value: '2',
+      })
+      .expect(constants.httpStatus.OK)
+      .end((err, res) => {
+        if (err) {
+          done(err);
+        }
+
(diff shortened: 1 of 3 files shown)


tests/api/v1/samples/upsert.js
```diff
@@ -505,38 +505,4 @@ describe(`api: POST ${path}`, () => {
...
-    it('existing sample is not duplicated', (done) => {
-      api.get('/v1/samples?name=' + `${subject.absolutePath}|${aspect.name}`)
-      .end((err, res) => {
-        if (err) {
-          done(err);
-        }
-
-        expect(res.body).to.have.length(1);
-        expect(res.body[0].name)
-        .to.equal(`${subject.absolutePath}|${aspect.name}`.toLowerCase());
```
annyhe: remove the test that has incorrect expectations: sample name should NOT be updated

tests/api/v1/samples/upsertBulkCaseSensitive.js
```diff
@@ -0,0 +1,133 @@
...
+const expect = require('chai').expect;
+const supertest = require('supertest');
+const api = supertest(require('../../../../index').app);
+const tu = require('../../../testUtils');
+const u = require('./utils');
+const Aspect = tu.db.Aspect;
+const Subject = tu.db.Subject;
+const Sample = tu.db.Sample;
+const path = '/v1/samples/upsert/bulk';
+
```
shriramshankar: Can you add a test for the non-bulk upsert path (/v1/samples/upsert) too?

annyhe: will do, even though both would test the same code path.

PR #377, opened 2017-05-22
Fix case sensitive bulk upsert with redis cache

annyhe:
- problem: with cache turned on, the upserted sample name is set based on the request body
- sample name should be set according to subject absolutePath and aspect name
- for upsert/bulk upsert for both update and create, the sample name should not be updated
- change the redis cache: subjectAbsPath -> subjectId mapping to subjectAbsPath -> subject, to retrieve proper case subjectAbsPath on sample upsert

### DIFF
--- a/cache/sampleStore.js
+++ b/cache/sampleStore.js
@@ -31,7 +31,7 @@ const constants = {
       'writers',
     ],
     sample: ['relatedLinks'],
-    subject: ['aspectNames'],
+    subject: ['aspectNames', 'tags', 'relatedLinks', 'geolocation'],
   },
   indexKey: {
     aspect: PFX + SEP + 'aspects',
--- a/tests/cache/models/samples/upsert.js
+++ b/tests/cache/models/samples/upsert.js
@@ -184,6 +184,26 @@ describe(`api::redisEnabled::POST::upsert ${path}`, () => {
     });
   });
 
+  it('when the sample does not exist, name should match subject absolutePath,' +
+    ' aspect name', (done) => {
+    const sampleName = `${subject.absolutePath}|${aspect.name}`;
+    api.post(path)
+    .set('Authorization', token)
+    .send({
+      name: sampleName.toLowerCase(),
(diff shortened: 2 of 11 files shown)


tests/cache/models/samples/upsert.js
```diff
@@ -184,6 +184,26 @@ describe(`api::redisEnabled::POST::upsert ${path}`, () => {
     });
   });
 
+  it.skip('when the sample does not exist, name should match subject absolutePath,' +
```
annyhe: skipping the failing tests, that expect new sample's name to match subject absolutePath, aspect name.

tests/cache/models/redisTestUtil.js
```diff
@@ -21,6 +21,54 @@ const Sample = tu.db.Sample;
 const testStartTime = new Date();
 
 module.exports = {
+  populateRedisWithoutSample(done) {
+    let a1;
```
shriramshankar: Do you want to do the tests without using populateRedisWithoutSample? Like what we talked about some time back.

annyhe:
Updated to check the sample name on upsert, for both new and existing samples. Please review and comment @shriramshankar @pallavi2209

annyhe:
After involving everyone in the conversation, we decided to go forward with storing subjects in redis. Updated the pull request with edited comment. Please review @shriramshankar @pallavi2209

shriramshankar:
@annyhe We need to do some manual tests with a larger data set to make sure there are no regression. I will set up a time for us to do it.

cache/sampleStore.js
```diff
@@ -31,7 +31,7 @@ const constants = {
       'writers',
     ],
     sample: ['relatedLinks'],
-    subject: ['aspectNames'],
+    subject: ['aspectNames', 'tags', 'relatedLinks', 'geolocation'],
   },
```
annyhe: @shriramshankar and I found a bug. It is corrected here.

PR #383, opened 2017-05-26
Perspectives subjects query

iamigo:
The perspective page only needs subject absolutePath and tags to load the subject dropdown and the subject tags dropdown so use the "fields" query param.

### DIFF
--- a/api/v1/swagger.yaml
+++ b/api/v1/swagger.yaml
@@ -1635,7 +1635,7 @@ paths:
       summary: Update selected metadata fields for the specified lens
       tags: [ lenses ]
       description: >-
-        Update selected metadata fields for the specified lens. If the Refocus configuration parameter `useAccessToken` is set to `true`, you must include an `Authorization` request header with your [JSON Web Token](https://tools.ietf.org/html/rfc7519) (JWT) as the value. You can get a token using `POST /v1/register` or `POST /v1/tokens`.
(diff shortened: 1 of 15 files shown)


annyhe:
There's a few more lines to change in view/perspective, since subjects was an array of objects, now it is an array of strings. 
This is a good start. I can do another commit on top of this, or open another pull request.

annyhe:
Tested locally, it works. We can take out fields from other resources, but that is for another pull request

annyhe:
travis CI failed with this test that is unrelated to the pull request
Have you seen this before @pallavi2209 ?

  1) jobWrapper: functions  "before all" hook:
     Uncaught ReferenceError: job is not defined
      at jobQueue.create.ttl.priority.save.err (jobQueue/jobWrapper.js:9:8775)
      at Job.testJobSave [as save] (node_modules/kue/lib/queue/test_mode.js:16:28)
      at Object.createJob (jobQueue/jobWrapper.js:9:8383)
      at Timeout.enqueue [as _onTimeout] (clock/scheduledJobs/sampleTimeoutJob.js:9:1827)

annyhe:
Demo app is live at http://localhost:3000/perspectives/NorthAmerica
http://localhost:3000/perspectives/NorthAmerica?beta=true

The create/edit modal loads, and the subjects GET requests is of format http://refocus-pr-383.herokuapp.com/v1/subjects?fields=isPublished,absolutePath,tags

annyhe:
@pallavi2209 This pull request is NOT removing the query params. It is to improve performance by GETting only the isPublished, absolutePath, and the tags field of subject. Hence only the GET url is changed.

PR #384, opened 2017-05-26
create docker image

dbabenko:
Sharing my docker config in case you decide it can be useful for other users of Refocus who want to get refocus running in their local environments.

* added Dockerfile  to create docker image with Refocus app
* added docker-compose.yml file to start refocus app locally (along with postgresql and redis). 
     - right now it references image from my public repo so users don't need to build their own image and only need to run "docker-compose up" to get local refocus app running. But we can change publish image to any other repo of your choice.  
* doc changes - updated only *.md files. Did not generate html files.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,17 @@
+version: '2'
+services:
+  refocus:
+    image: "dbabenko/refocus"
+    ports:
+     - "3000:3000"
+    depends_on:
+     - "redis"
+     - "pg"
+  redis:
+    image: "redis:3.2.8"
+    expose:
+       - 6379
+  pg:
+    image: "postgres:9.6.2"
+    expose:
+       - 5432
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.dockerignore
+Dockerfile
+docker-compose.yml
+.git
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: '2'
+services:
+  refocus:
+    image: "dbabenko/refocus"
```
iamigo: does the image name here need your name?

dbabenko: No. And I hope we can replace it with the reference to official public refocus docker image. 

This is a reference to my public docker repo https://hub.docker.com/r/dbabenko/refocus/
that I created to verify my changes. If the image exists then users don't need to build the image locally - one can just run "docker-compose up" and images for refocus, redis, postgres will be downloaded automatically. If there is no image  then user would have to build the image locally, modify docker-compose.yml to point to their own preferred repo. 

A couple of options to consider:

1. Refocus team can set up public docker repo and publish docker image there similar to how other projects release their images. 
2. Remove docker-compose.yml comletely and just provide instruction how one can create one. 

Maybe we should go with option 2 to start and see if there will be any interest in official public docker image. If you prefer option 2 I can create a new PR without any docker-compose.yml and move its content into docs section.

iamigo: ```option 2 to start```

sounds like a good way to start!

PR #385, opened 2017-05-30
Bot model

tausifmuzaffar:
Creating models one at a time

https://drive.google.com/file/d/0B9BRb52RN_BFdnhRVm1Qd0phU1U/view?usp=sharing


db/model/botData.js
```diff
@@ -0,0 +1,56 @@
...
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * db/model/botData.js
+ *
+ * Bot webapps will have data that refocus will want to retrieve
+ * or store. This table will define the data information to be
+ * used manipulate the bots data.
+ */
```
kamilsmuga: What are examples of such data? I'm having a hard time to find examples that would tie to the Bot model. If they exist - maybe should be part of the Bot model available as fields? 
All examples I can think of have a relationship with a Room model.

db/model/botData.js
```diff
@@ -0,0 +1,56 @@
...
+
+const constants = require('../constants');
+
+const assoc = {};
+
+module.exports = function user(seq, dataTypes) {
```
kamilsmuga: function name - user - is incorrect.
Side note - I don't think we need to name these functions at all - anonymous would do. However, other models seems to assign name to functions

db/helpers/botUtils.js
```diff
@@ -0,0 +1,135 @@
+/**
```
iamigo: Needs additional tests as described here: https://coveralls.io/builds/11832728/source?filename=db%2Fhelpers%2FbotUtils.js

iamigo: The nice thing about separating out these validation fns and exporting them as functions is that you can test them outside of the context of a running db.

db/model/bot.js
```diff
@@ -0,0 +1,81 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * db/model/bot.js
+ *
+ * Bots are distributed webapps that will interact with our
+ * different routes. This object will be used to track
+ * and organize the bots that are associated with refocus.
```
iamigo: Can you add a little more information here about bot actions and data, how those fields are populated and used?

iamigo:
@tausifmuzaffar once we merge this, subsequent field changes must have migrations, too.

tausifmuzaffar:
@iamigo @pallavi2209 thanks! Do field additions need migrations as well or just field changes?

iamigo:
Just field changes

PR #392, opened 2017-06-02
Worker time logging

jgraff2:


config/activityLog.js
```diff
@@ -59,6 +59,7 @@ module.exports = {
       jobId: 'None',
       jobType: 'None',
       queueTime: 'None',
+      queueResponseTime: 'None',
```
jgraff2: This is the time it takes for the result to be sent back from the worker after the job has finished. The discrepancy with the total time was because we weren't accounting for this. My solution was to track it and add tests to make sure queueTime + workTime + queueResponseTime = totalTime.

PR #395, opened 2017-06-07
Add 404 page

annyhe:
on 404 error:
- remove spinner
- large text appears: "Sorry. Perspective [perspective_name] not found."
- perspective picker is loaded the same way as under normal circumstances

this PR also includes the fix:
previously, on edit, the perspective loaded into modal is the CURRENT perspective, 
not necessarily the perspective in the same row as the pencil icon.
The fix necessitated GETting all the perspectives on page load. Thus the dropdown shows up later
than previously

### DIFF
--- a/view/newPerspective/utils.js
+++ b/view/newPerspective/utils.js
@@ -241,7 +241,7 @@ function getValuesObject(accumulatorObject) {
     getPerspectiveUrl,
     handleHierarchyEvent,
     handleLensDomEvent,
-    handleError,
+    customHandleError,
   } = accumulatorObject;
   const constants = require('../../api/v1/constants');
   const httpStatus = constants.httpStatus;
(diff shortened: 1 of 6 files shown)


annyhe:
If we want to show the combox for the dropdown faster, we can load the 
aspects, lenses, and subjects ONLY when the user clicks on the edit/create button. We can explore this at another time.

annyhe:
before: 
screen shot 2017-06-07 at 2 26 05 pm

after:
screen shot 2017-06-07 at 2 25 19 pm

view/newPerspective/utils.js
```diff
@@ -337,40 +301,97 @@ function getValuesObject(request, getPerspectiveUrl, handleHierarchyEvent, handl
...
+     *  valuesObj.perspective = null
+     */
+     if (returnedPerspective) {
+        valuesObj.perspective = returnedPerspective;
+        valuesObj.name = valuesObj.perspective.name;
+     } else if (named) {
+
+        // named perspective does not exist
+        const name = url.split('/').pop();
+        customHandleError('Sorry. Perspective ' + name + ' not found.');
```
iamigo: "Sorry, but the perspective you were trying to load, _______, does not exist. Please try selecting a perspective from the dropdown."

annyhe:
This is what the amended error message looks like.
Admittedly the text is a bit wide. However we have designers so a revision to this is expected soon.
screen shot 2017-06-08 at 10 50 20 am

annyhe:
@iamigo Review app is live at https://refocus-devbox.herokuapp.com/perspectives/dd?beta=true

PR #399, opened 2017-06-12
Room model

tausifmuzaffar:
These models are the basis for the Room and Room Type 

https://drive.google.com/file/d/0B9BRb52RN_BFWVUtTlk3QWNXRm8/view?usp=sharing


kamilsmuga:
@tausifmuzaffar can you update the model diagram to refer to Rooms instead of Responses?

db/model/room.js
```diff
@@ -0,0 +1,62 @@
...
+const assoc = {};
+
+module.exports = function user(seq, dataTypes) {
+  const Room = seq.define('Room', {
+    id: {
+      type: dataTypes.INTEGER,
+      autoIncrement: true,
+      primaryKey: true,
+    },
+    name: {
```
kamilsmuga: 👍

db/model/room.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Rooms are gathering point for incident response. This table
+ * will hold the basic details for each of these rooms.
+ */
+
+const constants = require('../constants');
+
+const assoc = {};
+
+module.exports = function user(seq, dataTypes) {
```
kamilsmuga: change function name to "room"

db/model/room.js
```diff
@@ -0,0 +1,62 @@
...
+      validate: {
+        is: constants.nameRegex,
+      },
+      comment: 'Create a named room ',
+    },
+    active: {
+      type: dataTypes.BOOLEAN,
+      defaultValue: false,
+      comment: 'Determines if room is still active',
+    },
```
kamilsmuga: extra comma

db/model/roomType.js
```diff
@@ -0,0 +1,82 @@
...
+ * and an action object that has the name of the action and parameters values
+ * of the action needed to be taken when rule is valid.
+ *
+ */
+
+const constants = require('../constants');
+const u = require('../helpers/roomTypeUtils');
+const assoc = {};
+
+module.exports = function user(seq, dataTypes) {
```
kamilsmuga: change the function name from user to "roomType"

db/model/roomType.js
```diff
@@ -0,0 +1,82 @@
...
+      unique: true,
+      validate: {
+        is: constants.nameRegex,
+      },
+      comment: 'Create a named room type',
+    },
+    active: {
+      type: dataTypes.BOOLEAN,
+      defaultValue: false,
+      comment: 'Determines if room type is still active',
```
kamilsmuga: what is the purpose of this flag?
If to enable or to disable room types available in the menu - maybe it should be called isEnabled?

db/model/room.js
```diff
@@ -0,0 +1,62 @@
...
+    name: {
+      type: dataTypes.STRING,
+      allowNull: false,
+      unique: true,
+      validate: {
+        is: constants.nameRegex,
+      },
+      comment: 'Create a named room ',
+    },
+    active: {
```
kamilsmuga: how about "isActive" for better readability as it's a boolean?

db/model/roomType.js
```diff
@@ -0,0 +1,82 @@
...
+      getRoomTypeAssociations() {
+        return assoc;
+      },
+
+      postImport(models) {
+        assoc.writers = RoomType.belongsToMany(models.User, {
+          as: 'writers',
+          through: 'RoomTypeWriters',
+          foreignKey: 'roomTypeId',
+        });
```
kamilsmuga: should the relationship RoomType.hasMany(Rooms) be added?

db/model/roomType.js
```diff
@@ -0,0 +1,82 @@
...
+      },
+      comment: 'Key/Value pairs for user specific settings',
+    },
+    rules: {
+      type: dataTypes.ARRAY(dataTypes.JSONB),
+      allowNull: true,
+      validate: {
+        contains: u.validateRulesArray,
+      },
+      comment: 'Logic and resulting actions for rooms',
```
kamilsmuga: will there be a way to enable debugging and/or logging to see which rules has passed/failed/kicked off?

kamilsmuga:
+ build is failing on styling issues

db/model/roomType.js
tausifmuzaffar: This will be done in the BotActions model

db/model/roomType.js
tausifmuzaffar: I dont think so? What are you expecting from the table?

db/model/room.js
tausifmuzaffar: where am i missing or adding a comma?

db/model/roomType.js
tausifmuzaffar: an active room is a room that currently has a incident. sometimes you might wanna go back to a room even when its not active. IsEnabled sounds more like a flag while this is more of a state

db/model/room.js
kamilsmuga: Line 40 contains an extra comma. id, name, actions are all objects and there's no need to put comma after the last one (it will work, though)

tausifmuzaffar: I believe its the refocus style guide to have it on every object even the last one

db/model/roomType.js
kamilsmuga: I totally agree with the active flag on the Room model. But this is a Room Type model. Can you explain more how it would be used?

tausifmuzaffar: OH, i didnt room type, just room. Yeah I think it makes to call it isEnable here

db/model/roomType.js
kamilsmuga: As far as I understand, adding this relationship would enable us to do a query like: RoomType.getRooms() that returns all Rooms of a type.
As it stands now, only Room.getRoomType() would be possible. 

(reference https://stackoverflow.com/questions/20290815/belongsto-vs-hasmany-in-sequelize-js)

db/model/roomType.js
kamilsmuga: What would be the use case to have "isEnabled" flag for the Room Type model?

db/model/room.js
kamilsmuga: okay, I can see this being present on all the other models. Let's disregard then.

db/model/roomType.js
tausifmuzaffar: While a team is creating/updating their room type they wouldnt want someone to use it?

kamilsmuga: Right, I thought about a similar scenario. There will be a case where people won't want their configuration (room type) to be visible

db/model/roomType.js
kamilsmuga: Can you update the model diagram to show what BotActions are?

tausifmuzaffar:
@iamigo have you gotten a chance to look at this PR?

db/helpers/roomTypeUtils.js
```diff
@@ -0,0 +1,153 @@
...
+function validateRulesArray(arr) {
+  if (Array.isArray(arr)) {
+    for (let i = 0; i < arr.length; i++) {
+      if ((arr[i].hasOwnProperty('rule')) &&
+        (arr[i].hasOwnProperty('action'))) {
+        validateRules(arr[i].rule);
+        validateActions(arr[i].action);
+      } else {
+        throw new ValidationError({
+          message: 'Object missing a rules or actions attribute',
```
iamigo: plurals should be singular?

db/model/roomType.js
```diff
@@ -0,0 +1,85 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * db/model/roomType.js
+ *
+ * Room Types are specialized configurations for rooms. They hold intial
```
iamigo: typo: initial

iamigo:
I see a ton of validation rules but not a lot of tests... are you sure you're covering the validation?

tausifmuzaffar:
@iamigo there are tests to cover each fields validations. the utils classes for validation have nearly 100% code coverage also. So I feel pretty good about it.

PR #400, opened 2017-06-12
On socket.io connection…

iamigo:
- check for valid user in session
- confirm ip address is whitelisted


shriramshankar:
@iamigo added the changes we discussed.

PR #403, opened 2017-06-13
perspectives should get real time events again

annyhe:


annyhe:
This is on the master branch, after the ?beta=true param on /perspectives is merged.
Here is a screenshot of the events working locally
screen shot 2017-06-13 at 2 31 44 pm

PR #404, opened 2017-06-14
Add get default perspective test

annyhe:
- on GET default perspective: if the perspective exists, redirect to that perspective using the key value in the response.
- add tests to show what happens on GET default perspetive

### DIFF
--- a/view/perspective/utils.js
+++ b/view/perspective/utils.js
@@ -353,8 +353,8 @@ function getValuesObject(accumulatorObject) {
      */
      if (!named && returnedPerspective) {
 
-      // the key field has the name of the default perspective
-      redirectToUrl('/perspectives/' + returnedPerspective.key);
+      // the value field has the name of the default perspective
+      redirectToUrl('/perspectives/' + returnedPerspective.value);
      }
 
      const promisesArr = [
--- a/view/perspective/utils.js
+++ b/view/perspective/utils.js
@@ -337,7 +337,6 @@ function getValuesObject(accumulatorObject) {
 
     // use ternary as if pespective does not exist, responses[1] is undefined
     const returnedPerspective = responses[1] ? responses[1].body : null;
-    console.log('in responses', named, returnedPerspective)
 
     // assign perspective-related values to the accumulator object
     valuesObj.persNames = valuesObj.perspectives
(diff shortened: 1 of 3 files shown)


annyhe:
DO NOT merge. the problem is window.location.href = '...' does not redirect until the end of the function, so if there is a default perspective, the url will be routed to the FIRST perspective, instead of the default.

annyhe:
This PR shows the default perspective is loaded correctly, if it exists. You can test it out at https://refocus-devbox.herokuapp.com/perspectives/

annyhe:
This is what the /perspectives page looks like, with NO perspectives
screen shot 2017-06-14 at 8 41 49 am

annyhe:
Updated view when no perspectives exist
screen shot 2017-06-14 at 8 47 22 am

view/perspective/utils.js
```diff
@@ -347,49 +344,64 @@ function getValuesObject(accumulatorObject) {
...
+     * GET default perspective: does NOT exist: perspective is the first
+     *  perspective in the perspectives array.
+     * If no perspectives exist, valuesObj.perspective = null
+     */
+     if (!named) {
+      if (returnedPerspective) {
+
+        // the value field has the name of the default perspective.
+        // need the return statement to skip executing the rest of code
+        return redirectToUrl('/perspectives/' + returnedPerspective.value);
```
pallavi2209: Always a good practice to check for attribute before using it.

view/perspective/utils.js
```diff
@@ -347,49 +344,64 @@ function getValuesObject(accumulatorObject) {
...
+      }
+
+      // default perspective does NOT exist.
+      // GET the first perspective by alphabetical order.
+      // Check to see there are perspectives
+      if (valuesObj.perspectives.length) {
+
+        // redirect to the first perspective. The rest of the code
+        // won't be executed.
+        return redirectToUrl('/perspectives/' + valuesObj.perspectives[0].name);
```
pallavi2209: Check for valuesObj.perspectives[0].name. Always a good practice to check for attribute before using it.

annyhe: I put this in another story.

view/perspective/utils.js
annyhe: I put this in another story.

PR #406, opened 2017-06-14
Bot API Routes

tausifmuzaffar:
GET /bots, /bots/{key}
PATCH /bots/{key}
POST /bots 
DELETE /bots/{key}


tausifmuzaffar:
@iamigo @pallavi2209 have yall had a chance to look at this PR?

api/v1/controllers/bots.js
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
iamigo: 2017

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
           $ref: "#/responses/genericError"
 
 # =============================================================================
+  /bots:
+    x-swagger-router-controller: bots
+    get:
+      summary: Retrieve all bots
```
iamigo: "summary" attribute is ok as sentence fragment but "description" attribute should be written as full sentence (or paragraph).

api/v1/controllers/bots.js
iamigo: What about endpoints for adding/removing/listing who is authorized to update or delete a Bot?

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
...
+          $ref: "#/responses/401"
+        404:
+          $ref: "#/responses/404"
+        default:
+          $ref: "#/responses/genericError"
+
+    get:
+      summary: Retrieve a bot
+      tags: [ bots ]
+      operationId: getBots
```
iamigo: This can only return ONE bot, so better name would be "getBot" rather than "getBots"

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
...
+        -
+          name: key
+          in: path
+          description: >-
+            The id or name of the bot to retrieve
+          required: true
+          type: string
+        -
+          name: name
+          description: Get bots by name
```
iamigo: Get bot by name.

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
...
+          in: query
+          required: false
+          type: string
+        -
+          name: active
+          description: Get active bots
```
iamigo: what does this actually mean?

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
...
+          $ref: "#/responses/404"
+        default:
+          $ref: "#/responses/genericError"
+
+    patch:
+      security:
+        - jwt: []
+      summary: Update specified fields in bot
+      tags: [ bots ]
+      description: Update bot feilds
```
iamigo: typo: fields

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
...
+      operationId: patchBots
```
iamigo: singular

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
...
+            The id or name of the bot to retrieve
+          required: true
+          type: string
+        -
+          name: queryBody
+          in: body
+          required: true
+          schema:
+            type: object
+            description: Create bot properties
```
iamigo: This is a patch, so is "Create" really the correct wording here?

api/v1/helpers/nouns/bots.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
pallavi2209: 2017

api/v1/swagger.yaml
```diff
@@ -6160,6 +6160,239 @@ paths:
           $ref: "#/responses/genericError"
 
 # =============================================================================
+  /bots:
+    x-swagger-router-controller: bots
+    get:
```
pallavi2209: So, you don't want get endpoint to be secured with token?

tests/api/v1/bots/delete.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/api/v1/bots/delete.js
```
pallavi2209: I see that you are not using token for any tests and most of the CRUD tests will fail if we enforce token in test environment. It would be great if you can include tokens in test so that we don't have to change them in future (when we plan to enforce token by default). Does that makes sense?

PR #408, opened 2017-06-14
fix the redirect error message

annyhe:


view/perspective/utils.js
```diff
@@ -386,25 +386,35 @@ function getValuesObject(accumulatorObject) {
...
+          'perspective from the dropdown.');
+      }
     }
 
     return Promise.all(promisesArr);
   })
   .then((responses) => {
+
+    // on redirect, responses is a string
+    // otherwise it is an array of API responses
```
annyhe: the problem was the returned redirect result, a string, came into the  last .then block as the variable named 'responses'. So the filter array failed on the string.

annyhe:
deployed on devbox https://refocus-devbox.herokuapp.com/perspectives/SR_Grid

PR #412, opened 2017-06-16
instrument realtime event pub, sub and socket emit

iamigo:


iamigo:
I don't think I'd want to turn this flag on in prod for a very long time, but I think it could be informative to get some details about our pub/sub usage which could help us figure out the redis subscribe buffer problem.

(I have this turned on in devbox now if you want to see it in action.)

iamigo:
@shriramshankar @pallavi2209 please review this

working on the redis pub/sub issue, heroku said we should focus on "improving the performance of your consumer mostly, making sure messages are read as fast as possible"

if we can get this instrumentation turned on in prod, we will be able to see delays between pub and sub ... it's going to add a lot of logging, so I wouldn't want it turned on for very long in prod, but i think it might be useful

PR #415, opened 2017-06-22
Collector database model

pallavi2209:


db/model/collector.js
```diff
@@ -0,0 +1,89 @@
...
+      {
+        name: 'CollectorUniqueLowercaseName',
+        unique: true,
+        fields: [
+          seq.fn('lower', seq.col('name')),
+        ],
+      },
+    ],
+  });
+  return Collector;
```
iamigo: needs writers, and assign creator as writer on create

pallavi2209:
@iamigo @shriramshankar updated the PR as per comments and discussion.

db/model/collector.js
```diff
@@ -0,0 +1,118 @@
...
+      validate: { isEmail: true },
+    },
+    helpUrl: {
+      type: dataTypes.STRING(constants.fieldlen.url),
+      validate: { isUrl: true },
+    },
+    registered: {
+      type: dataTypes.BOOLEAN,
+      allowNull: false,
+      defaultValue: true,
```
shriramshankar: If this is like isPublished, should the default be false instead?

db/model/collector.js
```diff
@@ -0,0 +1,118 @@
...
+      afterCreate(inst /* , opts*/) {
+        // Add createdBy user to Collector writers.
+        if (inst.createdBy) {
+          return new seq.Promise((resolve, reject) =>
+            inst.addWriter(inst.createdBy)
+            .then(() => resolve(inst))
+            .catch((err) => reject(err))
+          );
+        }
+
```
shriramshankar: need a beforeDelete hook to set isDeleted.

db/model/collector.js
```diff
@@ -0,0 +1,118 @@
...
+        return inst;
+      }, // hooks.beforeCreate
+    }, // hooks
+    indexes: [
+      {
+        name: 'CollectorUniqueLowercaseName',
+        unique: true,
```
shriramshankar: isDeleted should also be a part of the unique index.

db/model/collector.js
pallavi2209: Yeah, I thought so too, but I think this field corresponds to the 'collector is registered with Refocus' and if someone is creating a collector, then they want it registered by default. @iamigo What do you think?

iamigo: This record will only ever be created by a collector registering itself, so registered should default to true.

PR #417, opened 2017-06-24
fix docker-compose and use alpine linux

esemeniuc:
Features:
 - creates focusdb so database can be initialized (previously missing)
 - alpine cuts download size down to 1/5th of the old size
 - uses sem versioning to keep up with security patches
 - eliminates the finicky sleep env variable so node always starts after postgres is ready


iamigo:
Thx for the contribution, @cwaffles! @dbabenko, does this look ok for you?

dbabenko:
@iamigo the changes look great to me. I'm traveling right now outside of the country and cannot test the changes. Will be able to try them out closer to the end of the next week.

Cwaffles:
Hi all,

could you please let me know why I was mentioned in this? 

Kind regards


Von meinem iPhone gesendet

cassiodias:
@iamigo can we merge these changes, looks like it solves the current problem using Docker.

iamigo:
approved but still requires CLA

esemeniuc:
CLA signed. cwaffles was my old username (unrelated to the current user)

esemeniuc:
Going to make a new pull request since the system is bugged

PR #419, opened 2017-06-27
publish sample test

annyhe:
- if cache is enabled, in publish sample, get aspect and subject from cache
- previously was fetching  aspect and subject from db
- primary aim of this PR is to see if there's performance improvement: with redis cache on,
the web dyno should be able to take more samples.

How to test locally:
tab 1: refocus:  `ENABLE_REDIS_SAMPLE_STORE=true node .`
tab 2: refocus-client-example: `node samples.js` with this as the last line in samples.js file:
Repeat(upsertSamples).every(2, 'ms').for(15, 'seconds').start.in(2, 'sec');
This does not err. While the same setup with master branch errs out in the tab 2
with errors:
{ RequestError: Error: getaddrinfo ENOTFOUND localhost localhost:3000
...
   { Error: socket hang up
...

### DIFF
--- a/realtime/redisPublisher.js
+++ b/realtime/redisPublisher.js
@@ -11,6 +11,8 @@
  */
 'use strict'; // eslint-disable-line strict
 
+const redisStore = require('../cache/sampleStore');
+const redisClient = require('../cache/redisCache').client.sampleStore;
 const pub = require('../cache/redisCache').client.pub;
 const channelName = require('../config').redis.channelName;
 const sampleEvent = require('./constants').events.sample;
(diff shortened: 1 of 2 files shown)


annyhe:
While there is a minor improvement locally, testing on heroku refocus-devbox does not show significant improvements. Both master branch and this feature branch result in errors 
```
2017-06-27T18:40:50.941935+00:00 heroku[router]: at=info method=POST path="/v1/samples/upsert/bulk" host=refocus-devbox.herokuapp.com request_id=0fe716cc-faec-4ded-9d19-4bf9f3ef3d62 fwd="204.14.239.53" dyno=web.1 connect=1ms service=8421ms status=200 bytes=332 protocol=https
2017-06-27T18:40:51.065487+00:00 heroku[router]: at=error code=H11 desc="Backlog too deep" method=POST path="/v1/samples/upsert/bulk" host=refocus-devbox.herokuapp.com 
```

and 

```
[[B2017-06-27T18:45:53.668004+00:00 heroku[worker.1]: Process running mem=1184M(115.6%)
2017-06-27T18:45:53.668123+00:00 heroku[worker.1]: Error R14 (Memory quota exceeded)
``` after sustained 2ms or 1ms.

when the last line in refocus-client-example is `Repeat(upsertSamples).every(2, 'ms').for(15, 'seconds').start.in(2, 'sec');`

PR #420, opened 2017-06-30
amend response from populate samples

annyhe:
- for db response, remove null fields from the response
- for redis cached response, JSON.parse fields
- for both responses, remove aspect.writers from the response

### DIFF
--- a/realtime/redisPublisher.js
+++ b/realtime/redisPublisher.js
@@ -147,9 +147,9 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
     delete aspect.writers;
 
     // if field value is stringified, will parse the value.
-    sample = redisStore.removeNullsAndParseArrays(sample,
+    sample = redisStore.arrayStringsToJson(sample,
       constants.fieldsToStringify.sample);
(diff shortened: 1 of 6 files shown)


realtime/redisPublisher.js
```diff
@@ -142,7 +144,11 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
     const asp = responses[0];
     const sub = responses[1];
 
-    sample.aspect = asp.get ? asp.get() : asp;
+    let aspect = asp.get ? asp.get() : asp;
+    aspect = redisStore.removeNulls(aspect,
+      constants.fieldsToStringify.aspect);
+    delete aspect.writers;
+    sample = cleanAddAspectToSample(sample, aspect);
```
shriramshankar: What happens when we try to clean aspects and subjects that are from db ?

shriramshankar: Also, samples from DB already have aspects. Probably will have to work around the samples from DB

annyhe:
Updated with changes. Please review @pallavi2209 @shriramshankar

PR #421, opened 2017-07-03
move job cleanup to clock process

jgraff2:


config.js
```diff
@@ -72,10 +72,10 @@ const DEFAULT_JOB_QUEUE_TTL_SECONDS_ASYNC = 3600;
 const DEFAULT_JOB_QUEUE_TTL_SECONDS_SYNC = 25;
 
 // default set to 30 minutes
-const DEFAULT_JOB_REMOVAL_DELAY_SECONDS = 1800;
+const DEFAULT_JOB_REMOVAL_INTERVAL_SECONDS = 1800;
 
-const JOB_REMOVAL_DELAY_SECONDS = pe.KUE_JOBS_REMOVAL_DELAY ||
-  DEFAULT_JOB_REMOVAL_DELAY_SECONDS;
+const JOB_REMOVAL_INTERVAL_SECONDS = pe.KUE_JOBS_REMOVAL_INTERVAL ||
```
iamigo: since you're changing the name of the env var, you need to make sure that prod and sandbox are configured appropriately... you can add the new one anytime (before the release) but also need to make sure to remove the old one after the release

clock/scheduledJobs/jobCleanup.js
```diff
@@ -0,0 +1,55 @@
...
+        if (featureToggles.isFeatureEnabled('instrumentKue')) {
+          console.log(`[KJI] Removed completed job ${job.id}`);
+        }
+      })
+      .catch((err) => {
+        console.log(`Error removing job ${job.id}`, err);
+        return Promise.reject(err);
+      })
+    ))
+  )
```
iamigo: nice

config.js
jgraff2: It looks like this variable isn't set in prod or sandbox so there's no action needed.
But this also made me realize there was a variable that was changed in the worker ttl change that is set in prod. (TTL_KUE_JOBS was changed to TTL_KUE_JOBS_ASYNC)

I've updated it in prod and sandbox.

clock/scheduledJobs/jobCleanup.js
shriramshankar: @jgraff2 Can you confirm that the "q:jobs:complete" set does not have anything in it after rangeByStateAsync is run successfully?

jgraff2: @shriramshankar That's what jobQueue.completeCount() checks in the tests.

PR #425, opened 2017-07-06
db model for generator

shriramshankar:


db/model/generator.js
```diff
@@ -0,0 +1,183 @@
...
+      },
+
+      postImport(models) {
+        assoc.createdBy = Generator.belongsTo(models.User, {
+          foreignKey: 'createdBy',
+        });
+
+        assoc.collectors = Generator.belongsToMany(models.Collector, {
+          as: 'collectors',
+          through: 'GeneratorCollectors',
```
shriramshankar: This is the only name I could come up. Any other suggestions?

shriramshankar:
@annyhe your api changes and the fields in this model should match.

db/model/generator.js
```diff
@@ -0,0 +1,183 @@
...
+  properties: {
+    name: {
+      description: 'GeneratorTemplate name associated with this generator',
+      type: 'string',
+      maxLength: constants.fieldlen.normalName,
+      pattern: constants.nameRegex,
+      required: true,
+    },
+    version: {
+      description: 'Version of the generatorTemplate associated with this ' +
```
iamigo: ```description: 'Generator template version or version range',```

db/model/generator.js
```diff
@@ -0,0 +1,183 @@
...
+        'generator',
+      type: 'string',
+      pattern: constants.versionRegex,
```
iamigo: also accepts comparators, hyphen ranges, x-ranges, tilde ranges and caret ranges. See https://docs.npmjs.com/misc/semver for version grammar.

PR #426, opened 2017-07-10
ensure the GET /samples?status=... is case insensitive

annyhe:


api/v1/helpers/verbs/findUtils.js
```diff
@@ -114,7 +126,11 @@ function toSequelizeWhere(filter, props) {
        */
       if (Array.isArray(props.fieldsWithEnum) &&
         props.fieldsWithEnum.indexOf(key) > -ONE) {
-        const enumArr = filter[key];
+
+        // if specified in props, convert the array in query to camelcase.
+        const enumArr = (props.fieldsToCamelCase &&
+          props.fieldsToCamelCase.indexOf(key) > -ONE) ?
+          convertArrayElementsToCamelCase(filter[key]) : filter[key];
```
shriramshankar: this is a ternary operator correct? What happens when 

props.fieldsToCamelCase && props.fieldsToCamelCase.indexOf(key) > -ONE is false ?

shriramshankar: I think if the user just calls samples?status= without the actual status, this will throw an error correct? The same will happen when the user gives an invalid status in the query param. Something like samples?status=A.

annyhe: yes that the current status on master is the application throws an error with samples?status=A. Will create another story to give a more useful error message

PR #429, opened 2017-07-11
prep for implementing collectors api

iamigo:
- validation error if try to change status from Stopped to Paused
- add /api/v1/helpers/nouns/collectors
- fix swagger


tests/db/model/collector/update.js
```diff
@@ -41,11 +42,69 @@ describe('tests/db/model/collector/update.js >', () => {
...
+  });
+
+  it('Update status [Stopped --> Paused] invalid', (done) => {
+    expect(collectorDb.status).to.be.equal('Stopped');
+    collectorDb.update({ status: 'Paused' })
+    .then((obj) => {
+      done('Expecting error');
+    })
+    .catch((err) => {
+      console.log(err);
```
pallavi2209: Is this intentional console log? Or do you intend to check for error?

tests/db/model/collector/update.js
```diff
@@ -41,11 +42,69 @@ describe('tests/db/model/collector/update.js >', () => {
...
+    collectorDb.update({ status: 'Running' })
+    .then((obj) => {
+      expect(obj.status).to.be.equal('Running');
+      done();
+    })
+    .catch(done);
+  });
+
+  it('Update status [Stopped --> Paused] invalid', (done) => {
+    expect(collectorDb.status).to.be.equal('Stopped');
```
annyhe: do we want to add a use case for invalid: pause -> stopped?

tests/db/model/collector/update.js
iamigo: updated, thx

tests/db/model/collector/update.js
annyhe: what about a test for paused -> resume is ok?
test: stop -> resume is invalid

iamigo: I thought I would but I can see a use case for wanting to just stop a paused job.

annyhe: so stop -> paused is valid?

PR #430, opened 2017-07-12
moved publisher for sample store to the cache layer

shriramshankar:


api/v1/controllers/samples.js
```diff
@@ -302,7 +305,10 @@ module.exports = {
         .then((samples) => {
           samples.forEach((sample) => {
             if (!sample.isFailed) {
-              publisher.publishSample(sample, subHelper.model);
+              if (!featureToggles
+                .isFeatureEnabled(sampleStoreConstants.featureName)) {
+                publisher.publishSample(sample, subHelper.model);
+              }
             }
```
annyhe: what is the reason for doing it this way?
also, are we going to move the publishSample in delRlinksPromise to cache too?

shriramshankar: We are doing it to see if moving the publishSample to cache can introduce some kind of a throttle on the realtime events.

PR #439, opened 2017-07-13
Bot blob

tausifmuzaffar:
Add UI blob to DB

### DIFF
--- a/db/model/bot.js
+++ b/db/model/bot.js
@@ -50,6 +50,11 @@ module.exports = function bot(seq, dataTypes) {
       validate: { isUrl: true },
       comment: 'The URL to load bot',
     },
+    ui: {
+      type: dataTypes.BLOB,
+      allowNull: true,
+      comment: 'The packaged UI of the bot',
+    },
     active: {
       type: dataTypes.BOOLEAN,
       defaultValue: false,
--- a/tests/db/model/bot/update.js
+++ b/tests/db/model/bot/update.js
@@ -15,6 +15,10 @@ const expect = require('chai').expect;
 const tu = require('../../../testUtils');
 const u = require('./utils');
 const Bot = tu.db.Bot;
+const fs = require('fs');
+const path = require('path');
+const mt = path.join(__dirname, './uiBlob2');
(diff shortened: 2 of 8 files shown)


kamilsmuga:
@tausifmuzaffar do you want to expose it via API in a separate PR?

db/model/bot.js
```diff
@@ -50,6 +50,11 @@ module.exports = function bot(seq, dataTypes) {
       validate: { isUrl: true },
       comment: 'The URL to load bot',
     },
+    ui: {
+      type: dataTypes.BLOB,
```
kamilsmuga: Why binary data type?

tausifmuzaffar: What you mean? This is the same pattern as lens.

tausifmuzaffar:
@kamilsmuga API will be a seperate PR, but under the same user story W-4138685

db/model/bot.js
kamilsmuga: Clarified on standup. Looks good

tests/db/model/bot/utils.js
```diff
@@ -10,16 +10,20 @@
...
 const testStartTime = new Date();
 const n = `${tu.namePrefix}TestBot`;
 const n2 = n+'NonActive';
+const mt = path.join(__dirname, './uiBlob');
+const uiBlob = fs.readFileSync(mt);
 
 const standard = {
   name: n,
   url: 'http://www.bar.com',
+  ui: uiBlob,
```
pallavi2209: An update db test for this field would be great, but other than that, I am fine merging this PR.

tests/db/model/bot/update.js
```diff
@@ -48,6 +52,17 @@ describe('db: bot: update: ', () => {
       .catch(done);
     });
 
+    it('ok, bot ui updated', (done) => {
+      Bot.findOne({ where: { name: u.name } })
+      .then((o) => o.update({ url: 'http://www.test.com' }))
```
pallavi2209: ui instead of url

PR #442, opened 2017-07-17
add Generator to swagger

annyhe:


api/v1/controllers/generators.js
```diff
@@ -0,0 +1,137 @@
...
+   * authorized list of writers stored in the cache for this generator.
+   *
+   * @param {IncomingMessage} req - The request object
+   * @param {ServerResponse} res - The response object
+   * @param {Function} next - The next middleware function in the stack
+   */
+  deleteGeneratorWriter(req, res, next) {
+    res.send({ status: 204, text: 'Delete generator writer' });
+  },
+}; // exports
```
shriramshankar: Why are you "stubbing" all the responses.? We already have the generator model in the database.

iamigo: Because the controller changes require adding tons of tests, I separated out the swagger changes from the controller changes, just to keep things moving forward in small manageable chunks.

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+              - subjects
+              - subjectQuery
+              - createdAt
+              - description
+              - helpEmail
+              - helpUrl
+              - id
+              - name
+              - updatedAt
+              - - isActive
```
dexwiz: Should this be `- -isActive`. Otherwise the enum is `- isActive`.

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+        a list of fields to include in the response. If the Refocus configuration parameter `useAccessToken` is set to `true`, you must include an `Authorization` request header with your [JSON Web Token](https://tools.ietf.org/html/rfc7519) (JWT) as the value. You can get a token using `POST /v1/register` or `POST /v1/tokens`.
+      externalDocs:
+        description: >-
+          GET /generators
+        url: http://focus.salesforce.com/docs/api/generators
+      operationId: findGenerators
+      parameters:
+        -
+          $ref: "#/parameters/GeneratorFieldsParam"
+        - name: sort
```
dexwiz: Others in the file follow the pattern:
```yaml
-
  name: paramName
```

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+          $ref: "#/parameters/limitParam"
+        -
+          $ref: "#/parameters/offsetParam"
+        -
+          name: description
+          in: query
+          type: string
+          maxLength: 4096
+          description: >
+            A description of the generator.
```
dexwiz: Indicate that this a filter, like for /aspects. Similar comment for all further params on this endpoint.

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+          in: query
+          type: string
+          description: >
+            Id of the User who created this Generator.
+        -
+          name: id
+          in: query
+          type: string
+          description: >
+            The collector id.
```
dexwiz: Any reason for this to exists when `GET /generators/{key}` exists?

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+        400:
+          $ref: "#/responses/400"
+        default:
+          $ref: "#/responses/genericError"
+    post:
+      security:
+        - jwt: []
+      summary: Create a new generator.
+      tags: [ generators ]
+      description: >-
```
dexwiz: Inconsistent multiline indicator

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+          name: key
+          in: path
+          description: The id or key of the generator to retrieve.
+          required: true
+          type: string
+      responses:
+        200:
+          description: Success, returns the specified generator.
+          schema:
+            $ref: "#/definitions/GlobalConfigResponse"
```
dexwiz: Should this be `GeneratorResponse`

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+                description: Information about the connection.
+              context:
+                type: object
+                description: >
+                  Variables accessible inside the transform function.
+      responses:
+        200:
+          description: Success, returns the updated generator.
+          schema:
+            $ref: "#/definitions/GlobalConfigResponse"
```
dexwiz: Should return `GeneratorResponse`.

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+          $ref: "#/responses/404"
+        default:
+          $ref: "#/responses/genericError"
+    put:
+      security:
+        - jwt: []
+      summary: Update the specified generator
+      tags: [ generators ]
+      description: >-
+        Update the specified generator. Requires user to have write permision. If a field is not included in the query body, that field will not be updated. If the Refocus configuration parameter `useAccessToken` is set to `true`, you must include an `Authorization` request header with your [JSON Web Token](https://tools.ietf.org/html/rfc7519) (JWT) as the value. You can get a token using `POST /v1/register` or `POST /v1/tokens`.
```
dexwiz: `If a field is not included in the query body, that field will not be updated.`
Is this accurate for `PUT`?

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+          name: queryBody
+          description: Request body.
+          in: body
+          required: true
+          schema:
+            type: array
+            items:
+              type: string
+            description: >
+              User name or id
```
dexwiz: What happens if User A's ID is the same as User B's Name? Which gets priority? Wouldn't this open the API up to spoofing IDs?

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+      responses:
+        201:
+          description: >-
+            One or more of the valid users were added to the list of authorized writers.
+          schema:
+            type: array
+            items:
+              $ref: "#/definitions/GeneratorWriterResponse"
```
dexwiz: Should this be `UsersResponse`?

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+        403:
+          $ref: "#/responses/403"
+        404:
+            $ref: "#/responses/404"
+        default:
+          $ref: "#/responses/genericError"
+
+  # ---------------------------------------------------------------------------
+
+  /generators/{key}/writers/{userNameOrId}:
```
dexwiz: Again, if can I spoof a user by setting one user's name to another user's id?

api/v1/swagger.yaml
```diff
@@ -1790,6 +1790,586 @@ paths:
...
+          required: true
+          type: string
+      responses:
+        200:
+          description: >-
+            Success, returns the users having write permission to the generator.
+          schema:
+            type: array
+            items:
+              $ref: "#/definitions/UsersResponse"
```
dexwiz: Should this be an array of a single item for `GET /generators/{key}/writers/{userNameOrId}`? Generally endpoints with an id at the end of the path return a single item or 404.

api/v1/swagger.yaml
```diff
@@ -8144,6 +8724,32 @@ definitions:
...
+        type: string
+        readOnly: true
+        description: >
+          Id of a user having write permission to the generator
+      createdAt:
+        readOnly: true
+        type: string
+        description: >
+          Timestamp at which the the write permission was created
+      updatedAt:
```
dexwiz: Can these by modified? Or is it only set/unset?

annyhe: They are set when the generator is modified.

api/v1/swagger.yaml
annyhe: yeah this is the single user object. The GET /generators/{key}/writers returns an array of these.

api/v1/swagger.yaml
annyhe: updated the putGenerator description

api/v1/swagger.yaml
annyhe: yep fixed. The changes should only have "description: >-" instead of "description: >"

api/v1/swagger.yaml
annyhe: you can POST a user with { name: [another's user's id ], ... } 
However on subsequent GET v1/users/[another's user's id] the another user is returned. So the spoof does not work

api/v1/swagger.yaml
annyhe: GET /generators/{key}/writers/{userNameOrId} should be a single item or 404. Updated the description and schema.

api/v1/swagger.yaml
dexwiz: This object is not the `Generator` or the `Writer`, it is the association between the objects, which only seem to be created or destroyed.

api/v1/swagger.yaml
dexwiz: So ID takes priority over name?

api/v1/swagger.yaml
```diff
@@ -1790,6 +1792,567 @@ paths:
...
   # ---------------------------------------------------------------------------
+
+  /generators:
+    get:
+      security:
+        - jwt: []
+      summary: Get a generator or all generators.
+      tags: [ generators ]
+      description: >-
+        Get a generator or all generators. Find a lens or lenses. You may query using field filters with asterisk
```
pallavi2209: "Find a lens or lenses" ??

api/v1/swagger.yaml
annyhe: this is a good point. Will rethink whether it is needed for all the writer responses.

PR #444, opened 2017-07-18
Adding settings to room and bot

BrandonMurnane:
Added settings to both models and added a hook to room to initialise room settings based on room type picked
Added tests to bot and room for new settings


tests/db/model/bot/utils.js
```diff
@@ -21,6 +21,12 @@ const standard = {
   name: n,
   url: 'http://www.bar.com',
   active: true,
+  settings: [
+    {
+      key: 'key1',
+      helpText: 'helpText1'
```
kamilsmuga: Can you add a test for: "This is some text that helps to understand what the settings should do"

db/helpers/botUtils.js
```diff
@@ -38,6 +38,13 @@ const parameterArraySchema = Joi.alternatives().try(
   )
 );
 
+const settingsArraySchema = Joi.array().items(
+    Joi.object().keys({
+      key: Joi.string().regex(/^[0-9a-z_-]+$/i).required(),
+      helpText: Joi.string().regex(/^[0-9a-z_-]+$/i).required(),
```
kamilsmuga: are whitespaces allowed for helpText?

tests/db/model/bot/utils.js
BrandonMurnane: i dont understand what you mean

db/model/room.js
```diff
@@ -33,6 +34,14 @@ module.exports = function room(seq, dataTypes) {
       },
       comment: 'Create a named room ',
     },
+    settings: {
+      type: dataTypes.ARRAY(dataTypes.JSONB),
+      allowNull: true,
+      validate: {
+        contains: u.validateSettingsArray,
+      },
+      comment: 'Key/Value pairs for user specific settings',
```
iamigo: curious about why you're storing key/value pairs as an array rather than as an object

BrandonMurnane: in this case i just copied it from the roomType as they are initialised from those settings and are then only changed on a room by room basis so it made most sense for it to be an array.

iamigo: Can you and the IMC team discuss this design for storing these key/value pairs--I'd like to see a uniform approach consistent across the different models where applicable, but I'm not sure an array is the right way to store it. I'd rather get it right the first time, because field changes require a migration.

BrandonMurnane: https://github.com/salesforce/refocus/blob/master/db/model/roomType.js#L49
We have already implemented as part of a past pr for room types so this would be keeping it consistent. If we are too change it on this pr we would then need a migration to change it for room types.

tausifmuzaffar: I think we decided on the array of objects because its an unordered list. As brandon mentioned this design has been implemented on room types already and the room is inheriting it.

pallavi2209:
Ok to merge once Ian's comments are taken care of.

db/model/room.js
iamigo: How many elements do you expect a typical *whatever* settings object to have? As an array, you will always have to iterate over the array to find one element, which would def. be slower for large objects. I would urge you to reconsider storing as array unless you are also going to *enforce* that it can only contain some small maximum number of elements.

tausifmuzaffar: @BrandonMurnane I think that raises a good point here. The room type stores the initial values that will only be accessed once when the room is created, but as a bot developer you may hit the room settings often.

So since we are inheriting them anyway we could do the transform from [{key: name, value:info}] to {key:info} for the ease of the bot developer

BrandonMurnane: i think if we are going to change how it is here we should be changing it to be uniform across everything which means doing a migration for room type

tausifmuzaffar: I agree, if you could do that for your PR that will be appricated

iamigo:
I can review this once tests are passing.

BrandonMurnane:
If you could look at the tests that would be great as random ones keep
failing but locally everything is passing fine

Sent from my iPhone

On 1 Aug 2017, at 22:32, Ian Goldstein <notifications@github.com> wrote:

I can review this once tests are passing.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/salesforce/refocus/pull/444#issuecomment-319503329>,
or mute
the thread

iamigo:
requireTrailingComma: Missing comma before closing curly brace at migrations/20170731100411-settings-array-to-object.js :
    40 |      },
    41 |    });
    42 |  }
----------^
    43 |};
    44 |

iamigo:
plus a bunch of bots errors

db/helpers/botUtils.js
```diff
@@ -38,6 +38,13 @@ const parameterArraySchema = Joi.alternatives().try(
   )
 );
 
+const settingsArraySchema = Joi.array().items(
+    Joi.object().keys({
+      key: Joi.string().regex(/^[0-9a-z_-]+$/i).required(),
```
iamigo: limit key length?

db/model/bot.js
```diff
@@ -68,6 +68,14 @@ module.exports = function bot(seq, dataTypes) {
       },
       comment: 'List of actions a Bot can take',
     },
+    settings: {
+      type: dataTypes.ARRAY(dataTypes.JSONB),
+      allowNull: true,
+      validate: {
+        contains: u.validateSettingsArray,
+      },
+      comment: 'Array[ {key: name of key, helpText: describe the value its looking for},]',
```
iamigo: "its" should be "it is" (or use "it's" but you'd have to escape the apostrophe)

iamigo: So this is still an array, not an object where the key is the key?

BrandonMurnane: this is staying an array of help texts

iamigo: ok

PR #445, opened 2017-07-19
Filter by tags with fields param

jgraff2:


api/v1/helpers/verbs/doFind.js
```diff
@@ -68,13 +68,23 @@ function doFindAndCountAll(reqResNext, props, opts) {
  */
 function doFindAll(reqResNext, props, opts) {
```
jgraff2: doFindAndCountAll (above) doesn't apply the advanced tag logic, such as filtering for multiple tags, so there are no changes required there. Is there a reason for this?

api/v1/helpers/verbs/doGetHierarchy.js
```diff
@@ -29,6 +29,32 @@ function doGetHierarchy(resultObj) {
   const params = resultObj.params;
   const depth = Number(params.depth.value);
   resultObj.dbStartTime = Date.now();
+
+  const fields = params.fields.value;
+  const filterFields = fields && fields.length;
+  const subjectTags = params.subjectTags.value;
+  const filterByTags = subjectTags && subjectTags.length;
```
jgraff2: The fields filter is only applied to the top-level object. Is this correct? If it was applied to the entire hierarchy we would also need to make sure the other 3 filter fields are included (aspectTags, aspect, status).

api/v1/helpers/verbs/doFind.js
iamigo: That was merely an oversight--all the same filtering should apply there too (OK as separate story if it would add too much to this story)

api/v1/helpers/verbs/doGetHierarchy.js
iamigo: yes, apply to top-level object object

(we could always add additional stuff like "sampleFields=status,previousStatus..." later)

api/v1/helpers/verbs/doFind.js
jgraff2: W-4163665 created.

shriramshankar:
Will review it by the end of the day.

PR #446, opened 2017-07-21
Swagger docs interactive

BrandonMurnane:
Changed swagger docs so it will use whatever port it is on rather than giving it one


BrandonMurnane:
@iamigo its the same as trust uses i originally set it up there. Is there an easy way to create review apps ?

BrandonMurnane:
also do you know why the bulk upset test breaks regularly @iamigo

api/v1/swagger.yaml
```diff
@@ -41,8 +42,8 @@ basePath: /v1
 
 # =============================================================================
 schemes:
-  - https
   - http
+  - https
```
pallavi2209: @BrandonMurnane curious to know why this change? Does this change the "try it out feature" to default to http instead of https? Would this have any side effects in terms of api working?

BrandonMurnane: It was a known issue for swagger 2.0 when i originally did this on trust. It means that it will default to http unless the page is served over https in which case it will serve over https. Shouldnt make any difference to how api working just for docs as schemes were set to http and https anyway to allow it to accept both.

iamigo:
@BrandonMurnane holding off on approval and merging until I can see it running in heroku

iamigo:
Brandon why is this PR still open after 25 days? please move it along or close it

PR #453, opened 2017-07-27
Fields added to Profile db to control 'rw' access to IMC Resources

kriscfoster:
botAccess, eventAccess, roomAccess, roomTypeAccess fields added to Profile model. 
updated default 'rw' access for admin access in .config (new fields)
Added tests for profile db create (insert), update

### DIFF
--- a/tests/db/model/profile/update.js
+++ b/tests/db/model/profile/update.js
@@ -1,0 +1,82 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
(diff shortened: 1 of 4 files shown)


tests/db/model/profile/update.js
```diff
@@ -0,0 +1,82 @@
...
+      })
+      .then((o) => o.update({ aspectAccess: 'r' }))
+      .catch((err) => {
+        expect(err.name).to.equal('AdminUpdateDeleteForbidden');
+        done();
+      })
+      .catch(done);
+    });
+  });
+});
```
shriramshankar: I am not sure about the reason but it good to have a new line at the end of the file. You can take care of it in the next PR.

PR #456, opened 2017-07-27
job cleanup changes

jgraff2:
1. delete jobs in batches of configurable size to avoid crashing the process by trying to delete too many jobs
2. only delete jobs older than x minutes so completed jobs are guaranteed to be accessible for a certain period
3. Delete the q:ids key along with the older jobs in 2.
4. move the actual work to the worker, just use the clock to schedule it (like sample timeout)


clock/index.js
```diff
@@ -56,4 +56,7 @@ if (featureToggles.isFeatureEnabled('enableQueueStatsActivityLogs')) {
 }
 
 // Clean up completed jobs
-setInterval(jobCleanup.execute, conf.JOB_REMOVAL_INTERVAL_SECONDS * 1000);
+setInterval(jobCleanup.enqueue, conf.JOB_REMOVAL_INTERVAL);
+
+// Reset the job id counter
+setInterval(jobCleanup.resetCounter, conf.JOB_COUNTER_RESET_INTERVAL);
```
jgraff2: I added this as a separate clock process that runs less frequently. Resetting the id counter and deleting jobs every time will cause new jobs to overwrite the completed jobs that were skipped.

config.js
```diff
@@ -71,11 +71,24 @@ const readReplicas = configUtil.getReadReplicas(pe, replicaConfigLabel);
...
+  DEFAULT_JOB_REMOVAL_INTERVAL_MINUTES);
 
-const JOB_REMOVAL_INTERVAL_SECONDS = pe.KUE_JOBS_REMOVAL_INTERVAL ||
-  DEFAULT_JOB_REMOVAL_INTERVAL_SECONDS;
+const DEFAULT_JOB_REMOVAL_DELAY_MINUTES = 30;
+const JOB_REMOVAL_DELAY = 60 * 1000 * (pe.KUE_JOBS_REMOVAL_DELAY_MINUTES ||
+  DEFAULT_JOB_REMOVAL_DELAY_MINUTES);
+
+const DEFAULT_JOB_REMOVAL_WINDOW = 1000;
+const JOB_REMOVAL_WINDOW = pe.KUE_JOBS_REMOVAL_WINDOW ||
```
jgraff2: Any thoughts on these four default values?

iamigo:
2 tests failing?

clock/scheduledJobs/jobCleanup.js
```diff
@@ -16,40 +16,110 @@ const Promise = require('bluebird');
...
 const jobSetup = require('../../jobQueue/setup');
 const jobQueue = jobSetup.jobQueue;
+const conf = require('../../config');
 const rangeByStateAsync = Promise.promisify(kue.Job.rangeByState);
 kue.Job.prototype.removeAsync = Promise.promisify(kue.Job.prototype.remove);
 
 /**
  * Execute the call to clean up jobs
  *
+ * @param {Number} window - the number of jobs to delete in each batch
```
iamigo: It's legal to do it but it's better *not* to use "window" as a variable name only because it's a global variable when running javascript in a browser context.
Why not just "batchSize" or something like that?

config.js
iamigo: those defaults seem fine to me

maybe change var name to use something like batch size instead of window?

config.js
```diff
@@ -71,11 +71,24 @@ const readReplicas = configUtil.getReadReplicas(pe, replicaConfigLabel);
...
+  DEFAULT_JOB_REMOVAL_WINDOW;
+
+// This must be set to several times the value of JOB_REMOVAL_INTERVAL or
+// completed jobs may be overwritten sooner that JOB_REMOVAL_DELAY
```
iamigo: What about making the delay config var a multiplier of the interval minutes rather than a number of minutes itself? And if user provides any multiplier less than 1.5 just set it to 1.5, or something like that?

clock/index.js
shriramshankar: Looks like resetCounter is called from the clock process. Any reason for doing it this way than making the worker do it.?

clock/scheduledJobs/jobCleanup.js
```diff
@@ -16,40 +16,110 @@ const Promise = require('bluebird');
...
+
 } // execute
 
+/**
+ * Send the job to the worker or execute directly
+ */
+function enqueue() {
+  if (featureToggles.isFeatureEnabled('enableWorkerProcess')) {
+    const jobWrapper = require('../../jobQueue/jobWrapper');
+    const jobType = require('../../jobQueue/setup').jobType;
```
shriramshankar: require('../../jobQueue/setup') has already been required at the top. You could just use that.

clock/index.js
jgraff2: It seems wasteful to set up a whole new job type and make the job wait in the queue for something so simple.

clock/scheduledJobs/jobCleanup.js
jgraff2: You're right, I copied this from the sampleTimeout job. I will change it.

clock/scheduledJobs/jobCleanup.js
iamigo: Sorry to be a pest, @jgraff2, but can you change these variable names, too, please?

jgraff2: You told me to ignore your comments last standup, I thought you meant this one. I'll change it.

PR #457, opened 2017-07-28
DELETE BotData API

tausifmuzaffar:
DELETE /botData/key


pallavi2209:
@tausifmuzaffar Going away for 2 weeks. I am replacing @shriramshankar as reviewer instead of myself for most of the PRs.

api/v1/swagger.yaml
```diff
@@ -7537,6 +7537,42 @@ paths:
...
+          description: >-
+            The id or name of the botData to delete.
+          required: true
+          type: string
+      responses:
+        200:
+          description: >-
+            Success, Deleted BotData
+          schema:
+            type: array
```
shriramshankar: The response should the deleted object instead of an array.

PR #458, opened 2017-07-28
POST BotData API

tausifmuzaffar:
POST /botData


api/v1/swagger.yaml
```diff
@@ -7537,6 +7537,61 @@ paths:
...
+                description: Value of the bot data
+              botId:
+                type: string
+                maxLength: 254
+                description: UUID for the coresponding bot
+              roomId:
+                type: integer
+                maxLength: 254
+                description: Number for the coresponding room
+            required:
```
shriramshankar: "corresponding" typo in line 7573 and 7569.

api/v1/swagger.yaml
```diff
@@ -7537,6 +7537,61 @@ paths:
...
+        -
+          name: queryBody
+          in: body
+          required: true
+          schema:
+            type: object
+            description: Create botData properties
+            properties:
+              name:
+                type: string
```
shriramshankar: Wondering if there are any restrictions on the on the characters that can be used here; like we have for other resources.

tausifmuzaffar: The restrictions are done on the model level, do we need it on the API level as well?

iamigo: Putting them in swagger means fail faster--swagger middleware sends back error instead of having the request wind its way through the middleware to our controller and through to the db.

tausifmuzaffar: Post i have added the swagger validation

shriramshankar:
Ok to merge once the tests are passing.

tausifmuzaffar:
@iamigo I have added the api validation

PR #459, opened 2017-07-28
Blob UI API and Tests

tausifmuzaffar:
/bots add the UI column to API


tausifmuzaffar:
@iamigo @shriramshankar "jobCleanup delay - staggered -  jobs: 20, delay: 159ms:" - test is failing

 Do you think something in my code is causing this?  Im not familiar with that code at all.

iamigo:
@jgraff2

jgraff2:
This is from my code that was merged earlier today.

I had some difficulty getting these to pass consistently since they rely on timing. There's a tradeoff between increasing the timeouts to make it pass more reliably and having the tests take too long. I thought I had figured it out, but I guess not.

I will increase the timeout for those tests to make it more reliable.

iamigo:
2 bots tests failing now

tausifmuzaffar:
There was a bad merge where the put got moved, an aspect test is failing now. is that a known issue?

PR #460, opened 2017-07-28
Profile API can control "rw" access fields to IMC Resources

kriscfoster:
-  Updated swagger.yaml
-  User story said to update API controller, don't think I actually need to do that here
-  Added get tests, updated post/put tests


kriscfoster:
Will add last IMC resources to this PR when they're ready (Bot Data, Bot Actions)

PR #461, opened 2017-07-28
Only Admin Profile Users can add/edit/delete other profiles

kriscfoster:
W-4175897
Will now require an admin access token to create/edit/delete a profile. Also added some tests for this.


kriscfoster:
@iamigo I can now confirm that all of those scenarios are tested for. I added a new routine called "test-api-profiles" which is invoked with REQUIRE_ACCESS_TOKEN=true.. Let me know whether this was the correct way to go about this.

iamigo:
@kriscfoster not another new one... pls. use the "test-enforced" script in package.json

kriscfoster:
@iamigo Didn't see that, thanks.. I have made those changes now

package.json
```diff
@@ -41,7 +41,7 @@
     "test-api": "mocha -R dot --recursive tests/api",
     "test-api-log": "ENABLE_API_ACTIVITY_LOGS=true mocha -R dot --recursive tests/api",
     "test-disablehttp": "REQUIRE_HTTPS=true mocha -R dot --recursive tests/disableHttp",
-    "test-enforced": "REQUIRE_ACCESS_TOKEN=true mocha -R dot --recursive tests/enforceToken",
+    "test-enforced": "REQUIRE_ACCESS_TOKEN=true mocha -R dot --recursive tests/enforceToken tests/api/v1/profiles",
```
iamigo: So adding this here, we're now executing all the tests under tests/api/v1/profiles _once_ with REQUIRE_ACCESS_TOKEN=true and again with REQUIRE_ACCESS_TOKEN=false? @kriscfoster

kriscfoster: That's correct, all tests under v1/profiles are being called once with REQUIRE_ACCESS_TOKEN = true and once with false @iamigo

iamigo:
I'm ready to approve this once tests are passing.

tests/api/v1/profiles/post.js
```diff
@@ -45,11 +55,36 @@ describe(`api: POST ${path}`, () => {
...
+      });
+    });
+
+    it('fail, no token provided', (done) => {
+      api.post(`${path}`)
+      .send(p0)
+      .expect(constants.httpStatus.FORBIDDEN)
+      .end((err, res) => {
+        if (err) {
+          done(err);
```
shriramshankar: You don't have to change it in this PR but we ideally should be adding a "return" in front of done(err). If you don't add it, all the errors will fail, saying that "done" was called twice.

kriscfoster:
Added returns before done(err)

In tests/api/v1/profiles/get I changed the expectations for admins to have 'rw' access for imc resources because this may cause confections with yesterdays migration which initialized all of the new IMC resource access's to 'r' as default.

PR #463, opened 2017-07-28
Generator api post with created by

shriramshankar:


api/v1/helpers/verbs/doPut.js
```diff
@@ -70,9 +70,10 @@ function doPut(req, res, next, props) {
...
             o.changed(key, false);
           } else {
-
-            // value may have changed. set changed to true to
-            // trigger checks in the model
+            /**
+             * value may have changed. set changed to true to
+             * trigger checks in the model
+             */
             o.changed(key, true);
```
iamigo: Is this really necessary here? (I thought we were doing this just for samples and subjects in order to make sure we generated the realtime events.)

shriramshankar: I did not change line 77, just corrected the formatting for the comments.  Need to check why line 77 was added.

shriramshankar: Looks like 77 was added on May 18th to fix the PUT issues from the admin.

shriramshankar:
Merging this PR now.

PR #467, opened 2017-08-01
Reject write Requests if don't have write access to model

kriscfoster:
@iamigo Just wanted to know whether it was the correct way to go about it? Also, it will break a lot of tests where requests were made with just an ordinary refocus user token. Should I also change these tests in this PR? Might be a good idea to split it up into multiple PR's  because of the size of it.


kriscfoster:
If REQUIRE_ACCESS_TOKEN = true, profile permissions will be used. eg If profile permits 'r' for subjectAccess, POST, PATCH, PUT, DELETE requests will be forbidden.

kriscfoster:
@iamigo Some tests failing because the token provided in the test request comes from a profile which does not have 'rw' access. Is the solution right now to change the tests or change the OOTB normal profile?

iamigo:
I think we talked about modifying the OOTB normal profile to stay backwards compatible.

kriscfoster:
@iamigo It's getting messy trying to do this with middleware with some requests. I might revert back to doing it inside controllers if that's okay with you.

PR #476, opened 2017-08-03
RoomType Migration Typo

tausifmuzaffar:
The migration referenced "RoomType" when the tables name is "RoomTypes"

### DIFF
--- a/migrations/20170731100411-settings-array-to-object.js
+++ b/migrations/20170731100411-settings-array-to-object.js
@@ -7,6 +7,7 @@
  */
 'use strict';
 const u = require('../db/helpers/roomTypeUtils');
+const TBL = 'RoomTypes';
 
 module.exports = {
   up: function (qi, Sequelize) {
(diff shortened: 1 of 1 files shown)


migrations/20170731100411-settings-array-to-object.js
```diff
@@ -32,7 +33,7 @@ module.exports = {
       return queryInterface.dropTable('users');
     */
 
-    return qi.changeColumn('RoomType', 'settings', {
+    return qi.changeColumn(TBL, 'settings', {
       type: Sequelize.ARRAY(Sequelize.JSONB),
       allowNull: true,
```
shriramshankar: Bit curious here to know the choice of JSONB vs JSON here.

tausifmuzaffar: We wanted to avoid transforms on responses, and allow json searches, but this is on the down, because thats what we originally did. On the up we switch to JSON

PR #480, opened 2017-08-05
only allow an admin user to create/modify/delete the ssoconfig entry

iamigo:


api/v1/controllers/ssoconfig.js
```diff
@@ -58,41 +59,51 @@ function getApiLinks(key, props, method) {
...
-        .then((destroyedObj) => {
-          resultObj.dbTime = new Date() - resultObj.reqStartTime;
-          u.logAPI(req, resultObj, destroyedObj);
-          res.status(httpStatus.OK)
-          .json(responsify(destroyedObj, helper, req.method));
-        });
+    authUtils.isAdmin(req)
+    .then((ok) => {
+      if (ok) {
+        const resultObj = { reqStartTime: new Date() };
```
annyhe: do we want to move the reqStartTime to after the isAdmin check?

annyhe: same for the other operations on ssoConfig

iamigo:
ok @annyhe i've moved those start times up

PR #486, opened 2017-08-08
Fix GET /collectors

iamigo:
- add swagger defn params for limit, offset, and filtering
- add test

### DIFF
--- a/tests/api/v1/collectors/get.js
+++ b/tests/api/v1/collectors/get.js
@@ -75,12 +75,18 @@ describe(`api: GET ${path}`, () => {
   });
 
   it('get with limit ok', (done) => {
-    api.get(path + '?limit=1')
-    .set('Authorization', token)
-    .expect(constants.httpStatus.OK)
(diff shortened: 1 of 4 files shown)


tests/api/v1/collectors/get.js
```diff
@@ -61,6 +61,26 @@ describe(`api: GET ${path}`, () => {
...
+  });
+
+  it('get with limit ok', (done) => {
+    api.get(path + '?limit=1')
+    .set('Authorization', token)
+    .expect(constants.httpStatus.OK)
+    .expect((res) => {
+      expect(res.body.length).to.equal(1);
+    })
+    .end(done);
```
annyhe: limit of 1 doesn't really show limit is working, as there is only one collector anyway. please add a test for limit=0. empty array should be returned.

iamigo: added test which inserts another collector then calls GET with limit=1

(limit=0 is the same as no limit)

PR #487, opened 2017-08-08
Handle Rejection and run tests with api token

shriramshankar:


iamigo:
tests failing?

shriramshankar:
@iamigo pushed the changes.

shriramshankar:
@iamigo All the fixed done and committed to this pr.

shriramshankar:
I also deployed this PR to sandbox and do not see the unhandled rejection error happening.

api/v1/controllers/tokens.js
```diff
@@ -52,6 +52,9 @@ module.exports = {
           } else {
             u.forbidden(next);
```
iamigo: put a `return` in front of _this_ one

jobQueue/jobWrapper.js
```diff
@@ -130,7 +130,7 @@ function logJobOnComplete(req, job) {
       .then((resObj) => {
         logObject.user = resObj.username;
         logObject.token = resObj.tokenname;
-      });
+      }).catch(() => {}); // no-op
```
iamigo: It looks like getTokenDetailsFromRequest throws if its call to getTokenDetailsFromTokenString throws, which happens if its call to jsonwebtoken module's jwt.verify throws, which happens if the token's signature and optional expiration, audience, or issuer are not valid.
Would there be any value in logging this in the catch here? Maybe shed some light on why we're getting these invalid tokens?

tests/api/v1/helpers/utils.js
```diff
@@ -16,7 +16,7 @@ const Subject = tu.db.Subject;
 const expect = require('chai').expect;
 const apiUtils = require('../../../../api/v1/helpers/verbs/utils.js');
 
-describe('util function tests', () => {
+describe('api/v1/helpers/utils.j function tests', () => {
```
iamigo: .js

(when the file name is complete, my console allows me to click it to go straight to that code in my editor)

tests/cache/models/samples/upsertBulk.js
```diff
@@ -342,7 +343,7 @@ describe('api::redisEnabled::POST::bulkUpsert ' + path, () => {
           expect(res.body[1].value).to.equal('10');
           done();
         });
-      }, 500);
+      }, 100);
```
iamigo: I love shortening the time it takes to run tests, but would this make it more flappy?

tests/enforceToken/sampleUpsertBulk.js
```diff
@@ -102,7 +102,7 @@ describe('api: POST ' + path, () => {
         value: '4',
       },
     ])
-    .expect(200)
+    .expect(constants.httpStatus.OK)
     .end((err /* , res */) => {
       if (err) {
         done(err);
```
iamigo: make this `return done(err);` here

tests/cache/models/samples/upsertBulk.js
shriramshankar: We started having the timeouts for bulkUpsert tests only recently, so my guess is a tiny delay is enough. If this comes up as a flapper I will fix it immediately. Is this ok?

jobQueue/jobWrapper.js
shriramshankar: I was thinking about this too but in other places where we are calling this, we do not do anything. So, I left it this way to be consistent.  Also, this will not happen in prod or any environment where a token is required. Let me know if you still want to log something here.

iamigo: This is fine.

PR #490, opened 2017-08-09
Compress samples

annyhe:
- using base64 encoded string in publish. This is more memory efficient than hex (aka base 16).
- uncompressed buffer bytes: 1400/sample. 
- base 64 encoded string (from compressed buffer bytes) : around 850/sample. 
- hex encoded string (from compressed buffer bytes) : around 1250/sample. 
It is worth considering ascii, which is supposed to be more space efficient.
However computation to encode might be more taxing.


annyhe:
According to stack overflow: **what is the difference between hex and base64**
The radix. Base64 is base-64, surprise, and hex is base-16.
The encoding: base-64 encodes 3 source bytes into 4 base-64 characters (http://en.wikipedia.org/wiki/Base64#Examples); hex encodes 1 byte into 2 hex characters.

PR #492, opened 2017-08-09
generatorTemplate swagger

jgraff2:


api/v1/swagger.yaml
```diff
@@ -1988,27 +1997,632 @@ paths:
...
+          $ref: "#/responses/400"
+        403:
+          $ref: "#/responses/403"
+        default:
+          $ref: "#/responses/genericError"
+
+  # ---------------------------------------------------------------------------
+
+  /generators/{key}:
+    x-swagger-router-controller: generators
```
shriramshankar: are /generator/** swagger definitions from an older commit? The master branch has this already.

api/v1/swagger.yaml
```diff
@@ -1988,27 +1997,632 @@ paths:
...
+                  Array of tag names. Tag name: Max 60 characters; case
+                  insensitive; allows alpha-numeric characters, underscore (_)
+                  and dash (-). Tag names cannot start with a dash (-).
+              connection:
+                type: object
+                description: Information about the connection.
+              context:
+                type: object
+                description: >-
+                  Variables accessible inside the transform function.
```
iamigo: they're also available inside the toUrl fn *and* for variable substitution in connection.url

api/v1/swagger.yaml
```diff
@@ -1988,27 +1997,632 @@ paths:
...
+              - -name
+              - -version
+              - -createdAt
+              - -updatedAt
+        -
+          $ref: "#/parameters/limitParam"
+        -
+          $ref: "#/parameters/offsetParam"
+        -
+          name: description
```
iamigo: I don't think we are allowing query parameter filtering on description fields anywhere

api/v1/swagger.yaml
```diff
@@ -1988,27 +1997,632 @@ paths:
...
+             a negative sign to indicate that a aspect with that tag should
+             be excluded. For example, ?tags=-BAZ,FOO will return only the
+             aspects with tag name not equal to BAZ or FOO. A mix of tag
+             names with and without minus signs, for example ?tags=BAZ,-FOO
+             are not allowed. GeneratorTemplates without tags are not included in the
+             include result set.
+          type: string
+          required: false
+        -
+          name: bulk
```
iamigo: oops "bulk" needs to be moved _inside_ of the connection object as a connection attribute in the db model, and no need to have it here for query parameter filtering

api/v1/swagger.yaml
```diff
@@ -1988,27 +1997,632 @@ paths:
...
+              name:
+                type: string
+                readOnly: true
+                description: >-
+                  The generatorTemplate's name.
+              version:
+                type: string
+                description: >-
+                  The version, in semantic versioning format ( MAJOR.MINOR.PATCH )
+              bulk:
```
iamigo: see comment above about bulk

api/v1/swagger.yaml
jgraff2: I didn't add this. Looks like the diff is really messed up.

api/v1/swagger.yaml
jgraff2: will update.

api/v1/swagger.yaml
jgraff2: We are in aspects, subjects, and generators.
Should I remove them?

api/v1/swagger.yaml
jgraff2: ok, I’ll update the model in the other PR, and delete it here.

api/v1/swagger.yaml
jgraff2: This is actually on the generator. I didn't update this part, the diff is messed up. But the description is the same for contextDefinition on the generatorTemplate. I will update it there.

iamigo:
@jgraff2 OK don't remove the filtering by description from the other resources, but let's not add it to any other resources

api/v1/swagger.yaml
iamigo: remove this one but not the others

PR #494, opened 2017-08-09
generator - change keywords to tags

jgraff2:


jgraff2:
The tests won't pass until the swagger changes are merged. Waiting to add reviewers.

tests/api/v1/generators/get.js
```diff
@@ -191,4 +192,32 @@ describe(`api: GET ${path}`, () => {
...
+    });
+  });
+
+  it('find by tags, not found', (done) => {
+    api.get(`${path}?tags=blah`)
+    .set('Authorization', token)
+    .expect(constants.httpStatus.OK)
+    .end((err, res) => {
+      if (err) {
+        done(err);
```
shriramshankar: We can take care of this later but we want to start adding a return in front of done(err). This will avoid the "done() called multiple times" error message when the test fails.

db/model/generator.js
```diff
@@ -79,7 +79,7 @@ module.exports = function generator(seq, dataTypes) {
       type: dataTypes.ARRAY(dataTypes.STRING(constants.fieldlen.normalName)),
       allowNull: false,
     },
-    keywords: {
+    tags: {
```
iamigo: Doesn't this need a migration?

jgraff2: There aren't any Generators or GeneratorTemplates in prod yet.
Does it need a migration if there's nothing there already?

iamigo: I think you have to change the column name in the migration otherwise it won't happen from the sync cuz the table already exists

jgraff2: ok will do.

jgraff2: migration added.

migrations/20170816224803-generatorChangeKeywordsToTags.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
iamigo: 2017

migrations/20170816224803-generatorChangeKeywordsToTags.js
```diff
@@ -0,0 +1,32 @@
...
+module.exports = {
+  up(qi, Sequelize) {
+  /*
+   Add altering commands here.
+   Return a promise to correctly handle asynchronicity.
+
+   Example:
+   return qi.createTable('users', { id: Sequelize.INTEGER });
+  */
+  return qi.renameColumn('Generators', 'keywords', 'tags');
```
iamigo: indent

migrations/20170816224803-generatorChangeKeywordsToTags.js
```diff
@@ -0,0 +1,32 @@
...
+module.exports = {
+  up(qi, Sequelize) {
+    /*
+     Add altering commands here.
+     Return a promise to correctly handle asynchronicity.
+
+     Example:
+     return qi.createTable('users', { id: Sequelize.INTEGER });
+    */
+    return qi.renameColumn('Generators', 'keywords', 'tags');
```
iamigo: isn't this migration in another pr?? this is going to fail the way it's written, since it doesn't check for the existence of the column

jgraff2: This is for Generators, the other is for GeneratorTemplates. This PR isn't directly related to the api changes, it's just to change the column name for consistency, like we discussed at standup last week.

I don't see the other migrations checking if the column exists. I've tested it locally with no problems.

iamigo: ah, i didn't recognize the difference between generator and generatortemplate - thx

PR #496, opened 2017-08-09
jobCleanup test changes - tweak delays to prevent flapping

jgraff2:


tests/clock/jobCleanup.js
```diff
@@ -175,7 +175,7 @@ describe('jobCleanup', () => {
     const batchSize = 5;
 
     it('skip 3', (done) => {
-      const delay = 199;
+      const delay = 190;
```
jgraff2: I had set these to 99 because I was seeing flappers where it was returning too few results. Now we're seeing flappers with too many, so I'm putting it down to 90.
In the last changes I made I also significantly increased the duration so the margin of error is much bigger now, so I don't think the original problem will come up.

jgraff2:
@iamigo yeah, I spent a while trying to see if there was another way. The problem is the job cleanup relies on the specific end time of the jobs and kue doesn't expose this to be set manually, so the only straightforward way to test it is to actually create the jobs and wait for them to finish. 
I had some other ideas of how to get around this that I gave up on. I'm hoping this change will be enough, but if we see this test flapping again I will revisit them.

PR #497, opened 2017-08-10
Check access to resource on POST/PATCH/PUT/DELETE

kriscfoster:


kriscfoster:
Changing the perms of OOTB refocus user before pushing the rest of this to ensure API tests don't fail as forbidden.

iamigo:
@kriscfoster did you forget to add reviewers to this?

kriscfoster:
@iamigo I was waiting to see what the result of the IMC access fields migrations were. This uses them. Is all of that in prod now or next release?

db/model/profile.js
```diff
@@ -124,6 +124,16 @@ module.exports = function profile(seq, dataTypes) {
           .catch((err) => reject(err));
         });
       }, // isAdmin
+
+      hasWriteAccess(profileId, modelName) {
+        const accessModel = modelName.charAt(0).toLowerCase() + modelName.slice(1) + 'Access';
```
tausifmuzaffar: botData does not have "botDatas" is there another way to do this?

api/v1/helpers/authUtils.js
```diff
@@ -40,6 +40,28 @@ function getUser(req) {
...
 
 /**
+ * Determines whether the user has write access to a model
+ *
+ * @param {Request} req - The request object
+ * @param {String} modelName - The name of the model to write to
+ * @returns {Promise} - A promise which resolves to true if the user has
+ *  write access to the resource
+ */
+function hasWriteAccess(req, modelName) {
```
iamigo: Please rename this function to "profileHasWriteAccess". Otherwise people will confuse it with record-level write access.

api/v1/helpers/authUtils.js
```diff
@@ -40,6 +40,28 @@ function getUser(req) {
 } // getUser
 
 /**
+ * Determines whether the user has write access to a model
```
iamigo: `Determines whether the user's profile has write access to a model.`

api/v1/helpers/authUtils.js
```diff
@@ -40,6 +40,28 @@ function getUser(req) {
...
+ *
+ * @param {Request} req - The request object
+ * @param {String} modelName - The name of the model to write to
+ * @returns {Promise} - A promise which resolves to true if the user has
```
iamigo: ... if the user's profile has...

db/model/profile.js
iamigo: Better to create a mapping of the model to the profile access field name than to rely on naming convention... maybe add a static function to each of the models which returns its profile access column name?

db/model/profile.js
```diff
@@ -124,6 +124,16 @@ module.exports = function profile(seq, dataTypes) {
...
+        return new Promise((resolve, reject) => {
+          Profile.findById(profileId)
+          .then((p) => resolve(p &&
+            p[accessModel] === 'rw'.toLowerCase()))
+          .catch((err) => reject(err));
```
iamigo: `.catch(reject);`

kriscfoster:
@iamigo @tausifmuzaffar Made those changes and added a static function to all models which currently have an access column in profiles. I've tested it and it works but would like to add the check in controllers in a different PR.

db/model/aspect.js
```diff
@@ -135,6 +135,10 @@ module.exports = function aspect(seq, dataTypes) {
         return assoc;
       },
 
+      getAccessField() {
```
iamigo: I think "getProfileAccessField()" is a little clearer

db/model/user.js
```diff
@@ -58,6 +58,10 @@ module.exports = function user(seq, dataTypes) {
         return assoc;
       },
 
+      getAccessField() {
```
iamigo: pls add tests to all the db models you're updating here

kriscfoster:
@iamigo Added tests to models. I will add the write access check to controllers in the next PR so this will be enforced. This will mean that a valid token from a profile with 'rw' access to the resource will be required for all write operations on: aspect / bot / event / lens / perspective / profile / room / roomType /sample / subject / user

iamigo:
@tausifmuzaffar this one's waiting for your approval and merge

iamigo:
@kriscfoster please resolve jscs enforcement issues

tests/db/model/event/find.js
```diff
@@ -57,7 +57,7 @@ describe('tests/db/model/event/find.js >', () => {
   describe('Find event', () => {
     it('ok, event by roomId', (done) => {
       Room.findAll()
-      .then((rooms) => {
+      .then(rooms => {
```
iamigo: should be `.then((rooms) => Event.findOne({ where: { roomId: rooms[ZERO].id } }))`

PR #498, opened 2017-08-10
Profiles Migration Fix

tausifmuzaffar:
Add check for if property already exists for Profiles Migration

### DIFF
--- a/migrations/20170802152927-add_imc_profile_migration.js
+++ b/migrations/20170802152927-add_imc_profile_migration.js
@@ -11,73 +11,81 @@ const TBL = 'Profiles';
 
 module.exports = {
   up: function (qi, Sequelize) {
+    let attr;
     return qi.describeTable(TBL)
-    .then((profile) => {
-      if (!profile.hasOwnProperty('botAccess')) {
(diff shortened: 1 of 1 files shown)


tausifmuzaffar:
@iamigo @annyhe  this PR is for the migration that failed in production;
Checking for the column before adding it

migrations/20170802152927-add_imc_profile_migration.js
```diff
@@ -1,27 +1,77 @@
 module.exports = {
   up: function (qi, Sequelize) {
-    return qi.addColumn('Profiles', 'botAccess', {
-      type: Sequelize.ENUM('r', 'rw'),
-      defaultValue: 'r',
+    qi.sequelize.query('SELECT * FROM "Profiles" LIMIT 1',
```
iamigo: Better to use qi.describeTable than to try to select a record from the table. See /refocus/migrations/20161115215050-fixglobalconfig.js for an example where that is used to check for existence of column.

tausifmuzaffar: I tried this originally, but I forgot the "return" and it didnt work; Ive updated the PR

PR #499, opened 2017-08-10
fixes to bulkUpsert flapper and some minor modifications to get aspect tests

shriramshankar:
fixes to bulkUpsert flapper and some minor modifications to get aspect tests

### DIFF
--- a/tests/cache/models/samples/upsertBulk.js
+++ b/tests/cache/models/samples/upsertBulk.js
@@ -17,7 +17,6 @@ const constants = require('../../../../api/v1/constants');
 const tu = require('../../../testUtils');
 const rtu = require('../redisTestUtil');
 const samstoinit = require('../../../../cache/sampleStoreInit');
-const redisClient = require('../../../../cache/redisCache').client.sampleStore;
 const bulkUpsert = require('../../../../cache/models/samples.js')
                         .bulkUpsertByName;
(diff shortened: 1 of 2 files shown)


tests/cache/models/samples/upsertBulk.js
```diff
@@ -358,20 +357,26 @@ describe('api::redisEnabled::POST::bulkUpsert ' + path, () => {
...
-        api.get('/v1/samples?name=' +
-          `${tu.namePrefix}Subject|${tu.namePrefix}Aspect1`)
-        .set('Authorization', token)
-        .end((err, res) => {
-          if (err) {
-            done(err);
-          }
+        /*
+         * the bulk api is asynchronous. The delay is used to give sometime for
+         * the upsert operation to complete
```
iamigo: What if we check the GET /.../status endpoint in a while loop until status is complete, and *then* try to GET the sample?

iamigo: @shriramshankar

shriramshankar: We have some thing similar in tests/jobQueue/v1/getBulkUpsertStatus.js and tests/jobQueue/v1/bulkUpsert.js. Here we are testing the bulkUpsert api without the workers.

PR #428, opened 2017-07-11
gzip samples before publishing it to redis

shriramshankar:

PR #500, opened 2017-08-10
Compress samples

annyhe:
- using base 64 encoding for strings
- publish stringified compressed samples
- Memory savings: 
- Compressed stringified samples size: 850 bytes vs uncompressed stringified samples:  1250 bytes
- Compressed buffer size around 340. uncompressedBuffer size around 550

### DIFF
--- a/db/helpers/common.js
+++ b/db/helpers/common.js
@@ -18,6 +18,7 @@ const dbconf = require('../../config').db;
 const channelName = require('../../config').redis.channelName;
 const revalidator = require('revalidator');
 const ValidationError = require('../dbErrors').ValidationError;
+const zlib = require('zlib');
 
 // jsonSchema keys for relatedLink
 const jsonSchemaProperties = {
@@ -188,7 +189,18 @@ function publishChange(inst, event, changedKeys, ignoreAttributes) {
   }
 
   if (obj[event]) {
-    pub.publish(channelName, JSON.stringify(obj));
+    const objectAsString = JSON.stringify(obj);
+    zlib.deflate(objectAsString, (err, zippedValue) => {
+      if (err) {
+        console.log('Error deflating!', err);
+        return;
+      }
+
+      // encode the binary zip-data to base64 to
+      // be able to read it
+      const str = zippedValue.toString('base64');
+      return pub.publish(channelName, str);
+    });
   }
 
   return obj;
--- a/realtime/redisPublisher.js
+++ b/realtime/redisPublisher.js
@@ -17,6 +17,7 @@ const pub = require('../cache/redisCache').client.pub;
 const channelName = require('../config').redis.channelName;
 const sampleEvent = require('./constants').events.sample;
 const featureToggles = require('feature-toggles');
+const zlib = require('zlib');
 
 /**
  * When passed an sample object, either a sequelize sample object or
@@ -85,7 +86,17 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
 
   if (obj[event]) {
     const objectAsString = JSON.stringify(obj);
-    return pub.publish(channelName, objectAsString);
+    zlib.deflate(objectAsString, (err, zippedValue) => {
+      if (err) {
+        console.log('Error deflating!');
+        return;
+      }
+
+      // encode the binary zip-data to base64 to
+      // be able to read it
+      const str = zippedValue.toString('base64');
+      return pub.publish(channelName, str);
+    });
   }
 
   return obj;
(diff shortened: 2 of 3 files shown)


annyhe:
The previous PR also has the same error on inflate sample: https://github.com/salesforce/refocus/pull/428/files

realtime/redisSubscriber.js
```diff
@@ -21,14 +22,22 @@ const sub = require('../cache/redisCache').client.sub;
  */
 module.exports = (io) => {
   sub.on('message', (channel, mssgStr) => {
-    // message object to be sent to the clients
-    const mssgObj = JSON.parse(mssgStr);
-    const key = Object.keys(mssgObj)[0];
+    zlib.inflate(new Buffer.from(mssgStr, 'base64'),
```
annyhe: .inflate accepts a string input too. However the error `Error inflating! { Error: incorrect header check
    at Zlib._handle.onerror (zlib.js:370:17) errno: -3, code: 'Z_DATA_ERROR' }`
still happens with string input

annyhe:
tests are passing locally but are timing out on travis CI. Some of the failiing tests are not related to publishing samples though: create token, collectors, profiles. There are 19 of them so that is too many to be a flapper.

db/helpers/common.js
```diff
@@ -188,7 +189,18 @@ function publishChange(inst, event, changedKeys, ignoreAttributes) {
   }
 
   if (obj[event]) {
-    pub.publish(channelName, JSON.stringify(obj));
+    const objectAsString = JSON.stringify(obj);
```
iamigo: the compress/decompress stuff needs to be behind a short-term toggle

db/helpers/common.js
```diff
@@ -188,7 +189,18 @@ function publishChange(inst, event, changedKeys, ignoreAttributes) {
...
+    zlib.deflate(objectAsString, (err, zippedValue) => {
+      if (err) {
+        console.log('Error deflating!', err);
```
iamigo: if we get an error deflating, shouldn't we try to publish the event normally, rather than just give up and "return"?

realtime/redisPublisher.js
```diff
@@ -85,7 +86,17 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
 
   if (obj[event]) {
     const objectAsString = JSON.stringify(obj);
-    return pub.publish(channelName, objectAsString);
+    zlib.deflate(objectAsString, (err, zippedValue) => {
```
iamigo: behind feature flag

realtime/redisPublisher.js
```diff
@@ -85,7 +86,17 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
...
+      if (err) {
+        console.log('Error deflating!', err);
```
iamigo: give up or publish event without compressing?

realtime/redisSubscriber.js
```diff
@@ -21,14 +22,21 @@ const sub = require('../cache/redisCache').client.sub;
  */
 module.exports = (io) => {
   sub.on('message', (channel, mssgStr) => {
-    // message object to be sent to the clients
-    const mssgObj = JSON.parse(mssgStr);
-    const key = Object.keys(mssgObj)[0];
+    zlib.unzip(Buffer.from(mssgStr, 'base64'), (err, uncompressedBuffer) => {
```
iamigo: behind feature flag

we should try to identify here whether the message has been deflated or not--that way if we hit an error deflating it, we can send it raw, and not try to unzip it here

realtime/redisSubscriber.js
```diff
@@ -21,14 +22,21 @@ const sub = require('../cache/redisCache').client.sub;
...
+      if (err) {
+        console.log('Error inflating!', err);
```
iamigo: use winston not console

db/helpers/common.js
```diff
@@ -188,7 +189,18 @@ function publishChange(inst, event, changedKeys, ignoreAttributes) {
   }
 
   if (obj[event]) {
-    pub.publish(channelName, JSON.stringify(obj));
+    const objectAsString = JSON.stringify(obj);
+    zlib.deflate(objectAsString, (err, zippedValue) => {
```
iamigo: Put the zlib stringify/deflate/toString and unzip/parse functions into a util file which can be called from here, from redisPublisher and from redisSubscriber, then add some tests for the two util functions

shriramshankar:
@annyhe @iamigo There is a quite a bit of refracting on #527. Can we come back to this PR once we are fully done with #527 ?

iamigo:
@annyhe and @shriramshankar you two can work it out as to who gets in first - as far as i'm concerned I'd like to get both of these merged because they will both reduce the buffer overflow problem

shriramshankar:
@iamigo @annyhe  We can talk about this in person tomorrow.

annyhe:
@shriramshankar Since I will be on VTO on Thursday, if you can get #527 merged by Thursday evening, we can come back to this PR on Friday.

shriramshankar:
@annyhe ok thanks. Also, there is a sync version of "zlib", can we use that instead of this async version.?

PR #501, opened 2017-08-11
OOTB RefocusUser Profile Access 

kriscfoster:


kriscfoster:
Out of the box RefocusUser Profile getting RW access to existing resources (backwards compatibility)

migrations/20170811092638-ootb_refocus_profile.js
```diff
@@ -0,0 +1,48 @@
...
+      qi.sequelize.query('UPDATE "Profiles" SET "aspectAccess" = \'r\'' +
+        ', "botAccess" = \'r\', "eventAccess" = \'r\'' +
+        ', "lensAccess" = \'r\', "perspectiveAccess" = \'r\'' +
+        ', "roomTypeAccess" = \'r\', "sampleAccess" = \'r\'' +
+        ', "subjectAccess" = \'r\', "userAccess" = \'r\'' +
+        ' WHERE "name"=\'RefocusUser\';'
+      )
+    );
+  },
+};
```
shriramshankar: Can you confirm that you were able to run both the "up" and "down" migrations locally?

kriscfoster: @shriramshankar the migrations work as long as the fields 'botAccess', 'eventAccess' etc are present.

kriscfoster:
@iamigo I think a migration will have to be made for the admin profile similar to migrations/20170811092638-ootb_refocus_profile.js because when the new access columns were added (botAccess etc) the default given was 'r', but the admin should also have 'rw'.

Should I make that migration in this same PR or open a new one? Once this is done I can finish the loop and add checks for access in controllers. If this isn't done first, Admin won't have access to POST to 'Bots', 'Events' etc.

tausifmuzaffar:
@kriscfoster please resolve these conflicts so we can merge this PR

iamigo:
@kriscfoster test failures?

PR #502, opened 2017-08-11
Remove Bot Access

tausifmuzaffar:
Bot Access holding up build so removing it till we can figure out the problem


tausifmuzaffar:
@iamigo @shriramshankar this PR removes all the "botAccess" to profiles that Kris added

PR #505, opened 2017-08-11
Return unpublished subject on populate samples

annyhe:
- on unpublish subject, subject should still be returned from cache
- on GET /v1/subjects, the results are identical regardless of whether cache is on or off

### DIFF
--- a/db/model/subject.js
+++ b/db/model/subject.js
@@ -342,9 +342,9 @@ module.exports = function subject(seq, dataTypes) {
          * When the sample store feature is enabled do the following
          * 1. if subject is changed from unpublished to published -> update subject
          * 2. if subject is changed from published to unpublished -> update subject,
-         *   delete aspect map, remove its samples
-         * 3. if the asbsolutepath of the subject changes: rename the keys,
-         *   update subject, delete aspect map, and remove its samples
+         *   delete subject aspect map, remove its samples
+         * 3. if the asbsolutepath of the subject changes: rename the subject key,
+         *   update subject, delete subject aspect map, and remove its samples
         */
         if (featureToggles.isFeatureEnabled(sampleStoreFeature)) {
           if (inst.changed('absolutePath') ||
--- a/db/model/subject.js
+++ b/db/model/subject.js
@@ -367,7 +367,12 @@ module.exports = function subject(seq, dataTypes) {
               redisOps.addKey(subjectType, inst.absolutePath);
               redisOps.hmSet(subjectType, inst.absolutePath, inst.get());
             } else {
-              subjectUtils.removeFromRedis(inst.absolutePath);
+
+              /*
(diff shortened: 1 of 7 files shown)


annyhe:
tested locally:
- GET /v1/subjects returns the same result, regardless of whether the cache is on or off
- for unpublished subjects, the GET /v1/subject/absPathOfThatSubject returns the subject

db/model/subject.js
```diff
@@ -367,7 +367,13 @@ module.exports = function subject(seq, dataTypes) {
...
               redisOps.hmSet(subjectType, inst.absolutePath, inst.get());
             } else {
-              subjectUtils.removeFromRedis(inst.absolutePath);
+
+              /*
+               * Delete multiple possible entries in the sample master list of
+               * index
+               */
+              redisOps.deleteKeys(sampleType, subjectType, inst.absolutePath);
+              redisOps.hmSet(subjectType, inst.absolutePath, inst.get());
```
annyhe: this is for afterUpdate section: when the subject is unpublished, remove its references to samples, and update it

db/model/subject.js
```diff
@@ -367,7 +367,13 @@ module.exports = function subject(seq, dataTypes) {
               redisOps.addKey(subjectType, inst.absolutePath);
               redisOps.hmSet(subjectType, inst.absolutePath, inst.get());
             } else {
```
shriramshankar: Update the comments in line 341 to 347, to reflect this change.

db/model/subject.js
shriramshankar: Do we have to update the subAspMapType too?

annyhe: Not sure. Is subAspMap only for samples? If so, on unpublish there are no more samples, so subAspMap should be deleted.

annyhe: On master in afterUpdate in subjects, subAspMap does not change when absolutePath changes. Is this a bug?

annyhe:
Please review the updated PR @shriramshankar

db/model/subject.js
```diff
@@ -340,36 +340,35 @@ module.exports = function subject(seq, dataTypes) {
...
-         * 1. if subject is changed from published to unpublished -> delete
-         * 2. if subject is changed from unpublished to published -> add
-         * 3. if the asbsolutepath of the subject changes and the subject is
-         * puslished, rename the keys
+         * 1. if subject is changed from unpublished to published -> update subject
+         * 2. if subject is changed from published to unpublished -> update subject,
+         *   delete aspect map, remove its samples
+         * 3. if the asbsolutepath of the subject changes: rename the keys,
+         *   update subject, delete aspect map, and remove its samples
         */
```
shriramshankar: The rest looks good to me.

annyhe:
@shriramshankar The build passed. Please review

PR #507, opened 2017-08-12
specify redis retry strategy

iamigo:


iamigo:
I couldn't find a way to test the retry by forcing an ECONNRESET. Any ideas?

config/redisConfig.js
```diff
@@ -66,4 +66,15 @@ module.exports = {
...
+    /* Number of times to attempt to reconnect. Default 10. */
+    attempt: +pe.REDIS_RETRY_ATTEMPT || 10,
+
+    backoffFactor: +pe.REDIS_RETRY_BACKOFF_FACTOR || 100,
+
+    backoffMax: +pe.REDIS_RETRY_BACKOFF_MAX || 3000,
+
+    /* Time since the last successful connection. Default one hour. */
+    totalRetryTime: +pe.REDIS_RETRY_TOTAL_RETRY_TIME || 3600000,
+  },
```
shriramshankar: there is a + next to each of the pe object above.

iamigo: The plus will cast the string env var to numeric, and if it can't be cast as a number, we skip to the "||" :)

shriramshankar: ohh, did not know this.

shriramshankar:
@iamigo Just noticed your comment
" I couldn't find a way to test the retry by forcing an ECONNRESET. Any ideas?" I think deleting a redis instance will throw this error. When the redis instance is added back again, the dynos should connect to the added redis, without a restart.

PR #508, opened 2017-08-12
Fix admin controller test

iamigo:


shriramshankar:
Looks like this has commits from the redis config pr too.

PR #518, opened 2017-08-13
tests/cache tweaks

iamigo:


tests/cache/models/aspects/deleteWriters.js
```diff
@@ -222,27 +178,20 @@ describe('api: aspects: permissions', () => {
...
-        }
-
-        return done();
-      });
-        return null;
+        .set('Authorization', token)
+        .expect(constants.httpStatus.OK)
+        .expect((res) => {
+          console.log('\n\n\n>>>>>>>>>>\n', res.body);
+          expect(res.body).to.have.length(1);
```
shriramshankar: console.log

jgraff2:
Suggestion: we could move all the assertions to their own .expect((res) => {...}) block and change all the .ends to .end(done). That way we could get rid of all the "if (err) { return done(err) }" statements.
@iamigo 

@iamigo

iamigo:
Yes, I totally agree, Jon, but that was more than a cut/paste/regex job for me this weekend.

PR #521, opened 2017-08-13
tweaks to tests/clock, tests/config, tests/jobQueue

iamigo:


tests/config/configUtil.js
```diff
@@ -10,164 +10,132 @@
...
-    .to.be.eql([]);
-    done();
+  it('undefined string', () => {
+    expect(configUtil.csvToArray(undefined)).to.be.eql([]);
   });
 
-  it('null string', (done) => {
-    expect(configUtil.csvToArray(null))
-    .to.be.eql([]);
-    done();
```
shriramshankar: I thought we need to call done for all the tests, otherwise, it will timeout.

iamigo: you only need to call done if you're passing done into the it(...)

shriramshankar: Ok. Didn't know that.

PR #522, opened 2017-08-14
UI in response of POST /bot

kriscfoster:
@kamilsmuga @tausifmuzaffar an object containing the name & size of UI zip is in response.. Is this overkill? If so which would be better to have?

### DIFF
--- a/api/v1/controllers/bots.js
+++ b/api/v1/controllers/bots.js
@@ -86,12 +86,15 @@ module.exports = {
     const resultObj = { reqStartTime: new Date() };
     const reqObj = req.swagger.params;
     const seqObj = {};
+    const uiObj = {};
     try {
       for (const param in reqObj) {
         if (reqObj[param].value) {
           if (typeof (reqObj[param].value) === 'object' &&
             param === 'ui') {
             seqObj[param] = reqObj[param].value.buffer;
+            uiObj.name = reqObj[param].value.originalname;
+            uiObj.size = reqObj[param].value.size;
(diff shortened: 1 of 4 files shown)


api/v1/controllers/bots.js
```diff
@@ -86,12 +86,15 @@ module.exports = {
     const resultObj = { reqStartTime: new Date() };
     const reqObj = req.swagger.params;
     const seqObj = {};
+    const uiObj = {};
     try {
       for (const param in reqObj) {
         if (reqObj[param].value) {
           if (typeof (reqObj[param].value) === 'object' &&
```
shriramshankar: I think swagger will take care of this validation.

api/v1/controllers/bots.js
```diff
@@ -86,12 +86,15 @@ module.exports = {
     const resultObj = { reqStartTime: new Date() };
     const reqObj = req.swagger.params;
     const seqObj = {};
+    const uiObj = {};
```
iamigo: no tests?

api/v1/controllers/bots.js
kriscfoster: @shriramshankar I'm making that object to overwrite the value for UI in the response which is long binary data. I'm overwriting it at line 107.

api/v1/controllers/bots.js
kriscfoster: @iamigo I added expects for the UI in response now in api POST & PUT

PR #527, opened 2017-08-16
Publish partial sample

shriramshankar:
This PR has the following changes
- Refactored the "publishSample" method in "redisPublisher.js" by moving the attach subject and aspect to utils.
- Refactored redisSubscriber.js to parse and extract the objects before calling the emitter(). 
-  Added a feature to publish the samples without attaching the subject/aspect and letting the subscriber attach them. 
- Unit tests to prove these changes work.
- Some minor changes to fix flappers not related to the above changes.

### DIFF
--- a/config/toggles.js
+++ b/config/toggles.js
@@ -126,7 +126,12 @@ const shortTermToggles = {
 
   returnUser: environmentVariableTrue(pe, 'RETURN_CREATEDBY_ON_TOKEN_INPUT'),
 
-  fastFailDuplicateSubject: environmentVariableTrue(pe, 'FAST_FAIL_DUPLICATE_SUBJECT'),
+  fastFailDuplicateSubject: environmentVariableTrue(pe,
+    'FAST_FAIL_DUPLICATE_SUBJECT'),
+
+  // publish partial sample to the subscribers
+  publishPartialSample: environmentVariableTrue(pe, 'PUBLISH_PARTIAL_SAMPLE'),
+
 
 }; // shortTermToggles
 
--- a/config/toggles.js
+++ b/config/toggles.js
@@ -132,7 +132,6 @@ const shortTermToggles = {
   // publish partial sample to the subscribers
   publishPartialSample: environmentVariableTrue(pe, 'PUBLISH_PARTIAL_SAMPLE'),
 
-
 }; // shortTermToggles
 
 featureToggles.load(Object.assign({}, longTermToggles, shortTermToggles));
--- a/realtime/redisPublisher.js
+++ b/realtime/redisPublisher.js
@@ -98,6 +98,7 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
  */
 function publishPartialSample(sampleInst, event) {
   const eventType = event || getSampleEventType(sampleInst);
+
   // will be over written when unwrapping json.stringified fields
   const sample = sampleInst.get ? sampleInst.get() : sampleInst;
   publishObject(sample, eventType);
--- a/realtime/redisPublisher.js
+++ b/realtime/redisPublisher.js
@@ -92,6 +92,19 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
 } // publishChange
 
 /**
+ * [publishPartialSample description]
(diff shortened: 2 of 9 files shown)


realtime/redisPublisher.js
```diff
@@ -92,6 +90,22 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
 } // publishChange
 
 /**
+ * Publishes the sample without attaching the related subject and the aspect to
+ * the redis channel
+ * @param  {Object} sampleInst - The sampel instance to be published
```
annyhe: did you mean "sample" instead of "sampel?

config/toggles.js
```diff
@@ -126,7 +126,11 @@ const shortTermToggles = {
 
   returnUser: environmentVariableTrue(pe, 'RETURN_CREATEDBY_ON_TOKEN_INPUT'),
 
-  fastFailDuplicateSubject: environmentVariableTrue(pe, 'FAST_FAIL_DUPLICATE_SUBJECT'),
+  fastFailDuplicateSubject: environmentVariableTrue(pe,
+    'FAST_FAIL_DUPLICATE_SUBJECT'),
+
+  // publish partial sample to the subscribers
+  publishPartialSample: environmentVariableTrue(pe, 'PUBLISH_PARTIAL_SAMPLE'),
```
shriramshankar: I can change the name of the config if we do not like it.

realtime/redisPublisher.js
```diff
@@ -105,76 +123,19 @@ function publishObject(inst, event, changedKeys, ignoreAttributes) {
...
-        publishObject(sample, eventType);
-      }
-    }
-
+  const useSampleStore =
+    featureToggles.isFeatureEnabled('enableRedisSampleStore');
+  return rtUtils.attachAspectSubject(sampleInst, useSampleStore, subjectModel,
+    aspectModel)
+  .then((sample) => {
+    publishObject(sample, eventType);
```
annyhe: so in publishSample we publish the entire sample, but elsewhere we only publish the partial sample, if the flag publish partial sample is true?

realtime/redisSubscriber.js
```diff
@@ -24,11 +26,30 @@ module.exports = (io) => {
...
-    /*
-     * pass on the message received through the redis subscriber to the socket
-     * io emitter to send data to the browser clients.
-     */
-    emitter(io, key, mssgObj);
+      // assign the subject db model if sampleStore is not enabled
+      const subjectModel =
+        useSampleStore ? undefined :
+          require('../db/index').Subject; // eslint-disable-line global-require
+      rtUtils.attachAspectSubject(parsedObj, useSampleStore, subjectModel)
```
annyhe: we can conditionally pass in the subjectModel, correct? If sampleStore is off, pass in subjectModel, else don't pass it in. Thus we don't need the useSampleStore param

tests/realtime/realtimeUtils.js
```diff
@@ -216,5 +216,21 @@ describe('realtime utils Tests:', () => {
         }
       });
     });
+
+    describe('attachAspectSubject tests', () => {
+      it('useSampleStore = true', () => {
+        realtimeUtils.attachAspectSubject(looksLikeSampleObjNA, true)
```
annyhe: we need more tests for this. Can we do a test where the 
1) cache is on, and the subject and aspect are attached?
2) cache is off, and the subject and aspect are attached?

realtime/redisSubscriber.js
shriramshankar: 1. Yes, we can do it that way too but that will have some if statements with 5 lines of code VS this. If we think using "if " statement is more readable, I can change it. 
2. aspectModel is not being passed here. attachAspectSubject takes in aspectModel because it is used by the publisher too and sample post and sample deleteReleatedLinks call the publisher by passing it the aspectModel.

tests/realtime/realtimeUtils.js
shriramshankar: Yes, will add it in  tests/realtime/redisPublisher.js.

shriramshankar:
@iamigo @annyhe Updated the PR.

PR #531, opened 2017-08-17
move tag matching to db query to allow advanced tag logic in limit qu…

jgraff2:
…eries


api/v1/constants.js
```diff
@@ -42,8 +42,10 @@ module.exports = {
   SEQ_DESC: 'DESC',
   SEQ_LIKE: '$iLike',
   SEQ_CONTAINS: '$contains',
+  SEQ_OVERLAP: '$overlap',
```
iamigo: yay!

api/v1/swagger.yaml
```diff
@@ -215,7 +215,7 @@ paths:
           items:
             type: string
             maxLength: 60
-            pattern: ^[0-9A-Za-z_][0-9A-Za-z_\\-]{1,59}$
+            pattern: ^[0-9A-Za-z_\\-][0-9A-Za-z_]{1,59}$
```
iamigo: good catch!

PR #532, opened 2017-08-18
Don't require token if enableAccessToken is disabled and the resource…

jgraff2:
… is not write protected.

### DIFF
--- a/api/v1/helpers/verbs/utils.js
+++ b/api/v1/helpers/verbs/utils.js
@@ -182,18 +182,25 @@ function isWritable(req, modelInst) {
       jwtUtil.getTokenDetailsFromRequest(req)
       .then((resObj) => modelInst.isWritableBy(resObj.username))
       .then((ok) => ok ? resolve(modelInst) :
-          reject(new apiErrors.ForbiddenError())
+        reject(new apiErrors.ForbiddenError(
+          'Resource not writable for provided token'))
       )
       .catch((err) => reject(err));
     } else if (req.user) {
       // try to use the logged-in user
       modelInst.isWritableBy(req.user.name)
       .then((ok) => ok ? resolve(modelInst) :
-          reject(new apiErrors.ForbiddenError())
+        reject(new apiErrors.ForbiddenError(
+          'Resource not writable by this user'))
       )
       .catch((err) => reject(err));
     } else {
-      reject(new apiErrors.ForbiddenError());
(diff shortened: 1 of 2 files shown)


api/v1/helpers/verbs/utils.js
```diff
@@ -182,18 +182,25 @@ function isWritable(req, modelInst) {
...
       )
       .catch((err) => reject(err));
     } else {
-      reject(new apiErrors.ForbiddenError());
+      // check if isWritable by default (no writers)
+      modelInst.isWritableBy()
+      .then((ok) => ok ? resolve(modelInst) :
+        reject(new apiErrors.ForbiddenError('Resource is write protected'))
+      )
+      .catch((err) => reject(err));
```
iamigo: `catch(reject)` should work

api/v1/helpers/verbs/utils.js
```diff
@@ -182,18 +182,25 @@ function isWritable(req, modelInst) {
...
       .then((ok) => ok ? resolve(modelInst) :
-          reject(new apiErrors.ForbiddenError())
+        reject(new apiErrors.ForbiddenError(
+          'Resource not writable by this user'))
       )
       .catch((err) => reject(err));
     } else {
-      reject(new apiErrors.ForbiddenError());
+      // check if isWritable by default (no writers)
+      modelInst.isWritableBy()
```
iamigo: I don't understand this part... what does isWritableBy return when you don't pass a username or userid argument?

jgraff2: It will return true if there are no writers, false if there are.

iamigo: Can you expand the comment here to describe this?

api/v1/helpers/verbs/utils.js
```diff
@@ -182,18 +182,27 @@ function isWritable(req, modelInst) {
...
+        reject(new apiErrors.ForbiddenError(
+          'Resource not writable by this user'))
       )
-      .catch((err) => reject(err));
+      .catch(reject);
     } else {
-      reject(new apiErrors.ForbiddenError());
+      // check if isWritable with no user
+      // when not passed a user, isWritable will return true if
+      // the resource is not write protected, false if it is
```
shriramshankar: I think we want to use 
`/*
*
*/`  for multi-line comments instead of `//`. We can talk care of this later.

PR #533, opened 2017-08-18
db level: sample should not be CREATEd, or UPSERTed when subject is unpublished

annyhe:
db level only


tests/db/helpers/common.js
```diff
@@ -126,22 +126,6 @@ describe('tests/db/helpers/common.js >', () => {
...
-      }))
-      .then((samp) =>
-        common.sampleAspectAndSubjectArePublished(tu.db.sequelize, samp))
-      .then((pub) => {
-        expect(pub).to.equal(false);
-      })
-      .then(() => done())
-      .catch(done);
-    });
-
```
shriramshankar: why are you removing this test?

shriramshankar: Every thing else looks good.

annyhe: because this test tests a wrong assumption: the request to upsert sample with a unpublished subject should fail, instead of succeed.

PR #534, opened 2017-08-18
minor fix in doPost so on non-auth error, do not proceed to postPromise.then

annyhe:
Problem:
when POST sample with non-auth error, 
u.handleError returns nothing, so postPromise.then((o) => ... runs with 
undefined o. This throws an unintentional TypeError:
`TypeError: Cannot read property 'name' of undefined
    at Object.attachAspectSubject (/Users/anny.he/workspace/refocus/realtime/utils.js:330:27)
    at Object.publishSample (/Users/anny.he/workspace/refocus/realtime/redisPublisher.js:128:18)
    at postPromise.then (/Users/anny.he/workspace/refocus/api/v1/helpers/verbs/doPost.js:114:17)`
as the code tries to attach subjects and aspects to an undefined sample.
The fix is simple: throw err instead of handleError, to bypass the postPromise.then block

### DIFF
--- a/api/v1/helpers/verbs/doPost.js
+++ b/api/v1/helpers/verbs/doPost.js
@@ -80,7 +80,9 @@ function doPost(req, res, next, props) {
           props.model.create(toPost);
       }
 
-      return u.handleError(next, err, props.modelName);
+      // non FORBIDDEN error. Throw it to be caught by the latter .catch.
+      // this bypasses the postPromise.then function
+      throw err;
     });
   } else if (props.modelName === 'Sample') {
 
@@ -96,10 +98,6 @@ function doPost(req, res, next, props) {
   }
 
   return postPromise.then((o) => {
-    if (!o) { // could be here from u.handleError
-      return;
-    }
-
     resultObj.dbTime = new Date() - resultObj.reqStartTime;
     u.logAPI(req, resultObj, o);
 
--- a/api/v1/helpers/verbs/doPost.js
+++ b/api/v1/helpers/verbs/doPost.js
@@ -73,7 +73,6 @@ function doPost(req, res, next, props) {
       return props.model.create(toPost);
     })
     .catch((err) => {
-
       // if no user found, proceed with post sample
       if (err.status === httpStatus.FORBIDDEN) {
         return isCacheOn ?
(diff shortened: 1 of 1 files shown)


annyhe:
To test: 
mocha tests/cache/models/samples/post.js
should run successfully without any errors. 
On master:
two tests out of tests/cache/models/samples/post.js, which are supposed to fail with non-auth error, also trigger the TypeError, as mentioned above.

api/v1/helpers/verbs/doPost.js
```diff
@@ -97,6 +96,10 @@ function doPost(req, res, next, props) {
   }
 
   return postPromise.then((o) => {
+    if (!o) { // could be here from u.handleError
+      return;
```
iamigo: does this mean the user is getting back the error properly now? where is the test to show this?

shriramshankar: I kind of think o will never be "falsey". None of the postPromise will return a "falsey" value. The post promise will either throw an error which will be caught by the catch block below or it will return a created instance. Correct?

annyhe: not quite. o is falsey when u.handleError is called, which means postPromise.then returns a falsey value.

annyhe: I can add a test to show typeError(due to o being falsey) is not being thrown.

annyhe: Currently refactoring the doPost function to make it more testable. There will be a lot more tests coming.

annyhe: an alternative is to throw err on line 83, instead of u.handleError, which is not a promise and hence postPromise.then returns undefined.

api/v1/helpers/verbs/doPost.js
```diff
@@ -73,15 +73,16 @@ function doPost(req, res, next, props) {
...
       // if no user found, proceed with post sample
       if (err.status === httpStatus.FORBIDDEN) {
         return isCacheOn ?
           redisModelSample.postSample(req.swagger.params, false) :
           props.model.create(toPost);
       }
 
-      return u.handleError(next, err, props.modelName);
+      // non FORBIDDEN error. Throw it to be caught by the latter .catch.
+      // this bypasses the postPromise.then function
```
shriramshankar: I think we want ot use 
/*
*
*/
for multi line comments.

annyhe: ok will update

annyhe:
ready for review @shriramshankar @iamigo

PR #537, opened 2017-08-19
move tag matching to db query - fix tag param regex

jgraff2:


api/v1/swagger.yaml
```diff
@@ -215,7 +215,7 @@ paths:
           items:
             type: string
             maxLength: 60
-            pattern: ^[0-9A-Za-z_\\-][0-9A-Za-z_]{1,59}$
+            pattern: ^[0-9A-Za-z_-]{1,60}$
```
jgraff2: was there a reason the dash was being escaped?

jgraff2: Added the dash back in to make sure it allows tags with a dash in the middle, and condensed it since the first character can be a dash in the param.

jgraff2: Also, was there a reason the dash needed to be escaped? I don't think so.

iamigo: can you confirm we have tests for dash?

PR #543, opened 2017-08-22
RoomType Bots API

kriscfoster:
- GET / POST / PATCH RoomTypes returns response with Bots.
- POST / PATCH RoomTypes creates connection between roomTypes & Bots.
- Added tests.


db/model/roomType.js
```diff
@@ -123,19 +123,21 @@ module.exports = function roomType(seq, dataTypes) {
             resolve(inst);
           }
 
-          inst._previousDataValues.bots.forEach((botName) => {
-            seq.models.Bot.findOne({
-              where: {
-                name: {
-                  $iLike: botName,
+          if (inst._previousDataValues.bots) {
+            inst._previousDataValues.bots.forEach((botName) => {
```
iamigo: The work starting on line 127 and the work starting on line 142 will be occurring at the same time since you're not chaining them in a promise chain. Is that intentional? Is that safe?

kriscfoster: @iamigo This is something that I will need to change as it may not be safe. Thanks.

db/helpers/roomTypeUtils.js
```diff
@@ -152,7 +152,7 @@ function validateBotsArray(inst, seq) {
   const bots = inst.dataValues.bots;
 
   return new seq.Promise((resolve, reject) => {
-    if (!bots || !inst.changed('bots')) {
+    if (!bots || !inst.changed('bots') || bots.length === 0) {
```
shriramshankar: `!bots.length`; if you want to change it.

db/model/roomType.js
```diff
@@ -123,34 +123,41 @@ module.exports = function roomType(seq, dataTypes) {
...
+                  },
+                })
+                .then((o) => {
+                  inst.removeBots(o)
+                  .catch(reject);
+                });
+              });
+            }
+
+            resolve();
```
iamigo: This resolve means we don't proceed to the "then" on line 145... is that the intention? In that case, the promise could resolve empty or if goes all the way down to line 159 it resolves with the instance.

kriscfoster: @iamigo that "resolve()" is for a promise inside the promise that is being returned. I've tested it and I've gotten past the "then" on 145.

PR #547, opened 2017-08-23
Release phase

iamigo:
If running in heroku, migrations will NOT be executed during the regular node process start; instead, they will be executed just once at release time by executing `npm run on-release` as specified in Procfile.


scripts/onRelease.js
```diff
@@ -0,0 +1,14 @@
...
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * ./scripts/onRelease.js
+ *
+ * Executed during the release phase in a heroku environment.
+ */
```
harshkothari410: Do we need this file ? This is just one liner that's why.  can we do "on-release": "node db/migrate.js" ?

iamigo: We don't *need* it right now, but I wanted to set up the structure for adding stuff into the release phase as we find more uses for it.

harshkothari410: Can you add this to comment in file ?

Procfile
```diff
@@ -1,3 +1,4 @@
-web: npm start
-clock: npm run start-clock
-worker: npm run start-worker
+clock: IS_HEROKU=true npm run start-clock
+release: IS_HEROKU=true npm run on-release
+web: IS_HEROKU=true npm start
+worker: IS_HEROKU=true npm run start-worker
```
shriramshankar: I think "procfiles" are very specific to Heroku and if this application is not run on Heroku, they are just ignored. Considering this, wondering if we need the flag?

iamigo: Heroku will run the release as long as there's a "release" entry in Procfile, the presence of the env var is so that we can know within our app whether it is running in Heroku or not and whether we need to run migrations etc. ourselves in the start script or not.
By setting the env var here, we don't have to rely on an admin configuring the env var (which was my first attempt at this) or rely on a kludgey hack to check the process.env.NODE variable for the presence of the string "heroku" (which was my 2nd attempt, but I rejected because heroku could decide to change how they populate that env var at any time). By setting it explicitly in the Procfile, it will be reliable that it's true when Refocus is running in Heroku.

PR #549, opened 2017-08-23
fixed flapper by reducing the number of samples upserted

shriramshankar:


tests/jobQueue/v1/getBulkUpsertStatus.js
```diff
@@ -171,13 +173,13 @@ describe('tests/jobQueue/v1/getBulkUpsertStatus.js, ' +
     });
   });
 
-  it('Even when hundreds of upserts fail, the valid upsert should ' +
+  it('Even when lots of upserts fail, the valid upsert should ' +
```
iamigo: The code under test here is the same whether it's hundreds or 20 or 3, right?

shriramshankar: yes is it the same for most part. Only change is in the number of samples upserted.

iamigo: smaller number of samples could mean smaller setTimeout?
(500ms times multiple tests is expensive)

shriramshankar: Yes, definitely. But not sure where to draw the line. 400ms?

iamigo: Yeah, I guess the whole point is we can't afford to have it be flappy, so I guess it's more important to err on the side of slower but solid.

PR #551, opened 2017-08-23
default limit

jgraff2:


config.js
```diff
@@ -111,6 +111,8 @@ const JOB_QUEUE_TTL_SECONDS_SYNC = pe.TTL_KUE_JOBS_SYNC
 // set time interval for enableQueueStatsActivityLogs
 const queueStatsActivityLogsInterval = 60000;
 
+const DEFAULT_LIMIT = pe.DEFAULT_LIMIT || 10000;
```
iamigo: Can we make this `GET_REQUEST_DEFAULT_LIMIT`? Otherwise it feels ambiguous, confusing.

iamigo: Also, maybe `const GET_REQUEST_DEFAULT_LIMIT = +pe.GET_REQUEST_DEFAULT_LIMIT || 10000;` (the plus sign to cast the env var to number, and if it can't be cast to number, it'll go to the "|| 10000")

tests/api/v1/aspects/getWithLimit.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
iamigo: 2017 (for all the new files you're adding)

tests/api/v1/generators/utils.js
```diff
@@ -43,7 +43,7 @@ const GENERATOR_SIMPLE = {
  * @returns {Object} - Generator object
  */
 function getGenerator() {
-  return GENERATOR_SIMPLE;
+  return JSON.parse(JSON.stringify(GENERATOR_SIMPLE));
```
iamigo: riiiiiight--much better place to do this :)

tests/api/v1/samples/getWithLimit.js
```diff
@@ -0,0 +1,55 @@
+/**
```
iamigo: Test for when sample store is in postgres, sample store is in redis?

tests/api/v1/aspects/getWithLimit.js
jgraff2: I copied this from an existing file. Should we be updating all the files to 2017?

iamigo: Only *new* files we're adding. Anything written in 2016 should stay 2016.

tests/api/v1/samples/getWithLimit.js
jgraff2: I'm not using the default limit for the sample store. Should I be? The limit is applied after all the samples are retrieved so it won't help with performance.

iamigo: oh, yeah, i hadn't thought that through

for consistency, we *should* be implementing limit/offset on find samples too (esp. since this whole thing became an issue because of performance of GET /v1/samples)

i mean, it's still valuable at the db level for all other resources, but we need a solution for samples, too

iamigo: we can make that a new story though since that work is not covered here

tests/api/v1/aspects/getWithLimit.js
jgraff2: ok

jgraff2:
@iamigo I'm not sure there's anything that can be done for samples in redis. More discussion is needed.

iamigo:
@shriramshankar @jgraff2 can you please take a look at the merge conflict and resolve?

PR #554, opened 2017-08-24
jobCleanup test changes to remove dependency on timing

jgraff2:
and prevent flapping tests.

### DIFF
--- a/clock/scheduledJobs/jobCleanup.js
+++ b/clock/scheduledJobs/jobCleanup.js
@@ -17,7 +17,7 @@ const kue = require('kue');
 const jobType = require('../../jobQueue/setup').jobType;
 const jobWrapper = require('../../jobQueue/jobWrapper');
 const conf = require('../../config');
-const rangeByStateAsync = Promise.promisify(kue.Job.rangeByState);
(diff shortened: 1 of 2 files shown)


clock/scheduledJobs/jobCleanup.js
```diff
@@ -47,9 +47,10 @@ function execute(batchSize, delay) {
     if (n === 0) return Promise.resolve();
     const from = skippedJobCount;
     const to = skippedJobCount + n - 1;
+    if (!now) now = Date.now();
```
iamigo: style guide violation - even simple if blocks require curly braces

jgraff2: This says it's ok: https://github.com/airbnb/javascript#blocks--braces
Those are the rules we're using, right?

iamigo: I stand corrected! Thanks for pointing that out.

jgraff2:
@iamigo do you mean the actual jobCleanup process, or my changes to the test setup?
In the actual jobCleanup process I already have comments describing each step. Are those not clear?

jgraff2:
comments added for both.

PR #557, opened 2017-08-28
on UPSERT sample in cache, return NOT_FOUND if subject is unpublished

annyhe:
- test only


annyhe:
@iamigo only test change. The test was missing previously.

PR #558, opened 2017-08-28
on PATCH samples, return NOT_FOUND if subject is unpublished

annyhe:
- api level only
- existing test in upsert samples test return NOT_FOUND if subject is unpublished

### DIFF
--- a/tests/api/v1/samples/post.js
+++ b/tests/api/v1/samples/post.js
@@ -20,18 +20,18 @@ const expect = require('chai').expect;
 const ZERO = 0;
 
 describe('tests/api/v1/samples/post.js >', () => {
+  let token;
+  before((done) => {
+    tu.createToken()
+    .then((returnedToken) => {
+      token = returnedToken;
+      done();
+    })
+    .catch(done);
+  });
+
   describe(`POST ${path} >`, () => {
     let sampleToPost;
-    let token;
-
-    before((done) => {
-      tu.createToken()
-      .then((returnedToken) => {
-        token = returnedToken;
-        done();
-      })
-      .catch(done);
-    });
 
     beforeEach((done) => {
       u.doSetup()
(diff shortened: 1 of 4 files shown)


tests/api/v1/samples/put.js
```diff
@@ -172,4 +172,42 @@ describe(`tests/api/v1/samples/put.js, PUT ${path} >`, () => {
       done();
     });
   });
+
+  describe('subject isPublished false >', () => {
+    beforeEach((done) => {
+      tu.db.Subject.findById(subjectId)
```
shriramshankar: is the "beforeEach" hook required? just a "before" hook will do.

tests/api/v1/samples/post.js
```diff
@@ -254,19 +253,35 @@ describe('tests/api/v1/samples/post.js >', () => {
...
-  describe('aspect isPublished false >', () => {
+  describe('subject isPublished false >', () => {
     let sampleToPost;
-    let token;
-
-    before((done) => {
-      tu.createToken()
-      .then((returnedToken) => {
-        token = returnedToken;
+    beforeEach((done) => {
```
shriramshankar: is the "beforeEach" hook required? just a "before" hook will do.

tests/api/v1/samples/put.js
annyhe: beforeEach is required, otherwise the "before" block to update the subject/aspect will try to run before the "beforeEach" in the outer test block, which creates the subject/aspect.
From official mocha documentation: https://mochajs.org/#hooks
  before(function() {
    // runs before all tests in this block
  });

  beforeEach(function() {
    // runs before each test in this block
  });

tests/api/v1/samples/post.js
annyhe: see comment below

PR #560, opened 2017-08-28
Encrypt sgcontext

shriramshankar:


shriramshankar:
@pallavi2209 This branch should have all of what you need for your story.  Let me know if I need to do anything else.

db/helpers/generatorUtils.js
```diff
@@ -0,0 +1,142 @@
...
+ * generator encryption algorithm from the glocalConfig model
+ * and returns an object containing the secretKey and algorithm
+ * @param {Model} globalConfigModel - The sequelize Global Config Model
+ * @returns {Object} - containing the secret key and algorithm
+ */
+function getSGEncryptionKeyAndAlgorithm(globalConfigModel) {
+  const sgObj = {};
+  return globalConfigModel.findOne({
+    where: {
+      key: dbConstants.SGEncryptionKey,
```
iamigo: why two separate db select queries? can't you do this in one db select, find where (key = x OR key = y)?

db/model/generator.js
```diff
@@ -166,16 +165,18 @@ module.exports = function generator(seq, dataTypes) {
         const gtVersion = inst.generatorTemplate.version;
         if (inst.changed('generatorTemplate') || inst.changed('context')) {
           return seq.models.GeneratorTemplate.getSemverMatch(gtName, gtVersion)
-            .then((ret) => {
-              if (!ret) {
-                throw new ValidationError('No Generator Template matches ' +
-                `name: ${gtName} and version: ${gtVersion}`);
+            .then((gt) => {
+              if (!gt) {
+                throw new ValidationError('No Generator Template matches' +
```
iamigo: always prefer space at the end of the line rather than at the beginning of the next line

db/helpers/generatorUtils.js
```diff
@@ -0,0 +1,142 @@
...
+  decrypted += decipher.final('utf-8');
+  return decrypted;
+} // decrypt
+
+/**
+ * This function fetches the sample generator secret key and the sample
+ * generator encryption algorithm from the glocalConfig model
+ * and returns an object containing the secretKey and algorithm
+ * @param {Model} globalConfigModel - The sequelize Global Config Model
+ * @returns {Object} - containing the secret key and algorithm
```
iamigo: what happens if either key does not exist? throw an error and reject creating a sample generator if its template specifies encryption for any of its context vars?

tests/db/model/generator/update.js
```diff
@@ -179,4 +192,61 @@ describe('tests/db/model/generator/update.js >', () => {
...
+          token: 'newToken',
+        },
+      }))
+      .then(() => Generator.findById(generatorDBInstance.id))
+      .then((o) => {
+        expect(o.generatorTemplate.version).to.equal('>=1.1.0');
+        expect(o.generatorTemplate.name).to.equal('newName');
+
+        /*
+         * the two asserts belowprove that the encrypted=true fields are
```
iamigo: below prove

db/helpers/generatorUtils.js
shriramshankar: I think we talked about this, and we said that we will not do the encryption if the keys do not exist.

iamigo: Oh, I forgot. Can you add comments in the code here to make it very clear?

Also, let's get this information added to https://salesforce.github.io/refocus/docs/10-security.html (separate PR OK)

db/helpers/generatorUtils.js
shriramshankar: A single query will return an array of 2 elements instead. So, we have to go through the array, do some string comparison before returning the object back vs sending 2 queries on an indexed field. Do you think we should go with the array approach?

db/helpers/generatorUtils.js
shriramshankar: Sure, will do.

pallavi2209: Is there a reason why we chose 'no encryption' over 'throwing error' ?
Basically, we are saying that the context vars will be in plain text even if they have been specified to be "encrypted" ?

db/helpers/generatorUtils.js
```diff
@@ -0,0 +1,142 @@
...
+} // encryptSGContextValues
+
+/**
+ * Decrypts the encrypted sample generator context values.
+ * @param  {GlobalConfig} globalConfigModel - The global config model
+ * @param  {Object} sg - The sample generator object
+ * @param  {Object} sgt - The sample generatot template object.
+ * @returns {Objec} - sample generator object with decrypted context values
+ */
+function decryptSGContextValues(globalConfigModel, sg, sgt) {
```
pallavi2209: Can we move this function in utils/generatorUtils because I would need this function in api/v1/controllers/collectors.js.

db/helpers/generatorUtils.js
```diff
@@ -0,0 +1,148 @@
...
+} // decrypt
+
+/**
+ * This function fetches the sample generator secret key and the sample
+ * generator encryption algorithm from the glocalConfig model
+ * and returns an object containing the secretKey and algorithm
+ * @param {Model} globalConfigModel - The sequelize Global Config Model
+ * @returns {Object} - containing the secret key and algorithm
+ */
+function getSGEncryptionKeyAndAlgorithm(globalConfigModel) {
```
pallavi2209: I am not sure why are we passing model around. Is it not safe to assume that key/algo will always be in globalConfigModel?

db/helpers/generatorUtils.js
shriramshankar: If there no secret key and algorithm in the global config, the context variables will be stored in plain text.

db/helpers/generatorUtils.js
shriramshankar: Because the key and algorithm rows are added by the Admin, it is not safe to assume that these rows are always going to be there.

pallavi2209: Right, but the model will always be there, correct?

db/helpers/generatorUtils.js
iamigo: i suspect manipulating an array of two elements is cheaper than two roundtrip db queries, but we're talking about such small numbers it doesn't really matter

db/helpers/generatorUtils.js
pallavi2209: umm okay. I guess as a user I will not be happy if my secured variables are stored as plain text because of admin's decision. But I am assuming you have discussed it, so its the best solution as of now.

iamigo: thx for your comment @pallavi2209 I think we should reconsider that
@shriramshankar can you ask Iain and Charlie what they think?

db/helpers/generatorUtils.js
shriramshankar: > its just that we will have to import and pass model (eg in api controller) whenever we wanna use these functions.
utility/helper files should more or less have no dependency on other parts of the code. You already have access to the models in the api layer in api/v1/helpers/nouns

db/helpers/generatorUtils.js
shriramshankar: Moving just this might not really help. I was trying to make this as general purpose as possible. How many of the function exported in the module will you be using?

pallavi2209: I would need decryptSGContextValues and encrypt functions. I can just import these then. What do you think?

db/helpers/generatorUtils.js
shriramshankar: This is more or less like the "require https" feature we have. Yes, security is really important and we should have everything in place for a secure installation. But it should be made configurable and left to the Admin to decide how and what level of security is needed for their installation.

db/helpers/generatorUtils.js
shriramshankar: What about moving all of under utils and calling them encryptUtils or encryptDecryptUtils?

db/helpers/generatorUtils.js
shriramshankar: @meynet @iusiri  What are your thoughts about this? 
When a context variable is required to be encrypted but the Admin has not yet provided the secretKey and algorithm in the global config table. Should the post request, to add the generator to refocus be rejected?

db/helpers/generatorUtils.js
pallavi2209: Sounds good to me.

db/helpers/generatorUtils.js
shriramshankar: Iain thinks throwing an error is better. Let's see what Charlie says and we can do what most of us think is right to do.

pallavi2209: I am fine with either case after you have mentioned "require https" explanation. If we do not get his reply soon, we can probably merge this (I am waiting on this one) and create a new PR if we need to do any further change.

shriramshankar:
@iamigo @pallavi2209 This PR has been updated, with all of what was discussed in the comments, expect the decision on what needs to be done when the secret key and algorithm is not present in the GlobalConfig. Once Charlie shares his thoughts on this, we can decide the next course of action probably in the next PR.

tests/utils/crypt.js
```diff
@@ -0,0 +1,207 @@
...
+          secretInformation,
+          otherNonSecretInformation,
+        },
+      };
+      generatorUtils.encryptSGContextValues(GlobalConfig,
+        looksLikeSG, looksLikeSGT)
+      .then((sg) => {
+        expect(sg).to.not.equal(undefined);
+        expect(sg.context.otherNonSecretInformation)
+          .equal(otherNonSecretInformation);
```
pallavi2209: Why is otherNonSecretInformation equal to otherNonSecretInformation after encryption? It should be some random string, right?

pallavi2209: expect(sg.context.secretInformation).to.not.equal(secretInformation); please.

shriramshankar: if they are equal the decryption will fail. line 107 proves that they were not equal.

pallavi2209: I was saying *not.equal*, but nevermind, I tested this on local and it works. I can add it in my PR. I am fine merging the PR. Thanks!!

shriramshankar: Yes, "not.equal". Saying something is not equal to something does not help much and in this case, it does not tell you that the value was encrypted. 
According to the "chai expects"  library it is not a recommended way 
`expect(2).to.not.equal(1); // Not recommended` 
We have done in in quite a few places in the tests but it is something we should avoid. 
I tend to use not.equals usually before referencing the object's properties.

pallavi2209: Right. I absolutely agree. We should NOT prefer "not.equal" over "equal", but in cases we cannot test "equal", then I guess testing atleast "not.equal" is better. 
Like in this scenario, lets say encrypt/decrypt funcs are not doing anything and returning the object as it is, then the test will still pass.

shriramshankar: There are other tests that test the encrypt/decrypt functions, those tests will fail.

pallavi2209: I just commented out the encrypt and decrypt functions and return just data. All the tests pass in tests/utils/crypt.js

shriramshankar: if you returned a random string instead of the data then having the not.equal test would pass. I do not understand what you are trying to prove here.

pallavi2209: Nevermind, nothing to prove, just a remote discussion on whether adding another line in test would make it better or not. Hangouts would have been better :)

iamigo:
@shriramshankar please create new user story to reject SG post/patch/put if its SGT has any context vars flagged to be encrypted AND the Refocus instance does not have the encryption key and algorithm specified.
Error message should be something like "Unable to save this Sample Generator with encrypted context data. Please contact your Refocus administrator to set up the encryption algorithm and key to protect any sensitive information you may include in your Sample Generator's context."

shriramshankar:
@iamigo I was going to make that change as a part of my next PR, once the decision was made. Do you still need a story?

shriramshankar:
@iamigo Can you merge this, if there is nothing that needs to change apart from the error thing?

PR #565, opened 2017-09-01
Get generator should have collector field

annyhe:


tests/api/v1/generators/getWithCollector.js
```diff
@@ -0,0 +1,122 @@
...
+const Generator = tu.db.Generator;
+const GeneratorTemplate = tu.db.GeneratorTemplate;
+const path = '/v1/generators';
+const expect = require('chai').expect;
+const ZERO = 0;
+const ONE = 1;
+const TWO = 2;
+const THREE = 3;
+
+describe('tests/api/v1/generators/get.js >', () => {
```
shriramshankar: getWithCollector.js

api/v1/swagger.yaml
```diff
@@ -11245,6 +11245,11 @@ definitions:
         type: string
         readOnly: true
         description: The query to append to GET subjects.
+      collectors:
+        description: Created collector resources.
```
iamigo: description: List of collectors this Sample Generator should try to use.

tests/db/model/generator/withCollectors.js
```diff
@@ -0,0 +1,130 @@
...
+
+  describe('with one collector', () => {
+    let collector;
+    let relodedGenerator;
+
+    before((done) => {
+      tu.db.Collector.create({ name: 'snow' })
+      .then((_collector) => {
+        collector = _collector;
+        return createdGenerator.addCollectors([collector]);
```
annyhe: @iamigo @shriramshankar given that collectors are added onto generators, shall we still POST with a collector?
if yes, the collectors would be added onto generators in the db/models, and the validation for whether the collector exists and is registered would be in the generator controller.

annyhe: Updated PR with the changes you mentioned @shriramshankar @iamigo

iamigo: yes, we want to post a new generator with a list of collectors
if there is no match for any of the collectors specified in the list, then reject the POST
i don't care if the check is made in the api layer or the db layer, but if it's in the db layer it would need to be in a beforeCreate hook

shriramshankar:
@annyhe this is only for GET, correct? Are we doing post later?

annyhe:
@shriramshankar Yes this is only for the GET. POST will follow in another PR

PR #568, opened 2017-09-05
Load Rooms List and Grid

tausifmuzaffar:
Make a first page for rooms


view/perspective/app.js
```diff
@@ -95,7 +96,7 @@ function handleError(err) {
   }
 
   ERROR_INFO_DIV.innerHTML = msg;
-  removeSpinner();
+  u.removeSpinner(SPINNER_ID);
```
iamigo: good

iamigo:
ready to approve after styling fixes (and if you were going to move the getPromiseWithUrl out to utils)

pallavi2209:
The code looks fine to me. Consider it approved once Ian's comments are taken care of.

tausifmuzaffar:
@kamilsmuga waiting on your approval

view/rooms/index.pug
```diff
@@ -0,0 +1,35 @@
...
+  For full license text, see LICENSE.txt file in the repo root or
+  https://opensource.org/licenses/BSD-3-Clause
+
+doctype html
+html
+  head
+    title Refocus Room
+    link(rel='shortcut icon', href='favicon.ico')
+    link(rel='stylesheet', type='text/css', href='../static/css/salesforce-lightning-design-system.min.css')
+    script(src='http://code.jquery.com/jquery-1.11.1.js')
```
kamilsmuga: load only https resources - this will fail the page to load when hosted on SSL enabled servers

kamilsmuga: also - can you remind me why do we need jquery?

view/rooms/list/index.pug
```diff
@@ -0,0 +1,21 @@
...
+  https://opensource.org/licenses/BSD-3-Clause
+
+doctype html
+html
+  head
+    title Refocus Room
+    meta(name='viewport', content='width=device-width, initial-scale=1.0')
+    link(rel='shortcut icon', href='favicon.ico')
+    link(rel='stylesheet', type='text/css', href='../../static/css/salesforce-lightning-design-system.min.css')
+    script(src='http://code.jquery.com/jquery-1.11.1.js')
```
kamilsmuga: use https instead

view/rooms/index.pug
iamigo: I thought that the SSL only restriction would only apply to incoming requests.
Also, if jquery is required, maybe better to do the require in the code and let webpack bundle it up?

kamilsmuga: I can't prove it easily (lack of ssl-enabled dev env), but I'm 99% it would fail to load.

See: "An HTTPS page that includes content fetched using cleartext HTTP is called a mixed content page" [..] "Pages like this are only partially encrypted" [..] "That leaves the pages unsafe." [...] "Starting in Firefox 23, mixed active content is blocked by default" 
from https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content

tausifmuzaffar: @kamilsmuga ive updated the resources to use the secure version; the bdk and bots use/may use jquery respectively

view/rooms/index.pug
```diff
@@ -11,7 +11,7 @@ html
     title Refocus Room
     link(rel='shortcut icon', href='favicon.ico')
     link(rel='stylesheet', type='text/css', href='../static/css/salesforce-lightning-design-system.min.css')
-    script(src='http://code.jquery.com/jquery-1.11.1.js')
+    script(src='https://code.jquery.com/jquery-1.11.1.js')
```
iamigo: if loading remote resources (and why loading remote resources?) should loading the minified versions of them

tausifmuzaffar:
@iamigo @kamilsmuga I downloaded the scripts locally and moved them to a public/scripts folder; I moved the sockets script there too.

PR #572, opened 2017-09-06
Post generator response should have collectors field

annyhe:
on post generator:
       * 1. validate the collectors field: if include non-existent collector, abort the POST operation with 404 error
       * 2. create the generator
       * 3. add collectors (if any)


api/v1/helpers/verbs/postUtils.js
```diff
@@ -54,6 +54,10 @@ function makePostPromise(params, props, req) {
   } else {
 
     // cache is off and returnUser is false.
+    if (props.modelName === 'Generator') {
+      return props.model.createWithCollectors(toPost);
```
iamigo: we should decompose the "makePostPromise" function so that we can call the *common* code from *every* model without having a bunch of "if model is Generator/Sample/etc." statements in it, and then put the Generator- and Sample-specific stuff in their own controllers

(that could involve changes to the rest of the controllers and the way they call the common POST functions, so if this should be a separate story, we can leave this here for now and tackle that in another sprint)

db/model/generator.js
```diff
@@ -154,6 +156,45 @@ module.exports = function generator(seq, dataTypes) {
...
+
+      /**
+       * 1. validate the collectors field: if succeed, save the collectors in temo var for
+       *  attaching to generator. if fail, abort the POST operation
+       * 2. create the generator
+       * 3. add the saved collectors (if any)
+       */
+      createWithCollectors(requestBody, user) {
+        const options = {};
+        let collectors; // will be populate with actual collectors
```
iamigo: did you mean "populated"?

db/model/generator.js
```diff
@@ -154,6 +156,45 @@ module.exports = function generator(seq, dataTypes) {
...
+        options.where = utils.whereClauseForNameInArr(requestBody.collectors || []);
+        return new seq.Promise((resolve, reject) =>
+          seq.models.Collector.findAll(options)
+          .then((_collectors) => {
+
+            /**
```
iamigo: single asterisk opening this comment block (double asterisks are for module-level and function-level comments)

db/model/generator.js
```diff
@@ -154,6 +156,45 @@ module.exports = function generator(seq, dataTypes) {
...
+
+            const err = new dbErrors.ResourceNotFoundError();
+            err.resourceType = 'Collector';
+            err.resourceKey = requestBody.collectors;
+            throw err;
+          }) // if successful create, add collectors
+          .then((createdGenerator) => createdGenerator.addCollectors(collectors))
+          .then(() => Generator.findOne({ where: { name: requestBody.name } }))
+          .then((findresult) => resolve(findresult.reload()))
+          .catch((err) => reject(err))
```
iamigo: .catch(reject)

tests/api/v1/generators/postWithCollector.js
```diff
@@ -0,0 +1,121 @@
...
+
+      expect(res.body.collectors.length).to.equal(THREE);
+      expect(res.body.collectors[ZERO].name).to.equal(collector1.name);
+      expect(res.body.collectors[ONE].name).to.equal(collector2.name);
+      expect(res.body.collectors[TWO].name).to.equal(collector3.name);
+      done();
+    });
+  });
+
+  it('404 error for request body with an non-exitent collector',
```
iamigo: non-existant

tests/api/v1/generators/postWithCollector.js
```diff
@@ -0,0 +1,121 @@
...
+        return done(err);
+      }
+
+      expect(res.body.errors[0].type).to.equal('ResourceNotFoundError');
+      expect(res.body.errors[0].source).to.equal('Generator');
+      done();
+    });
+  });
+
+  it('404 error for request body with an existing and a non-exitent collector',
```
iamigo: non-existant

tests/db/model/generator/createWithCollectors.js
```diff
@@ -0,0 +1,129 @@
...
+      expect(o.generatorTemplate.name).to.equal('refocus-ok-template');
+      expect(o.generatorTemplate.version).to.equal('1.0.0');
+      expect(typeof o.getWriters).to.equal('function');
+      expect(typeof o.getCollectors).to.equal('function');
+      done();
+    })
+    .catch(done);
+  });
+
+  it('404 error for request body with an non-exitent collector', (done) => {
```
iamigo: non-existant

tests/db/model/generator/createWithCollectors.js
```diff
@@ -0,0 +1,129 @@
...
+    .catch((err) => {
+      expect(err.status).to.equal(NOT_FOUND_STATUS_CODE);
+      expect(err.name).to.equal('ResourceNotFoundError');
+      expect(err.resourceType).to.equal('Collector');
+      expect(err.resourceKey).to.deep.equal(localGenerator.collectors);
+      done();
+    });
+  });
+
+  it('404 error for request body with an existing and a ' +
```
iamigo: non-existant

db/utils.js
```diff
@@ -365,6 +413,8 @@ if (!Array.prototype.includes) {
 }
 
 module.exports = {
+  sortArrayAccordingToAnotherArray,
+  whereClauseForNameInArr,
```
shriramshankar: there is a db/helpers/common.js file for these kind of this. Can you move it there?

db/model/generator.js
annyhe: sorry yes I meant that. Will push a different commit.

api/v1/helpers/verbs/postUtils.js
annyhe: ok story created.

db/utils.js
annyhe: ok will do

tests/db/model/generator/createWithCollectors.js
annyhe: yep changed in a new commit

annyhe: yep changed accordingly

annyhe:
Updated as per your comments. Please review @shriramshankar @iamigo

api/v1/helpers/verbs/postUtils.js
shriramshankar: For this PR,  are we leaving it this way or are we moving the createWithCollector to the controller?

db/helpers/common.js
```diff
@@ -32,6 +32,54 @@ const changeType = {
...
+ * order of the objects, return a copy of the first array ordered
+ * by the second array. This sort is NOT in-place.
+ * ELSE return empty array.
+ *
+ * @param {Array} objArray Array of objects, with key 'name'
+ * @param {Array} keyArray Array of strings specifying the order of names
+ * @returns {Array} objArray with key name sorted according to keyArray,
+ * or [] for err cases
+ */
+function sortArrayAccordingToAnotherArray(objArray, keyArray) {
```
shriramshankar: why are we do this sort?

db/helpers/common.js
```diff
@@ -32,6 +32,54 @@ const changeType = {
 };
 
 /**
+ * Replicated here from api/v1/helpers/verbs/utils.js
+ * Return the sequelize query { name: { $in: arr } }
+ * @param {Array} arr The array to return in the query.
+ * @returns {Object} the query object
+ */
+function whereClauseForNameInArr(arr) {
```
shriramshankar: if we will be calling the generator models class method from the controller, we might not have to replicate this. We can use the one in the api layer and pass it as another parameter to the custom create method. Corret?

db/model/generator.js
```diff
@@ -154,6 +155,45 @@ module.exports = function generator(seq, dataTypes) {
...
         });
       },
+
+      /**
+       * 1. validate the collectors field: if succeed, save the collectors in temo var for
+       *  attaching to generator. if fail, abort the POST operation
+       * 2. create the generator
+       * 3. add the saved collectors (if any)
+       */
+      createWithCollectors(requestBody, user) {
```
shriramshankar: looks like "user" is not used.

shriramshankar: The function definition is also missing.

tests/db/helpers/common.js
```diff
@@ -20,6 +20,40 @@ const common = require('../../../db/helpers/common');
 
 describe('tests/db/helpers/common.js >', () => {
   after(u.forceDelete);
+
+  describe('tests/db/helpers/common.js, sortArrayAccordingToAnotherArray >', () => {
```
shriramshankar: I think we want the file location only in the top most describe.

db/helpers/common.js
```diff
@@ -32,6 +32,54 @@ const changeType = {
...
+ * or [] for err cases
+ */
+function sortArrayAccordingToAnotherArray(objArray, keyArray) {
+  if (!objArray || !keyArray ||
+    (objArray.length != keyArray.length)) {
+    return [];
+  }
+
+  // Create a temporary hash table to store the objects
+  let tempObj = {};
```
shriramshankar: use const instead.

db/model/generator.js
```diff
@@ -154,6 +155,45 @@ module.exports = function generator(seq, dataTypes) {
...
+          .then((_collectors) => {
+
+            /*
+             * If requestBody does not have a collectors field, OR
+             * if the number of collectors in requestBody MATCH the
+             * GET result, order the collectors AND create the generator.
+             * Else throw error since there are collectors that don't exist.
+             */
+            if (!requestBody.collectors ||
+              (_collectors.length === requestBody.collectors.length)) {
```
shriramshankar: what do we do when "requestBody.collectors" contains duplicate collector names?

api/v1/helpers/verbs/postUtils.js
annyhe: we are leaving it this way for now. This PR is already getting large. I created a story for the moving of all model-related code in makePostPromise

db/helpers/common.js
annyhe: because findAll returns results in a non-orderly fashion. Sequelize can sort by ascending or descending, but that will be very complex as the order of collectors is not alphabetical.
We can try squeezing the ordering logic into the findAll(options), but that is more complex and less testable than the function here.

db/helpers/common.js
annyhe: I tried requiring the file in the api layer, but there was some error on runtime. Also wouldn't it introduce unnecessary dependencies if we require a file from the api when we are in db? This function is very small.

db/model/generator.js
annyhe: good point. will remove in another commit

annyhe: Did you mean the @param and @returns? added in new commit

tests/db/helpers/common.js
annyhe: ok taking it out

db/helpers/common.js
annyhe: ok changed to const

db/model/generator.js
annyhe: adding a test for that

tests/db/model/generator/createWithCollectors.js
```diff
@@ -52,6 +52,8 @@ describe('tests/db/model/generator/createWithCollectors.js >', () => {
     .catch(done);
   });
 
+  // delete generator after each test
+  afterEach(() => tu.forceDelete(tu.db.Generator, new Date()));
```
iamigo: Are you sure you want to be sending new date here? I think that will delete generators created *after* that date.

iamigo: @annyhe

db/model/generator.js
annyhe: ok added the test to show the result of POSTing with duplicate collectors is 400

annyhe:

db/helpers/common.js
```diff
@@ -32,6 +32,28 @@ const changeType = {
...
 
 /**
+ * Returns duplicate strings in an given flat array
+ *
+ * ie. [a, b, c, c, a] => [a, c]
+ *
+ * @param {Array} arr Array of strings
+ * @returns {Array} Array of duplicate strings
+ */
+function findDuplicatesInStringArray(arr) {
```
shriramshankar: Wondering if we can do something like, adding the elements of the array to the set and check if the length of the set and the array are same and return true or false based on this.?

annyhe:
@shriramshankar Updated as per your comment. Please review

tests/db/helpers/common.js
```diff
@@ -20,6 +20,35 @@ const common = require('../../../db/helpers/common');
 
 describe('tests/db/helpers/common.js >', () => {
   after(u.forceDelete);
+
+  describe('checkDuplicatesInStringArray >', () => {
+    it('empty input returns false', () => {
+      expect(common.checkDuplicatesInStringArray()).to.be.false;
+    });
+
+    it('false returns empty array', () => {
```
annyhe: needs to say false

db/helpers/common.js
```diff
@@ -32,6 +32,26 @@ const changeType = {
...
+ *
+ * @param {Array} arr Array of strings
+ * @returns {Boolean} Does the input array contain duplicates
+ */
+function checkDuplicatesInStringArray(arr) {
+  if (!arr || !arr.length) {
+    return false;
+  }
+
+  // if all the elements are distinct,
```
annyhe: multi line comment

PR #577, opened 2017-09-08
Remove UI blob from GET /bots response

tausifmuzaffar:
W-4269584

The response was too large because every UI was loaded each time, so we just make it respond with size of the UI for the find


api/v1/helpers/verbs/doFind.js
```diff
@@ -45,11 +45,17 @@ function doFindAndCountAll(reqResNext, props, opts) {
   .then((o) => {
     resultObj.dbTime = new Date() - resultObj.reqStartTime;
     reqResNext.res.set(COUNT_HEADER_NAME, o.count);
+
     const retval = o.rows.map((row) => {
       if (props.modelName === 'Lens') {
         delete row.dataValues.library;
       }
-
+      if ((props.modelName === 'Bots') &&
```
iamigo: On a find, I'd like to see you try to do this in a way which does *not* load all bot library data from the db *then* delete it. That's too expensive.

tausifmuzaffar:
@iamigo I decided remove the ui from the default scope and create UI scope

I created a new "getScopes" and edited doGet to look for the scope. If you dont think I should modify "doGet" like that i can do it in the controller

db/model/bot.js
```diff
@@ -104,6 +104,18 @@ module.exports = function bot(seq, dataTypes) {
           foreignKey: 'botId',
           through: 'RoomTypeBots',
         });
+
+        Bot.addScope('botUI', {
+          attributes: { include: ['ui'] },
+        }, {
+          override: true,
```
iamigo: I think "override" is only necessary for the "defaultScope" ... are you sure about this here?

tausifmuzaffar:
@kamilsmuga @pallavi2209 I think the branch is ready to be merge unless you have more comments

kamilsmuga:
[LGTM](http://www.lgtm.in/i/8OVYoSSTp)

PR #580, opened 2017-09-11
Throw error when no key/algo in globalconfig when decrypting SG context values

pallavi2209:


api/v1/apiErrors.js
```diff
@@ -128,6 +128,18 @@ apiErrors.create({
...
 
+apiErrors.create({
+  code: 11111,
+  status: 400,
+  name: 'SampleGeneratorContextDecryptionError',
+  parent: apiErrors.ValidationError,
+  fields: [],
+  defaultMessage: 'Unable to decrypt the Sample Generator context data. ' +
+  'Please contact your Refocus administrator to set up the encryption ' +
+  'algorithm and key which was used to protect the sensitive information ' +
```
iamigo: i think get rid of the words "which was used"

pallavi2209:
@iamigo updated the PR

PR #581, opened 2017-09-12
Enforce Profile Access For Write Protected Resources

kriscfoster:
@iamigo Wanted to check that you're okay with the way I'm doing this for RoomTypes before I enforce it in other controllers. This will mean that if someone is making a write request, it will only be allowed if that users profile has 'rw' access to that resource. (Means that an access token will always be required for write operations (incl POST, PATCH, PUT, DELETE) to that resource)

### DIFF
--- a/db/model/profile.js
+++ b/db/model/profile.js
@@ -146,7 +146,7 @@ module.exports = function profile(seq, dataTypes) {
         return new Promise((resolve, reject) => {
           Profile.findById(profileId)
           .then((p) => resolve(p &&
-            p[accessModel] === 'rw'.toLowerCase()))
+            p[accessModel] === 'rw'))
           .catch(reject);
         });
       }, // hasWriteAccess
--- a/db/model/profile.js
+++ b/db/model/profile.js
@@ -146,7 +146,7 @@ module.exports = function profile(seq, dataTypes) {
         return new Promise((resolve, reject) => {
           Profile.findById(profileId)
           .then((p) => resolve(p &&
-            p[accessModel] === 'rw'))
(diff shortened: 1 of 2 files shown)


db/model/profile.js
```diff
@@ -142,7 +142,7 @@ module.exports = function profile(seq, dataTypes) {
       }, // isAdmin
 
       hasWriteAccess(profileId, model) {
-        const accessModel = model.getAccessField();
+        const accessModel = model.getProfileAccessField();
```
iamigo: ``` === 'rw'.toLowerCase()```

This doesn't make sense--we're lowercasing the string "rw"? Is it the *other* side of the equation which is supposed to be lowercased?

iamigo:
Yes, this is fine (except my one comment about lower case rw).

(I still would prefer to see all this access stuff happen in middleware, but let's save that for another day.)

kriscfoster:
@iamigo Going to try to do this as middleware again, would be cleaner and a lot less tedious

db/model/profile.js
```diff
@@ -146,7 +146,7 @@ module.exports = function profile(seq, dataTypes) {
         return new Promise((resolve, reject) => {
           Profile.findById(profileId)
           .then((p) => resolve(p &&
-            p[accessModel] === 'rw'.toLowerCase()))
+            p[accessModel] === 'rw'))
```
iamigo: lines 148-149 fit on one line now so pls. get rid of the line break

iamigo:
> do this as middleware

Yes!!

PR #582, opened 2017-09-13
Add columns to collector model with migration and tests

pallavi2209:
Only migration and models changes with tests. Will do another PR for swagger and API changes.

### DIFF
--- a/migrations/20170911221051-add-osInfo-processInfo-version-to-collector.js
+++ b/migrations/20170911221051-add-osInfo-processInfo-version-to-collector.js
@@ -14,6 +14,7 @@ module.exports = {
       }))
       .then(() => qi.addColumn(TBL, 'version', {
         type: Sequelize.STRING,
+        defaultValue: '0.0.0',
         allowNull: false,
       })));
   },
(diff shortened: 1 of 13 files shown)


migrations/20170911221051-add-osInfo-processInfo-version-to-collector.js
```diff
@@ -0,0 +1,27 @@
...
+        type: Sequelize.JSONB,
+        allowNull: true,
+      })
+      .then(() => qi.addColumn(TBL, 'processInfo', {
+        type: Sequelize.JSONB,
+        allowNull: true,
+      }))
+      .then(() => qi.addColumn(TBL, 'version', {
+        type: Sequelize.STRING,
+        allowNull: false,
```
iamigo: does migration work if adding a column with `allowNull: false` if there are already records in the db? might have to provide default value (if so, maybe just set default value to `0.0.0` to imply we don't have a version yet)?

shriramshankar: We had the default version set as 1.0.0 for SGT. Wondering if we want to do the same for the collectors too.?

iamigo: default ver for SGT is different ... here we're just making sure nothing breaks if a record existed before this column was added

pallavi2209:

PR #584, opened 2017-09-13
Bot Settings Migration

tausifmuzaffar:
The bots route settings is missing so created this migration to add the column


tausifmuzaffar:
It was in the last release

PR #589, opened 2017-09-15
Collector: add new fields in swagger and check for token in PATCH api

pallavi2209:


api/v1/controllers/collectors.js
```diff
@@ -138,13 +137,31 @@ function getCollector(req, res, next) {
...
+ * a collector.
  *
  * @param {IncomingMessage} req - The request object
  * @param {ServerResponse} res - The response object
  * @param {Function} next - The next middleware function in the stack
  */
 function patchCollector(req, res, next) {
-  doPatch(req, res, next, helper);
+  let verifyCtrToken = false;
+  for (let i = 0; i < fieldsWritableByCollectorOnly.length; i++) {
```
iamigo: maybe something like https://www.npmjs.com/package/just-intersect ...
e.g. 

```
verifyCtrTkn = intersect(Object.keys(req.body), fieldsWritableByCollectorOnly).length > 0;
```

api/v1/controllers/collectors.js
```diff
@@ -34,6 +31,8 @@ const config = require('../../../config');
 const encryptionAlgoForCollector = config.encryptionAlgoForCollector;
 const ZERO = 0;
 
+const fieldsWritableByCollectorOnly = ['osInfo', 'processInfo', 'version'];
```
shriramshankar: can we move this to the nouns file?

api/v1/controllers/collectors.js
```diff
@@ -138,13 +137,31 @@ function getCollector(req, res, next) {
...
+    }
+  }
+
+  if (verifyCtrToken) { // verify that token belongs to collector
+    return jwtUtil.verifyCollectorToken(req)
+    .then(() => doPatch(req, res, next, helper))
+    .catch((err) => u.handleError(next, err, helper.modelName));
+  } else {
+    return doPatch(req, res, next, helper);
+  }
```
shriramshankar: the else block here is not necessary.

iamigo: More specifically, the "else" is not necessary, but what's inside the else is still necessary.

api/v1/controllers/collectors.js
pallavi2209: How about this: 
`  // verify controller token if atleast one field is writable by collector
  let verifyCtrToken = false;
  const reqBodyKeys = Object.keys(req.body);
  for (let i = 0; i < fieldsWritableByCollectorOnly.length; i++) {
    const fieldName = fieldsWritableByCollectorOnly[i];
    if (reqBodyKeys.indexOf(fieldName) > -1) {
      verifyCtrToken = true;
      break;
    }
  }`

Same complexity and less dependency on third party package. What do you think?

pallavi2209:
@shriramshankar @iamigo updated the PR. @iamigo I added a comment for intersect. Let me know if you still think we should use the package and I will make the changes.

PR #591, opened 2017-09-15
Watch stuck jobs

shriramshankar:


shriramshankar:
The watch stuckJobs won't run until we set the WATCH_STUCK_JOBS_INTERVAL_MILLI_SECONDS config var.

config.js
```diff
@@ -74,6 +73,8 @@ const replicaConfigLabel = 'REPLICAS';
 // an array of read-only data base URLs
 const readReplicas = configUtil.getReadReplicas(pe, replicaConfigLabel);
 
+const WATCH_STUCK_JOBS_INTERVAL_MILLI_SECONDS = Number.MAX_SAFE_INTEGER ||
```
jgraff2: isn't this backwards?

shriramshankar: :) . Will fix it

config.js
```diff
@@ -74,6 +73,8 @@ const replicaConfigLabel = 'REPLICAS';
 // an array of read-only data base URLs
 const readReplicas = configUtil.getReadReplicas(pe, replicaConfigLabel);
 
+const WATCH_STUCK_JOBS_INTERVAL_MILLI_SECONDS =
```
iamigo: "milliseconds" is itself a real word so does not take a `_` in the middle

PR #592, opened 2017-09-15
fix the samples surviving subject delete issue

annyhe:
- tested locally. Resolves the issue.
testing methodology:
- run server with cache turned on: `REDIS_SAMPLE_STORE=true node .`
- create published subject root and root.child, 1 published aspect
- create samples out of the subjects and aspects
- GET /v1/samples to make sure the samples are there
- DELETE child subject
- GET /v1/samples should show only one sample, the sample that is attached to root
- DELETE root subject
- GET /v1/samples should show []

### DIFF
--- a/db/helpers/subjectUtils.js
+++ b/db/helpers/subjectUtils.js
@@ -110,11 +110,14 @@ function throwNotMatchError(parentId, parentAbsolutePath) {
  * redis sample store.
  *
  * @param {String} absolutePath - The absolutePath of the subject
+ * @returns {Promise}
  */
 function removeFromRedis(absolutePath) {
-  redisOps.deleteKey(subjectType, absolutePath);
-  redisOps.deleteKeys(sampleType, subjectType, absolutePath);
-  redisOps.deleteKey(subAspMapType, absolutePath);
+  return Promise.all([
+    redisOps.deleteKey(subjectType, absolutePath),
+    redisOps.deleteKeys(sampleType, subjectType, absolutePath),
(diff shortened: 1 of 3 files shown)


db/helpers/subjectUtils.js
```diff
@@ -110,11 +110,14 @@ function throwNotMatchError(parentId, parentAbsolutePath) {
  * redis sample store.
  *
  * @param {String} absolutePath - The absolutePath of the subject
+ * @returns {Promise}
  */
 function removeFromRedis(absolutePath) {
-  redisOps.deleteKey(subjectType, absolutePath);
-  redisOps.deleteKeys(sampleType, subjectType, absolutePath);
-  redisOps.deleteKey(subAspMapType, absolutePath);
+  return Promise.all([
```
annyhe: redisOps.deleteKey returns a Promise

tests/cache/models/subjects/subjectCRUD.js
```diff
@@ -281,24 +282,27 @@ describe('tests/cache/models/subjects/subjectCRUD.js >', () => {
     .catch(done);
   });
 
+  it('removeFromRedis removes all the related samples ' +
```
annyhe: not sure if this file is the best place to test subjectUtils, but subjectUtils need to be tested. 
perhaps we can move the new test to a different file, once we get enough tests on subjectUtils.

iamigo: one test is enough for a new file :)

PR #597, opened 2017-09-18
Generify do patch

annyhe:


annyhe:
@shriramshankar please review

api/v1/helpers/verbs/patchUtils.js
```diff
@@ -0,0 +1,59 @@
...
+
+const featureToggles = require('feature-toggles');
+const u = require('./utils');
+const publisher = u.publisher;
+const event = u.realtimeEvents;
+const httpStatus = require('../../constants').httpStatus;
+const constants = require('../../../../cache/sampleStore').constants;
+const redisModelSample = require('../../../../cache/models/samples');
+const helper = require('../nouns/perspectives');
+const redisCache = require('../../../../cache/redisCache').client.cache;
```
shriramshankar: few this required here are not actually used in this file. So can we please remove them?

api/v1/helpers/verbs/patchUtils.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Sends the udpated record back in the json response
+ * with status code 200.
+ *
+ * @param {Object} resultObj For logging
+ * @param {Object} req From Express
+ * @param {Object} retVal The updated instance
+ * @param {Object} props From db model
```
shriramshankar: We need to say what these objects are, rather than just saying  from db or from express. Can you look at the other files in the verb folder and correct it. ?

api/v1/helpers/verbs/doPatch.js
```diff
@@ -13,6 +13,7 @@
 
 const featureToggles = require('feature-toggles');
 const u = require('./utils');
+const utils = require('./patchUtils');
```
shriramshankar: we have many utils fields created, can the name be more specific ?

api/v1/helpers/verbs/patchUtils.js
annyhe: yep removing them now.

api/v1/helpers/verbs/patchUtils.js
annyhe: yep updating the fields to be more specific.

PR #599, opened 2017-09-18
added error handling for not valid sample name

harshkothari410:


realtime/utils.js
```diff
@@ -400,7 +400,16 @@ function isIpWhitelisted(addr, whitelist) {
...
 function attachAspectSubject(sample, useSampleStore, subjectModel,
   aspectModel) {
-  const nameParts = sample.name.split('|');
+  let nameParts;
+
+  // if name has not '|' then throw error
+  try {
+    nameParts = sample.name.split('|');
+  } catch (err) {
+    return Promise.resolve({ isFailed: true,
```
iamigo: Where does this Promise.resolve end up? ... logged?
If this is an unexpected and undesired state, shouldn't it be a Promise.reject and get handled and logged somewhere?

harshkothari410:
I and Shriram talked about this thing and came up with this for logging bug. @iamigo @shriramshankar

realtime/utils.js
```diff
@@ -400,7 +401,15 @@ function isIpWhitelisted(addr, whitelist) {
  */
 function attachAspectSubject(sample, useSampleStore, subjectModel,
   aspectModel) {
-  const nameParts = sample.name.split('|');
+  let nameParts;
```
shriramshankar: Can we do something like 
if (!sample.name) {
 log error and return undefined 
}

and in "redisPublisher.js" check for "falsey" in line 134 before calling publishObject ?

harshkothari410: I mean we can do but this is just for debugging that's why. If this is the final solution then we can use this.

harshkothari410: Let me know what do you think? @shriramshankar

realtime/utils.js
```diff
@@ -400,7 +401,15 @@ function isIpWhitelisted(addr, whitelist) {
...
+
+  // check if sample object contains name
+  if (sample.name) {
+    nameParts = sample.name.split('|');
+  } else {
+    logger.error('sample object does not contain name', sample);
+  }
```
iamigo: we should not attempt to continue here if sample is falsey OR sample does contain "name" attribute OR sample.name does not contain "|" character, AND we should go back to caller and try to figure out under what circumstances we're passing an invalid sample through (more logging there?)

realtime/utils.js
shriramshankar: I think Ian's comments kind of answers your question.

harshkothari410:
Ok working on it @shriramshankar @iamigo

realtime/redisPublisher.js
```diff
@@ -129,11 +129,14 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
   const eventType = event || getSampleEventType(sampleInst);
   const useSampleStore =
     featureToggles.isFeatureEnabled('enableRedisSampleStore');
+
   return rtUtils.attachAspectSubject(sampleInst, useSampleStore, subjectModel,
     aspectModel)
   .then((sample) => {
-    publishObject(sample, eventType);
-    return sample;
+    if (sample) {
```
iamigo: This doesn't help shed any light on what circumstances would lead to publishSample getting called with an empty or invalid sampleInst argument. Please go deeper, go upstream, we need to understand how bad data is getting through this far.

harshkothari410: I did this https://github.com/salesforce/refocus/pull/599/commits/acfbd5dab37d4089dc70f851ca19881c0d36eb10 but @shriramshankar suggested these changes that's why I reverted back. @iamigo So what should we need now.

PR #601, opened 2017-09-19
Add catch block in worker processes to call done(err)

iamigo:
for proper kue management


worker/jobs/getHierarchyJob.js
```diff
@@ -9,14 +9,14 @@
 /**
  * /worker/jobs/getHierarchyJob.js
  */
+const logger = require('winston');
```
shriramshankar: Wondering if you wanted to add some logging in the catch or winston was required by mistake

iamigo:
updated, @shriramshankar

PR #602, opened 2017-09-20
Validate sg context

shriramshankar:


shriramshankar:
@iamigo @harshkothari410 This PR has the validations defined in the story. Wondering if we have to make it dummy proof? For example, do we have to validate if SG context has some extra fields that are not in the SGT context definition?

db/helpers/generatorUtil.js
```diff
@@ -13,6 +13,31 @@ const common = require('./common');
 const dbErrors = require('../dbErrors');
 
 /**
+ * Function to validate the context filed of the sample generator based on the
```
iamigo: "field"

db/helpers/generatorUtil.js
```diff
@@ -13,6 +13,31 @@ const common = require('./common');
...
+ * contextDefinition field of its related generator template.
+ * @param  {Object} sgCtx - The generator context object     [description]
+ * @param  {Object} sgtCtxDef - The related sample generator template context
+ * definition object
+ * @throws {MissingRequiredFieldrror} If the generator context field does not
+ * have the attributes that are requred by the context definiton field of the
```
iamigo: "required"

shriramshankar:
@iamigo @harshkothari410 Added extra validations and updated the PR.

db/helpers/generatorUtil.js
```diff
@@ -13,6 +13,39 @@ const common = require('./common');
...
+ * have the attributes that are required by the context definiton field of the
+ * sample generator template
+ * @throws {ValidationError} If the number of keys in the generator context do
+ * not match the number of keys in the generator template context.
+ */
+function validateGeneratorCtx(sgCtx, sgtCtxDef) {
+  const sgtCtxDefKeys = sgtCtxDef ? Object.keys(sgtCtxDef) : [];
+  const sgCtxKeys = sgCtx ? Object.keys(sgCtx) : [];
+
+  if (sgCtxKeys.length > sgtCtxDefKeys.length) {
```
iamigo: I think you'd also want to do something like this:

```
const invalidKeys = sgCtxKeys.filter((key) => !sgtCtxDefKeys.includes(key));
if (invalidKeys.length) {
  throw new dbErrors.ValidationError(
  { explanation: `Sample generator context contains invalid keys: ${invalidKeys}` });
}
```

shriramshankar: Nice. I think just this will handle both the cases. I will add this.

shriramshankar:
@iamigo @harshkothari410 PR updated.

db/helpers/generatorUtil.js
```diff
@@ -13,6 +13,41 @@ const common = require('./common');
...
+ * @throws {MissingRequiredFieldrror} If the generator context field does not
+ * have the attributes that are required by the context definiton field of the
+ * sample generator template
+ * @throws {ValidationError} If the number of keys in the generator context do
+ * not match the number of keys in the generator template context.
+ */
+function validateGeneratorCtx(sgCtx, sgtCtxDef) {
+  const sgtCtxDefKeys = sgtCtxDef ? Object.keys(sgtCtxDef) : [];
+  const sgCtxKeys = sgCtx ? Object.keys(sgCtx) : [];
+  const sgtCtxDefKeysSet = new Set(sgtCtxDefKeys);
```
iamigo: Converting to a set is interesting--makes it seem like it should *also* be an error if you have the same key defined more than once in ctx and/or ctx def, right?

shriramshankar: Created a user story for this. Looks like we have the same problem for all the JSON/JSONB columns. The story covers adding this validation across all the refocus models.

PR #603, opened 2017-09-20
refresh salesforce lightning design system version 2.4.3

kamilsmuga:
- Bump from 2.1.0-rc.3 to 2.4.3
- using actual minified file (file in the master isn't minified)

Affected files (excluding Rooms - that's fine):
- view/admin/index.pug
- view/perspective/perspective.pug
- view/perspectiveBeta/perspective.pug


kamilsmuga:
@iamigo @annyhe after some further thought - there might be an impact on existing pages (namely: view/admin/index.pug, view/perspective/perspective.pug, view/perspectiveBeta/perspective.pug) - in general, lightning design system is backward compatible, but it changes some things like how bold fonts are, etc. 

I can think of two ways forward (1) do you want to verify that perspectives and admin look good enough with new design (it's subjective and tests will not catch this) (2) we can create a new file in static and append release number - this way only stuff that really wants the new version of design system will use it

iamigo:
screen shots please

annyhe:
We barely use the admin pages so drastic changes are less risky there. The perspective page is widely used. We want to make sure changing the style does not corrupt existing styles on the perspectives page.
@iamigo do we want to release this behind a ?beta=true queryParam?

kamilsmuga:
screen shot 2017-09-21 at 7 48 32 pm
@iamigo @annyhe it breaks perspectives (left is new lightning, right is old).

Should we maintain what we have and add a new file under /static? 
I do not have the bandwidth to fix perspectives css at this time.

iamigo:
We can't break the "edit perspective" modal like this, so please version/namespace the new lightning if you need to use the latest, AND please create a user story to bring the perspective page up to current lightning ver.

kamilsmuga:
Ok, versioning will be added in #605

PR #608, opened 2017-09-22
removed FAST_FAIL_DUPLICATE_SUBJECT flag

harshkothari410:


api/v1/controllers/subjects.js
```diff
@@ -415,7 +415,6 @@ module.exports = {
      */
     if (featureToggles.isFeatureEnabled(sampleStoreConstants.featureName) &&
```
shriramshankar: Looks good.  Just the comment in line 411 needs to be updated to remove the words "fast fail".

PR #604, opened 2017-09-20
Refactor do put

annyhe:
- move on cached sample PUT to sample controller
- move PUT on generator to generator controller
- reference shared functions in doPut

PR #609, opened 2017-09-22
Order collectors on GET, POST, PUT, and PATCH generator

annyhe:
- api level only
- change api/generator tests to check for order of collectors


annyhe:
Dependent on #604

api/v1/helpers/verbs/doGet.js
```diff
@@ -76,10 +76,20 @@ function doGet(req, res, next, props) {
     }
 
     getPromise.then((o) => {
+
+      // o is read only.
+      let returnObj = o;
+
+      // order collectors by name
+      if (props.modelName === 'Generator' && o.collectors) {
+        returnObj = JSON.parse(JSON.stringify(o.get()));
```
shriramshankar: I think you do not need to do stringify and then parse the JSON. just doing o.get is enough.

api/v1/helpers/verbs/postUtils.js
```diff
@@ -81,6 +82,14 @@ function handlePostResult(o, resultObj, props, res, req) {
       realtimeEvents.sample.add, props.associatedModels.aspect);
   }
 
+  // order collectors by name
+  if (props.modelName === 'Generator' && o.collectors) {
```
shriramshankar: I thought one of the purposes of the refactor, out, patch and post story was not to do something like this ''props.ModelName ==='Generator'  for a specific model. Wondering if you want to move it to the controller. Same comment for get, patch and put.

api/v1/helpers/verbs/utils.js
```diff
@@ -87,8 +94,22 @@ function handleUpdatePromise(resultObj, req, retVal, props, res) {
...
+    .json(responsify(returnObj, props, req.method));
+}
+
+/**
+ * @param {Array} arr Array of objects
+ * @param {String} fieldName The field to sort by
+ * @returns {Array} Array sorted by field
+ */
+function sortArrayObjectsByField(arr, fieldName) {
+  const arrCopy = JSON.parse(JSON.stringify(arr));
```
shriramshankar: Wondering why are we stringifying and parsing the json here?

api/v1/helpers/verbs/doGet.js
annyhe: the reason we are doing a deep copy is because o.get() returns a read only object. Hence we deep copy the o.get().

api/v1/helpers/verbs/utils.js
annyhe: if we don't deep copy, the sort will be in-place. We do the deep copy to preserve the purity of the function (ie. not modify the params)

api/v1/helpers/verbs/postUtils.js
annyhe: added moving the generator specific part to the refactor responsify story. 
this will be done in a separate PR.

annyhe:
@shriramshankar Updated. Please review.

api/v1/helpers/verbs/utils.js
```diff
@@ -87,8 +93,21 @@ function handleUpdatePromise(resultObj, req, retVal, props, res) {
...
+}
+
+/**
+ * Sorts an array in-place.
+ *
+ * @param {Array} arr Array of objects
+ * @param {String} fieldName The field to sort by
+ */
+function sortArrayObjectsByField(arr, fieldName) {
+  arr.sort((a, b) => a[fieldName].localeCompare(b[fieldName]));
```
shriramshankar: arr.sort((a, b) => a[fieldName].localeCompare(b[fieldName])); is the only thing this function is doing.  Probably does not add value to make it a function ?

annyhe: making it into a function is good for:
1) testing
2) function header describes what the one liner does.
Any other thoughts or comments or feedback?

PR #610, opened 2017-09-22
Db model: AuditEvents

pallavi2209:


db/model/auditevent.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * db/model/auditevent.js
+ *
+ */
+
+const u = require('../helpers/auditEventUtils');
+const assoc = {};
+
+module.exports = function bot(seq, dataTypes) {
```
iamigo: `function bot...` ?

db/model/auditevent.js
```diff
@@ -0,0 +1,52 @@
...
+    eventLog: {
+      type: dataTypes.JSONB,
+      allowNull: false,
+      validate: {
+        contains: u.validateEventLog,
+      },
+    },
+  }, {
+    classMethods: {
+      getTagAssociations() {
```
shriramshankar: association with Tags is not required.

shriramshankar:
We will allowing only create on this table. No updates or deletes.

db/model/auditevent.js
pallavi2209: oops, I should try not to copy paste.

pallavi2209:
oh-oh Only create. That would have saved some work. The story said CRUD tests. Let me update the PR then.

pallavi2209:
@shriramshankar Updated the PR. Let me know if I missed something.

pallavi2209:

db/model/auditevent.js
```diff
@@ -0,0 +1,57 @@
...
+      type: dataTypes.STRING(constants.fieldlen.longish),
+      allowNull: false,
+    },
+    resourceType: {
+      type: dataTypes.STRING(constants.fieldlen.longish),
+      allowNull: false,
+    },
+    isError: {
+      type: dataTypes.BOOLEAN,
+      allowNull: false,
```
iamigo: Let's maybe make isError *not* required but default to false, so if you don't include that it will be false.

tests/db/model/auditevent/create.js
```diff
@@ -0,0 +1,148 @@
...
+    AuditEvent.create(auditEventObj)
+    .then(() => done(tu.valError))
+    .catch((err) => {
+      expect(err.name).to.equal('SequelizeValidationError');
+      expect(err.message).to.contain('resourceType cannot be null');
+      done();
+    });
+  });
+
+  it('Error, create auditEvent, no isError', (done) => {
```
iamigo: update for isError default false

db/model/auditevent.js
```diff
@@ -0,0 +1,57 @@
...
+module.exports = function auditevent(seq, dataTypes) {
+  const AuditEvent = seq.define('AuditEvent', {
+    id: {
+      type: dataTypes.UUID,
+      primaryKey: true,
+      defaultValue: dataTypes.UUIDV4,
+    },
+    loggedAt: {
+      type: dataTypes.BIGINT,
+      allowNull: false,
```
iamigo: I'm thinking it might make sense to allow this to be null but when it's null to default it to current time. What do you think?

tests/db/model/auditevent/create.js
```diff
@@ -0,0 +1,148 @@
...
+    AuditEvent.create(auditEventObj)
+    .then(() => done(tu.valError))
+    .catch((err) => {
+      expect(err.name).to.equal('SequelizeValidationError');
+      expect(err.message).to.contain('isError cannot be null');
+      done();
+    });
+  });
+
+  it('Error, create auditEvent, null loggedAt', (done) => {
```
iamigo: update if allowing null>>now

pallavi2209:
@iamigo updated the PR. Please feel free to add comments if any :)

PR #611, opened 2017-09-25
On cache off browser logs realtime events

annyhe:
- formerly browser only logs realtime events when cache is ON

### DIFF
--- a/realtime/utils.js
+++ b/realtime/utils.js
@@ -391,7 +391,7 @@ function isIpWhitelisted(addr, whitelist) {
  * When passed in a sample, its related subject and aspect is attached to the
  * sample. If useSampleStore is set to true, the subject ans aspect is fetched
  * for the cache instead of the database.
- * @param {Object} sample - The sample instance.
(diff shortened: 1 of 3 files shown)


realtime/utils.js
```diff
@@ -445,8 +446,19 @@ function attachAspectSubject(sample, useSampleStore, subjectModel,
...
     sub = sub.get ? sub.get() : sub;
+
+    // if asp or sub from redis, writers field == [] instead of undefined
+    if (asp.writers === '[]') {
+      delete asp.writers;
+    }
+
+    if (sub.writers === '[]') {
+      delete sub.writers;
+    }
```
shriramshankar: I think we can delete the aspect and subject writers no matter what.

annyhe: What if the aspect.writers or subject.writers = ['user1', 'user2', ...]? Do we still want to delete the aspect and subject writers?

annyhe:
Updated. Please review @shriramshankar @iamigo

PR #614, opened 2017-09-26
add toggle to throw error on hmset with undefined key

annyhe:
- throws error whenever hmset with undefined key
- excluding the hmset in clock and db folder, as those are the least likely to hmset with undefined key


annyhe:
This is the error, it includes the object being set
screen shot 2017-09-26 at 11 47 29 am

PR #620, opened 2017-09-29
Post audit events

shriramshankar:


shriramshankar:
@iamigo @pallavi2209 This PR is for a minor change to the audit table along with some tests to make sure that date comparison will work as expected. I will incoroprate any comments on this PR into my next one.

api/v1/swagger.yaml
```diff
@@ -1330,7 +1330,6 @@ paths:
                     "Collector", "Subject", etc.
               required:
                 - details
```
shriramshankar: What about making details not required and defaulting it to {}; similar to what we did for the "loggedAt" column?

iamigo: OK feel free to make that change here in swagger, adding default value

shriramshankar: OK, will do it.

shriramshankar:
@iamigo @pallavi2209 Anything else that needs changes in the PR?

shriramshankar:
I have included the changes in #621. Can we merge this, if there is nothing else to review ? @iamigo @pallavi2209

PR #624, opened 2017-09-30
POST aspects/subjects: Correct stringified tags/related links

pallavi2209:


tests/cache/models/aspects/aspectCRUD.js
```diff
@@ -334,4 +334,36 @@ describe('tests/cache/models/aspects/aspectCRUD.js, ' +
     })
     .catch(done);
   });
+
+  it('Create, isPublished false, check tags and related links', (done) => {
+    Aspect.create({
+      name: `${tu.namePrefix}ASPECTNAME1`,
+      isPublished: false,
+      timeout: '110s',
```
iamigo: maybe add tags and related links to this aspect?

tests/cache/models/subjects/subjectCRUD.js
```diff
@@ -332,3 +332,40 @@ describe('tests/cache/models/subjects/subjectCRUD.js >', () => {
...
+    tu.toggleOverride('enableRedisSampleStore', true);
+    done();
+  });
+
+  afterEach(rtu.forceDelete);
+  after(() => tu.toggleOverride('enableRedisSampleStore', false));
+
+  it('post subject with isPublished false, check tags and related links',
+  (done) => {
+    Subject.create({ name: `${tu.namePrefix}s4`, isPublished: false })
```
iamigo: maybe add tags and related links to this subject?

db/model/subject.js
```diff
@@ -315,8 +315,14 @@ module.exports = function subject(seq, dataTypes) {
            * is published and the sampleStoreFeature is enabled
            */
           if (featureToggles.isFeatureEnabled(sampleStoreFeature)) {
+            // Prevent any changes to original inst object data
+            const instData = JSON.parse(JSON.stringify(inst.get()));
```
shriramshankar: can we have the same variable name for this in here and aspect as well.?

shriramshankar:
Wondering if we have the same problem for updates as well?

db/model/subject.js
pallavi2209: I changed the name because here we are getting only datavalues (using get) but in aspects we have full instance. Let me know if you still think we should have same var name and I will change it.

pallavi2209:
@shriramshankar Yes, we have same problem for updates as well. Let me correct that and add tests for that as well.

db/model/subject.js
shriramshankar: Oh ok. I think we should be doing inst.get in aspects too.

pallavi2209: Yeah, I think so too, updated that! Thx!

pallavi2209:

PR #626, opened 2017-10-02
Make swagger docs interactive

BrandonMurnane:


BrandonMurnane:
deployed to devbox and works there aswell as locally

api/v1/swagger.yaml
```diff
@@ -40,9 +40,9 @@ basePath: /v1
 #   url: http://focus.salesforce.com/docs/api/
 
 # =============================================================================
-schemes:
-  - https
-  - http
+# schemes:
+#   - https
+#   - http
```
kamilsmuga: what "schemes" exactly do? I couldn't find the explanation in swagger docs. Just making sure we aren't disabling anything important.

BrandonMurnane: the idea is that you tell it what schemes u want used. But it doesnt work like that and instead will always take the first schema and only use that. If you dont specify which schemas it will read the schema from the url in the window its in

BrandonMurnane:
@kamilsmuga anymore questions or concerns?

PR #629, opened 2017-10-03
POST, PUT and PATCH for aspects: helpEmail or helpUrl required

pallavi2209:


tests/api/v1/aspects/patch.js
```diff
@@ -174,6 +174,240 @@ describe('tests/api/v1/aspects/patch.js >', () => {
         return done();
       });
     });
+
+    describe(`PATCH ${path} helpEmail, helpUrl not set in db >`, () => {
+      before(() => tu.toggleOverride('requireHelpEmailOrHelpUrl', true));
+      after(() => tu.toggleOverride('requireHelpEmailOrHelpUrl', false));
```
iamigo: the "after" function should be restoring the toggle to its original state, and the original state is not necessarily false -- better to save the original value in the before, before doing the override

tests/api/v1/aspects/post.js
```diff
@@ -207,4 +207,104 @@ describe('tests/api/v1/aspects/post.js >', () => {
...
+      .send(aspectToPost)
+      .expect(constants.httpStatus.BAD_REQUEST)
+      .end((err, res) => {
+        if (err) {
+          return done(err);
+        }
+
+        expect(res.body.errors[0].type).to.equal('ValidationError');
+        expect(res.body.errors[0].description).to.equal(
+          'Atleast one these attributes are required: helpEmail,helpUrl'
```
iamigo: "At least one of these attributes is required: ..."

utils/common.js
```diff
@@ -81,8 +81,27 @@ function looksLikeId(key) {
   return constants.POSTGRES_UUID_RE.test(key);
 }
 
+/**
+ * Validates that atleast one of the given fields is present in object.
+ * @param  {Object} reqBody - Request body
+ * @param  {Array} fieldsArr - Fields array
+ * @throws {ValidationError} - If none of the given fields are presnt in object
+ */
+function validateAtleastOneFieldPresent(reqBody, fieldsArr) {
```
iamigo: I think I'd like to see some unit tests for this function in isolation, testing all the combinations of reqBody with no keys/some keys/all keys, and fieldArr with no elements, one element, multiple elements, and null or undefined for each arg, etc.

Comprehensive unit testing of the helper function means we don't have to test all the different permutations every time some other module uses this fn.

api/v1/controllers/aspects.js
```diff
@@ -207,6 +207,14 @@ module.exports = {
    */
   postAspect(req, res, next) {
     validateRequest(req);
+
+    // check that atleast one of the given fields is present in request
+    if (featureToggles.isFeatureEnabled('requireHelpEmailOrHelpUrl')) {
```
shriramshankar: Did we want release it behind a flag ?

pallavi2209:

PR #630, opened 2017-10-04
fixed hmset logging to log the right key

shriramshankar:


shriramshankar:
@annyhe UPdated the PR

PR #632, opened 2017-10-04
Refactor dopost

shriramshankar:


shriramshankar:
@iamigo I will test this in dev box too.

api/v1/helpers/verbs/utils.js
```diff
@@ -786,64 +786,23 @@ function forbidden(next, modelName) {
...
 function checkDuplicateRLinks(rLinkArr) {
-  const uniqlinks = [];
+  if (!rLinkArr) {
+    return;
+  }
+
+  const uniqlinks = new Set();
   rLinkArr.forEach((rLinkObj) => {
-    if (rLinkObj.name && uniqlinks.includes(rLinkObj.name.toLowerCase())) {
+    if (rLinkObj.name && uniqlinks.has(rLinkObj.name.toLowerCase())) {
```
annyhe: is Array.has() an official es6 syntax? Where is .has defined?

tests/cache/models/samples/post.js
```diff
@@ -257,7 +255,30 @@ describe('tests/cache/models/samples/post.js >', () => {
...
 
         expect(res.body.name).to.be.equal(sampleName);
-        done();
+        return done();
+      });
+    });
+
+    it('basic post /samples with invalid token', (done) => {
+      api.post(path)
+      .set('Authorization', 'invalidtoken')
```
annyhe: are we supposed to expect the POST to succeed with invalid token? this test wasn't in the original  tests/cache/models/samples/post.js file

shriramshankar: This test was missed in the original file. It covers line 232 in api/v1/controllers/samples.js now and would have covered line 43 in the pervious version of api/v1/helpers/verbs/postUtils.js

shriramshankar:
@iamigo @annyhe Test this branch locally and in dev box. Realtime events were not send on post, which I fixed. Everything else looks good.

iamigo:
@shriramshankar you did manual testing in devbox for post samples, but what about aspects, subjects, perspectives, users, ... ? Don't they all use these common fns too?

shriramshankar:
@iamigo Done will all the manual testing on dev-box.

PR #633, opened 2017-10-04
Remove post v1/collectors

annyhe:


shriramshankar:
@annyhe Can you also assign the story to your self and close it. ? I think we just triaged it today.

PR #634, opened 2017-10-05
GET /v1/lenses/:key should have a "user" attribute when returnUser toggle is on

annyhe:
- with get.js tests
- add tests on /v1/lenses methods (POST, DELETE, etc.) to show the user is returned

### DIFF
--- a/tests/api/v1/lenses/delete.js
+++ b/tests/api/v1/lenses/delete.js
@@ -67,7 +67,8 @@ describe('tests/api/v1/lenses/delete.js >', () => {
     });
   });
 
-  describe('with returnUser toggle off, user object should be returned: ', () => {
+  describe('with returnUser toggle off, user object should not  ' +
+    'be returned: ', () => {
     beforeEach((done) => {
       u.doSetup()
       .then((lens) => {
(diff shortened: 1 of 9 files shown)


tests/api/v1/lenses/delete.js
```diff
@@ -22,53 +22,93 @@ const ZERO = 0;
...
+        expect(res.body.user).to.be.an('object');
+        expect(res.body.installedBy).to.equal(userId);
+        done();
+      });
+    });
   });
 
-  afterEach(u.forceDelete);
-  after(tu.forceDeleteUser);
+  describe('with returnUser toggle off, user object should be returned: ', () => {
```
jgraff2: "user object should not be returned"

tests/api/v1/lenses/delete.js
```diff
@@ -22,53 +22,93 @@ const ZERO = 0;
...
+      .expect(constants.httpStatus.OK)
+      .end((err, res) => {
+        if (err) {
+          return done(err);
+        }
 
-      expect(res.body.isDeleted).to.not.equal(ZERO);
-      done();
+        expect(res.body.isDeleted).to.not.equal(ZERO);
+        expect(res.body.installedBy).to.not.be.defined;
```
jgraff2: is this supposed to be res.body.user?

tests/api/v1/lenses/delete.js
annyhe: yes will update

tests/api/v1/lenses/delete.js
annyhe: yes will update

PR #635, opened 2017-10-06
convert revalidator to joi

harshkothari410:


db/helpers/common.js
```diff
@@ -17,6 +17,7 @@ const pub = require('../../cache/redisCache').client.pubPerspective;
 const dbconf = require('../../config').db;
 const channelName = require('../../config').redis.perspectiveChannelName;
 const revalidator = require('revalidator');
+const joi = require('joi');
```
iamigo: what about removing references to the revalidator module, and removing it from package.json too?

tests/db/model/generatortemplate/create.js
```diff
@@ -205,13 +205,12 @@ describe('tests/db/model/generatortemplate/create.js >', () => {
     _gt.author.email = 'notanemail';
     _gt.author.url = 'invalid url';
     GeneratorTemplate.create(_gt)
-    .then(() => {
+    .then((x) => {
+      console.log(x);
```
iamigo: no console.log

tests/db/model/generatortemplate/create.js
```diff
@@ -237,11 +236,11 @@ describe('tests/db/model/generatortemplate/create.js >', () => {
     const _gt = u.getGeneratorTemplate();
     _gt.connection.method = 'SPY';
     GeneratorTemplate.create(_gt)
-    .then(() => {
+    .then((x) => {
```
iamigo: why declaring x if not using?

tests/db/model/generatortemplate/create.js
```diff
@@ -237,11 +236,11 @@ describe('tests/db/model/generatortemplate/create.js >', () => {
...
       done(' Error: Expecting validation error');
```
iamigo: no space before "Error:"

db/model/generatortemplate.js
```diff
@@ -13,82 +13,44 @@ const common = require('../helpers/common');
...
-    email: {
-      description: 'Author\'s email address',
-      type: 'string',
-      format: 'email',
-      require: true,
-    },
-  },
-};
+const authorSchema = joi.object().keys({
+  url: joi.string().uri().description('Optional Url given by the author')
```
iamigo: can't use the standard joi uri validator here because we have to allow {{...}}}

tests/db/model/generatortemplate/utils.js
```diff
@@ -31,7 +31,7 @@ const GT_SIMPLE = {
     url: 'git+https://github.com/templates/template-generators.git',
   },
   connection: {
-    url: '{{baseTrustUrl}}/v1/instances/status/preview',
+    url: 'baseTrustUrl/v1/instances/status/preview',
```
iamigo: {{...}} must be allowed as part of connection.url and we must have tests for this

tests/db/model/generatortemplate/create.js
```diff
@@ -205,9 +205,8 @@ describe('tests/db/model/generatortemplate/create.js >', () => {
     _gt.author.email = 'notanemail';
     _gt.author.url = 'invalid url';
     GeneratorTemplate.create(_gt)
-    .then((x) => {
-      console.log(x);
-      done(' Error: Expecting validation error');
+    .then(() => {
```
iamigo: `.then(() => done('Error: Expecting validation error'))`

harshkothari410: in every test its like only i have not modified this . @iamigo So should I change or not ?

db/helpers/common.js
```diff
@@ -265,10 +265,9 @@ function setIsDeleted(Promise, inst) {
  */
 function validateObject(object, schema) {
   const options = { additionalProperties: false };
```
shriramshankar: Looks like options is an unused variable.

harshkothari410: Yeah I missed that. NIce catch

db/model/generatortemplate.js
```diff
@@ -13,82 +13,44 @@ const common = require('../helpers/common');
...
-      'all of the designated subjects in a single request. When set to true, ' +
-      'the url string or url function may only reference context attributes ' +
-      'with defaults.',
-      type: 'boolean',
-      defaultValue: false,
-    },
-  },
-};
+      '["url", "toUrl"] is required.'),
+  toUrl: joi.any().description('The string body of a function which returns ' +
```
iamigo: I think right now it allows array or string, but it should only allow string

PR #642, opened 2017-10-09
env var override for CACHE_EXPIRY_IN_SECS

harshkothari410:
- env var override for CACHE_EXPIRY_IN_SECS
- invalidate the cache on PUT/PATCH operation on Perspective
- Added test to verify above behavior


api/v1/controllers/perspectives.js
```diff
@@ -160,6 +161,11 @@ module.exports = {
    */
   patchPerspective(req, res, next) {
     doPatch(req, res, next, helper);
+
+    // Remove from cache after successfull PATCH
```
iamigo: successful has only one "l"

tests/enableCache/perspectives.js
```diff
@@ -100,6 +100,70 @@ describe(`tests/enableCache/perspectives.js, api: GET ${path} >`, () => {
...
+
+        redisCache.get('/v1/perspectives', (cacheErr, reply) => {
+          expect(reply).to.not.exist;
+          return done();
+        });
+      });
+    });
+  });
+
+  it('After PUT on perspective, response should not present in cache',
```
iamigo: "should not be"

api/v1/controllers/perspectives.js
```diff
@@ -160,6 +161,11 @@ module.exports = {
...
+    if (featureToggles.isFeatureEnabled('enableCachePerspective')) {
+      redisCache.del('/v1/perspectives');
```
iamigo: I guess I don't remember how perspective caching is working... are you deleting the entire perspective cache when a single perspective gets updated?

harshkothari410: Yes only one request we cache not any other request we cache in perspective. I guess it makes more sense to remove cache rather then to find that perspective in cache and update it.

PR #644, opened 2017-10-09
Flapper: break after all statements for upsertBulkCaseSensitive

pallavi2209:


tests/cache/models/redisTestUtil.js
```diff
@@ -115,6 +119,17 @@ module.exports = {
     .catch(done);
   },
 
+  forceDelete(done) {
+    tu.forceDelete(tu.db.Sample, testStartTime)
+    .then(() => tu.forceDelete(tu.db.Aspect, testStartTime))
```
harshkothari410: why it is twice ?

pallavi2209: Keeping this because it is used by other tests.

PR #657, opened 2017-10-16
heartbeat

jgraff2:


api/v1/controllers/collectors.js
```diff
@@ -77,6 +78,59 @@ function reEncryptSGContextValues(sg, authToken, timestamp) {
...
+ * has been updated to use it
+ */
+/**
+ * Register a collector. Access restricted to Refocus Collector only.
+ *
+ * @param {IncomingMessage} req - The request object
+ * @param {ServerResponse} res - The response object
+ * @param {Function} next - The next middleware function in the stack
+ */
+function postCollector(req, res, next) {
```
jgraff2: I added this back in for now so the heartbeat tests work. Once the auto-registration is set up we can change the tests to use it and remove this again.

api/v1/swagger.yaml
```diff
@@ -2205,6 +2205,67 @@ paths:
   # ---------------------------------------------------------------------------
   /collectors:
     x-swagger-router-controller: collectors
+    post:
```
jgraff2: I added this back in for now so the heartbeat tests work. Once the auto-registration is set up we can change the tests to use it and remove this again.

api/v1/helpers/verbs/utils.js
```diff
@@ -841,6 +841,96 @@ function responsify(rec, props, method) {
   return o;
 } // responsify
 
+const collectorMap = {};
```
iamigo: collector map feels like it belongs in its own module (rather than adding code to a thousand-line module)

jgraff2: Makes sense. I moved it to verbs/heartbeatUtils.js.

api/v1/controllers/collectors.js
```diff
@@ -77,6 +79,59 @@ function reEncryptSGContextValues(sg, authToken, timestamp) {
...
+  const collectorToPost = req.swagger.params.queryBody.value;
+  const resultObj = { reqStartTime: req.timestamp };
+  const toPost = req.swagger.params.queryBody.value;
+  helper.model.create(toPost)
+  .then((o) => {
+    if (helper.loggingEnabled) {
+      resultObj.dbTime = new Date() - resultObj.reqStartTime;
+      utils.logAPI(req, resultObj, o);
```
shriramshankar: "utils" is not defined.

api/v1/controllers/collectors.js
```diff
@@ -77,6 +79,59 @@ function reEncryptSGContextValues(sg, authToken, timestamp) {
...
+ * @param  {Object} sg - Sample generator object
+ * @return {Object} Sample generator with generatorTemplate attribute set to the
+ *  full matching generator template object.
+ */
+function attachTemplate(sg) {
+  const { name, version } = sg.generatorTemplate;
+  return GeneratorTemplate.getSemverMatch(name, version)
+  .then((gt) => {
+    if (sg.context && gt.contextDefinition) {
+      sg = JSON.parse(JSON.stringify(sg));
```
shriramshankar: Any reason why a new copy of "sg" is being created?

api/v1/controllers/collectors.js
```diff
@@ -159,44 +213,91 @@ function deregisterCollector(req, res, next) {
...
-  /*
-   * TODO populate generatorsUpdated
-   * - for generators which were already assigned to this collector, look up
-   *   any changes made to the generator since the last heartbeat
-   * - reEncrypt context values using reEncryptSGContextValues function
-   */
-
-  res.status(httpStatus.OK).json(retval);
+  //verify token
+  jwtUtil.verifyCollectorToken(req)
```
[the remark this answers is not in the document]
shriramshankar: Looks like this is done just to make sure that the request is made with an collector token. When the "request header" story is done, we can directly look up the header do to the check. Any thoughts?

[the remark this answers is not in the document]
iamigo: Correct. There is a story to make any changes like that (after your jwtUtils story and after my REQUIRE_ACCESS_TOKEN story).

api/v1/controllers/collectors.js
```diff
@@ -159,44 +214,91 @@ function deregisterCollector(req, res, next) {
...
+  //find collector object
+  .then(() => u.findByKey(helper, req.swagger.params))
+  .then((o) => {
+    keyName = o.name;
+
+    //validate collector
+    if (keyName !== tokenName) {
+      throw new apiErrors.ForbiddenError({
+        explanation: 'Token does not match the specified collector',
+      });
```
shriramshankar: Line 234 to 242 is not required. "verifyCollectorToken" does it already and if we end up using using the "request header" feature, this check would already have been done before the headers are added to the request.

api/v1/controllers/collectors.js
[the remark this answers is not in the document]
shriramshankar: I think Jon was working on moving this check further up in the middleware. That is not required. Correct?

api/v1/controllers/collectors.js
```diff
@@ -159,44 +214,91 @@ function deregisterCollector(req, res, next) {
...
+    )
+  ))
+  .then((added) => retval.generatorsAdded = added)
+  .then(() => Promise.all(
+    retval.generatorsUpdated.map((sg) =>
+      attachTemplate(sg)
+      .then((sg) => reEncryptSGContextValues(sg, authToken, timestamp))
+    )
+  ))
+  .then((updated) => retval.generatorsUpdated = updated)
```
shriramshankar: Are you trying to return something here ? This reads something like `return retval.generatorsUpdated = updated`. Same comment in line 290.

api/v1/controllers/collectors.js
```diff
@@ -159,44 +214,91 @@ function deregisterCollector(req, res, next) {
...
+        explanation: 'Token does not match the specified collector',
+      });
+    } else if (o.status !== 'Running' && o.status !== 'Paused') {
+      throw new apiErrors.ForbiddenError({
+        explanation: `Collector must be running or paused. Status: ${o.status}`,
+      });
+    }
+
+    //setup retval
+    if (heartbeatUtils.collectorMap[o.name]) {
```
shriramshankar: Wondering, if the copy of "collectorMap" that this file has access to, will get the updates made by the generator patch/post requests. I am not very sure but we might have to do something like we did for the "config" object in the refocus-collector repo.

api/v1/controllers/collectors.js
jgraff2: It's possible for the same generator object to be present in multiple places in the collectorMap. This is necessary to prevent it from being encrypted twice.

When I originally wrote this things were structured differently and this was the best place to put it, but now it definitely makes more sense to do this in trackGeneratorChanges before the generator object is added to the collectorMap. Thanks for catching this, I will move it.

api/v1/controllers/collectors.js
jgraff2: I only added this as a temporary solution to make the heartbeat tests work, and they do, so that if statement must never be true. I think it's fine as-is.

api/v1/controllers/collectors.js
jgraff2: verifyCollectorToken only checks that the token belongs to any valid collector that exists. This checks that the collector token matches the collector id or name in the path.

api/v1/controllers/collectors.js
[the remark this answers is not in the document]
jgraff2: Note that I'm also moving lines 239-242 to the middleware. Is that still necessary?

api/v1/controllers/collectors.js
jgraff2: It does. It works fine in the tests.

api/v1/controllers/collectors.js
jgraff2: It doesn't matter, does it? Nothing is being used in the next "then" so it's just a formatting choice.

api/v1/controllers/collectors.js
jgraff2: Maybe I should just update verifyCollectorToken to do this?
Are there any endpoints a collector needs to access that aren't tied to the key in the path?

api/v1/helpers/verbs/heartbeatUtils.js
```diff
@@ -0,0 +1,107 @@
...
+const UPDATED = 3;
+const NOCHANGE = 4;
+
+/**
+ * Track generator changes to be sent in the heartbeat response.
+ * Maintain collectorMap such that it always maps collector names to lists of
+ * generators added, deleted, and updated since the last heartbeat from that
+ * collector.
+ *
+ * @param {Object} inst - The Generator that has been changed
```
shriramshankar: generator instead of inst.

api/v1/controllers/collectors.js
shriramshankar: There is a story to make sure that collector token cannot be spoofed. So, there will not be a case where the those 2 are different.

api/v1/controllers/collectors.js
shriramshankar: By not enclosing `retval.generatorsUpdated = updated` within the curly brackets you are trying to return the assignment operator.

jgraff2: Yeah, I know. Does it matter?
I'm happy to add brackets if you think it makes it more readable.

shriramshankar: No problem. You can leave it this way.

api/v1/controllers/collectors.js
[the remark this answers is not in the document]
jgraff2: I'm not able to move it to the middleware cleanly and it sounds like there's more work coming that will impact this. I think we should leave it as-is for now and revisit later.
I added tests for the token verification and TODOs to move this once it's set up in the middleware.
@iamigo @shriramshankar Ok?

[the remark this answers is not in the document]
jgraff2: This also applies to lines 239-242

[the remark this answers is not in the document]
iamigo: OK

shriramshankar:
Ok to merge, once the Travis build is fixed.

PR #659, opened 2017-10-17
On unpublish aspect, do not remove aspect from redis cache

annyhe:


cache/redisOps.js
```diff
@@ -142,15 +142,16 @@ function deleteKey(type, name) {
  * @param  {String} type - The type of the master list on which the
  *  set operations are to be performed
  * @param  {String} objectName - The object name (like Subject, Aspect, Sample)
- * @param {String} name - Name of the key to be deleted
+ * @param {String} _name - The name of the key to be deleted
```
iamigo: prefer leave the arg name in the fn signature just as "name" (not prefixing with underscore char)

cache/redisOps.js
```diff
@@ -142,15 +142,16 @@ function deleteKey(type, name) {
...
  * @returns {Promise} - which resolves to the values returned by the redis batch
  * command
  */
-function deleteKeys(type, objectName, name) {
+function deleteKeys(type, objectName, _name) {
   if (type !== sampleType) {
     return Promise.reject(false);
   }
 
+  const name = _name.toLowerCase();
```
iamigo: why declaring variable for this? if only using it to construct the nameKey, can't you just pass name.toLowerCase() in line 155?

tests/cache/models/samples/put.js
```diff
@@ -87,7 +87,7 @@ describe(`tests/cache/models/samples/put.js, api: cache: PUT ${path}`, () => {
 
           const _err = res.body.errors[ZERO];
           expect(_err.type).to.equal('ResourceNotFoundError');
-          expect(_err.description).to.equal('Aspect not found.');
+          expect(_err.description).to.equal('Sample not found.');
```
iamigo: Is this the same exact error and message as when using db?

annyhe: actually db says `Aspect not found`
Will double check why the cache says otherwise

annyhe: The message changes from missing Aspect to missing Sample, because in cache the !sample test which throws 404 is performed BEFORE the !aspect test.
Since the redisOps.deleteKeys now actually deletes the samples, the !samples fails before the !aspects check, hence the change in the error code.

cache/redisOps.js
```diff
@@ -171,7 +171,8 @@ function deleteKeys(type, objectName, name) {
       const subjectKey = nameParts[0];
       const aspect = nameParts[1];
       if ((objectName.toLowerCase() === subjectType && nameKey === subjectKey)
-        || (objectName.toLowerCase() === aspectType && name === aspect)) {
+        || (objectName.toLowerCase() === aspectType &&
+          name.toLowerCase() === aspect)) {
```
annyhe: name may be camelCase, but aspect will be lowercase. 
If the case don't match, samples won't be deleted.

tests/cache/models/samples/put.js
iamigo: They should return the same error whether from db or cache

annyhe:

annyhe:
Both db and cache return 404, but db returns no error message and cache returns error message of whether it was the Sample or Aspect or Subject that was missing.
I can go ahead and add the proper error message to db, and update the db tests to show the appropriate error message.

iamigo:
if it's a quick easy fix, please do, otherwise please create bug for it

cache/models/samples.js
```diff
@@ -666,7 +666,7 @@ module.exports = {
       return redisOps.getHashPromise(aspectType, aspectName);
     })
     .then((aspObj) => {
-      if (!aspObj) {
+      if (!aspObj || (aspObj.isPublished == 'false')) {
```
shriramshankar: probably do something like  `if (!aspObj || !aspObj.isPublishe)`

cache/models/samples.js
```diff
@@ -485,7 +485,7 @@ module.exports = {
...
```
shriramshankar: probably do something like  `if (!aspObj || !aspObj.isPublished)`

cache/redisOps.js
shriramshankar: name here will always be lower case. the name is the key to be deleted and the toKey in "/cache/sampleStore.js" function already converts it to lower case

db/model/aspect.js
```diff
@@ -303,17 +303,14 @@ module.exports = function aspect(seq, dataTypes) {
...
           } else if (inst.changed('isPublished')) {
-            if (inst.isPublished) {
-              // Prevent any changes to original inst dataValues object
-              const instDataObj = JSON.parse(JSON.stringify(inst.get()));
 
+            // Prevent any changes to original inst dataValues object
+            const instDataObj = JSON.parse(JSON.stringify(inst.get()));
+            redisOps.hmSet(aspectType, inst.name, instDataObj);
+            if (inst.isPublished) {
               redisOps.addKey(aspectType, inst.name);
```
shriramshankar: the aspect has to be added to the aspect master list even if it is not published.

cache/models/samples.js
annyhe: aspObj.isPublished is a string. !'true' == !'false' => false

cache/models/samples.js
annyhe: see comment above. the =='false' is to coerce the string into a boolean

db/model/aspect.js
annyhe: ok will add.

cache/redisOps.js
annyhe: although the key will be lowecase, the name is can be camelCase as it is passed as a param, from the db/model/aspect.js

annyhe:
The db tests have error messages matching those coming from the cache.
Please review @iamigo @shriramshankar

PR #661, opened 2017-10-17
access token always required (no more feature toggle)

iamigo:
update tests accordingly


iamigo:
Note: I set ".skip" on a heartbeat test because that test required POST /v1/collectors which is in flux right now.

tests/api/v1/collectors/heartbeat.js
```diff
@@ -52,7 +52,7 @@ describe('tests/api/v1/collectors/heartbeat.js >', () => {
     afterEach(u.forceDelete);
     after(tu.forceDeleteUser);
 
-    it('config ok', (done) => {
+    it.skip('config ok', (done) => {
       api.post(`/v1/collectors/${u.toCreate.name}/heartbeat`)
```
annyhe: is skipping this test intentional?

iamigo: Yes @annyhe pls. see my comment on this PR from 21hrs ago.

tests/cache/models/subjects/getAll.js
```diff
@@ -33,11 +33,12 @@ describe('tests/cache/models/subjects/getAll.js, ' +
       token = returnedToken;
       done();
     })
-    .catch(done);
+    .catch(err=>console.log(err) && done(err));
```
jgraff2: Did you mean to leave this in?

iamigo: Good catch - will remove in a minute

tests/cache/models/samples/upsertBulkCaseSensitive.js
```diff
@@ -73,27 +74,29 @@ describe('tests/cache/models/samples/upsertBulkCaseSensitive.js, ' +
...
+  it('new sample: different case name should NOT modify sample name',
+  (done) => {
     const path = '/v1/samples/upsert/bulk';
     const sampleName = '___Subject1|___Aspect2';
 
     api.post(path)
     .set('Authorization', token)
     .send([{ name: sampleName.toLowerCase(), value: '6' }])
-    .then(() => {
+    .then((err, res) => {
```
annyhe: is the err and res necessary here?

iamigo: nope, I had added them so I could drill into where the bug was, but I'll pull them out now

PR #663, opened 2017-10-19
added extra request headers

shriramshankar:
Changes to add extra fields to the request headers like userName, tokenName, profileName, isCollector and isAdmin.

### DIFF
--- a/tests/api/v1/users/put.js
+++ b/tests/api/v1/users/put.js
@@ -135,7 +135,7 @@ describe(`tests/api/v1/users/put.js, PUT ${path} >`, () => {
         expect(res.body.errors).to.have.length(1);
         expect(res.body.errors)
         .to.have.deep.property('[0].type', 'AdminUpdateDeleteForbidden');
-        done();
+        return done();
       });
     });
 
@@ -156,7 +156,7 @@ describe(`tests/api/v1/users/put.js, PUT ${path} >`, () => {
         }
 
         expect(res.body.profileId).to.equal(profileTwoId);
-        done();
+        return done();
       });
     });
   });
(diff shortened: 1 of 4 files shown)


tests/api/v1/users/put.js
```diff
@@ -156,63 +156,63 @@ describe(`tests/api/v1/users/put.js, PUT ${path} >`, () => {
...
         expect(res.body.profileId).to.equal(profileTwoId);
-        done();
+        return done();
       });
     });
   });
 
   describe('normal user >', () => {
     it('cannot change its profileId',
       (done) => {
```
iamigo: pls. get rid of line break before `(done) => {

tests/utils/jwtUtil.js
```diff
@@ -0,0 +1,169 @@
...
+  let userInst;
+  let collectorInst;
+  let userToken;
+  let collectorToken;
+  const testStartTime = new Date();
+  const predefinedAdminUserToken = jwtUtil.createToken(
+      adminUser.name, adminUser.name
+  );
+
+  // dummpy callback that returns a promise.
```
iamigo: dummy?

utils/jwtUtil.js
```diff
@@ -141,7 +159,14 @@ function verifyUserToken(req, cb) {
...
 
-    req.user = user;
+    req.user = user.get();
+    return Profile.isAdmin(req.user.profileId);
+  })
+  .then((isAdmin) => {
+    assignKeyValue(req.headers, 'userName', req.user.name);
+    assignKeyValue(req.headers, 'tokenName', decodedData.tokenname);
+    assignKeyValue(req.headers, 'profileName', req.user.profile.name);
+    assignKeyValue(req.headers, 'isAdmin', isAdmin);
```
iamigo: needs "IsCollector" set to false - for all three situations (bot, user, collector) we need to set ALL of these new named headers so that we are certain to be overwriting something the user might send in to try to spoof us

shriramshankar: Yes. Makes sense.

tests/utils/jwtUtil.js
```diff
@@ -0,0 +1,169 @@
...
+        expect(request.headers.profileName).to.equal('___myProfile');
+        expect(request.headers.tokenName).to.equal('___myRefocusUser');
+        expect(request.headers.isAdmin).to.equal(false);
+        expect(request.headers.isBot).to.equal(false);
+        expect(request.headers.isCollector).to.equal(false);
+        return done();
+      }).catch(done);
+    });
+
+    it('verifyToken with invalid token', (done) => {
```
pallavi2209: I am a bit confused with this test - shouldn't this throw some kind of forbidden error because token is invalid?

shriramshankar: Yes, I expected it to throw an error too but it did not. So I convinced myself with the following argument. 
"When an error is thrown by verifyToken function, all we are doing it calling the callback with the error as an argument. The callback that I am using just resolves to true no matter what. The callback used by the "swaggerSecurity" middleware is probably sophisticated enough to handle all the cases" 
Does this make sense ?

pallavi2209: Yes, I think that makes sense. And we have corresponding test in tests/enforceToken/verbs.js. Thanks for explaining!

iamigo:
@shriramshankar pls resolve conflicts

shriramshankar:
I was reading about JWT token earlier in the day. Any thoughts about encoding all the required header information in the jwt claim while creating the token and adding it back to the header immediately after the token is encoded ?

utils/jwtUtil.js
```diff
@@ -19,10 +19,52 @@ const User = require('../db/index').User;
...
+
+/**
+ * Assigns user related information to the request header.
+ * @param  {Object}  req  - The request object
+ * @param  {String}  tokenName - Token name parsed from the token
+ * @param  {Boolean} isAdmin  - Flag to indicate if the request was made by an
+ * Admin user
+ */
+function assignUserHeaders(req, tokenName, isAdmin) {
+  assignKeyValue(req.headers, 'userName', req.user.name);
```
iamigo: I had intentionally written these header names up in the story as starting with upper-case letters so that they would *not* look like variable names, since they're actually internal constants... what was your rationale for making them start with lower-case? Should we prefix then with "Refocus" like RefocusTokenName, RefocusIsBot, etc.? Or change them to TOKEN_NAME, PROFILE_NAME, USER_NAME, etc. so they look more like constant names? Thoughts?

shriramshankar: Missed it!! I can change it to start with an upper case. I like the way it is defined in the story.

shriramshankar:
@iamigo @pallavi2209 PR updated.

utils/jwtUtil.js
```diff
@@ -19,10 +19,52 @@ const User = require('../db/index').User;
...
+ * @param  {String} userName  - The username parsed from the token
+ * @param  {String} tokenName - The tokenname parsed from the token
+ */
+function assignCollectorHeaders(req, userName, tokenName) {
+  assignKeyValue(req.headers, 'UserName', userName);
+  assignKeyValue(req.headers, 'TokenName', tokenName);
+  assignKeyValue(req.headers, 'IsCollector', true);
+  assignKeyValue(req.headers, 'IsAdmin', false);
+  assignKeyValue(req.headers, 'IsBot', false);
+} // assignCollectorHeaders
```
iamigo: Would assigning 'ProfileName' to empty string actually set a header with an empty string, or does it delete the named header? In either case, we should do that.

PR #665, opened 2017-10-20
Change Bot Data Values to Text

tausifmuzaffar:
Bot Data will be using very big values sometimes and so we need to make that column, because strings max at 255


tausifmuzaffar:
@iamigo  tested it locally up and down. Do i do a manually release to devbox to test it? And the Payload limit is how we want to limit it

iamigo:
Yes, manually deploy to devbox

tausifmuzaffar:
@iamigo i have tested it on devbox

tausifmuzaffar:
@kamilsmuga please can you look at this pr?

PR #671, opened 2017-11-02
rename the rate limit environment variables

jgraff2:


docs/docs/01-quickstart.md
```diff
@@ -301,10 +301,11 @@ Refocus lets you limit the number of requests per user--once the user reaches th
 
 Configure your limits using the following environment variables:
```
iamigo: Now that the env var names are in line with the actual express limiter module, also add a link to that module here so people can get more info if they need it.

config.js
```diff
@@ -60,10 +60,11 @@ const CACHE_EXPIRY_IN_SECS = pe.CACHE_EXPIRY_IN_SECS ||
   DEFAULT_CACHE_EXPIRY_IN_SECS;
 
 // request limiter settings
-const rateLimit = pe.DDOS_RATE_LIMIT;
-const rateWindow = pe.DDOS_RATE_WINDOW;
-const endpointToLimit = pe.DDOS_ENDPOINT_TO_LIMIT;
-const httpMethodToLimit = pe.DDOS_HTTP_METHOD_TO_LIMIT;
+const expressLimiterPath = pe.EXPRESS_LIMITER_PATH;
+const expressLimiterMethod = pe.EXPRESS_LIMITER_METHOD;
+const expressLimiterLookup = pe.EXPRESS_LIMITER_LOOKUP || 'headers.UserName';
```
iamigo: I think this one needs to JSON parse the string, in case the user provides an array of strings here

config.js
```diff
@@ -60,10 +60,12 @@ const CACHE_EXPIRY_IN_SECS = pe.CACHE_EXPIRY_IN_SECS ||
...
 
 // request limiter settings
-const rateLimit = pe.DDOS_RATE_LIMIT;
-const rateWindow = pe.DDOS_RATE_WINDOW;
-const endpointToLimit = pe.DDOS_ENDPOINT_TO_LIMIT;
-const httpMethodToLimit = pe.DDOS_HTTP_METHOD_TO_LIMIT;
+const expressLimiterPath = pe.EXPRESS_LIMITER_PATH;
+const expressLimiterMethod = pe.EXPRESS_LIMITER_METHOD;
+const expressLimiterLookup =
+  configUtil.csvToArray(pe.EXPRESS_LIMITER_LOOKUP || 'headers.UserName');
```
jgraff2: how's this? 
So the user would specify "headers.UserName,headers.x-forwarded-for"

docs/docs/01-quickstart.md
```diff
@@ -300,11 +300,13 @@ POST to `v1/subjects/USA/child` using the following body:
 Refocus lets you limit the number of requests per user--once the user reaches the limit for the specified time interval, subsequent requests will get http status code 429 (Too Many Requests).
 
 Configure your limits using the following environment variables:
+For more details, see the [express-limiter](https://www.npmjs.com/package/express-limiter) module.
```
jgraff2: link added.

config.js
iamigo: ok, but then we need to be more verbose in the documentation around how to format the lookup

shriramshankar: Does it work if a CSV is passed.? A test to prove that this work will be good.

jgraff2: just updated it.

docs/docs/01-quickstart.md
```diff
@@ -300,11 +300,13 @@ POST to `v1/subjects/USA/child` using the following body:
...
 
-- `DDOS_RATE_WINDOW` - the time interval in milliseconds after which the limits are reset
-- `DDOS_RATE_LIMIT` - the total number of requests allowed by each user in the specified time interval
-- `DDOS_ENDPOINT_TO_LIMIT` - the URI to limit; allows wildcards; use "*" to limit  all requests
-- `DDOS_HTTP_METHOD_TO_LIMIT` - the HTTP method to limit; use "all" (in lower case) to limit all HTTP methods
+- `EXPRESS_LIMITER_PATH` - the URI to limit; allows wildcards; use "*" to limit  all requests
+- `EXPRESS_LIMITER_METHOD` - the HTTP method to limit; use "all" (in lower case) to limit all HTTP methods
+- `EXPRESS_LIMITER_LOOKUP` - the value(s) to lookup on the request object. default: 'headers.UserName'. To specify multiple values, separate them with commas, e.g. "headers.UserName,headers.x-forwarded-for".
```
jgraff2: here

config.js
iamigo: Just modify the existing test to use a comma-separated lookup instead of just headers.UserName :)

shriramshankar: I was talking about the limiter. We do not have any test to show that the lookup works when it is asked to lookup by more than one parameters.

iamigo: Ohhh, no, we don't. Let's create a story to add that--don't want too much scope creep on Jon's story

jgraff2: Actually, I was already working on that. Just checked it in.

PR #672, opened 2017-11-02
Remove the "cacheGetSamplesByNameWildcard" short-term toggle.

iamigo:
Standard behavior now is to cache the response from a samples query if using wildcard search by name. The field list must also match to return a cached response.

### DIFF
--- a/api/v1/controllers/samples.js
+++ b/api/v1/controllers/samples.js
@@ -46,7 +46,8 @@ const redisCache = require('../../../cache/redisCache').client.cache;
  * @param {String} cacheKey - Cache Key
  * @param {Integer} cacheExpiry -  Cache expiry time
  */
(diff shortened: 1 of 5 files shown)


api/v1/controllers/samples.js
```diff
@@ -86,22 +87,30 @@ module.exports = {
...
+    // Use cache for wildcard sample query
+    const query = req.query.name;
+    helper.cacheEnabled = query && (query.indexOf('*') > -1);
+    helper.cacheKey = helper.cacheEnabled ? query : null;
+
+    /*
+     * Include field list as part of cache key so that we only use a cached
+     * response if it has the same field list.
+     */
+    if (helper.cacheKey && req.query.fields) {
```
harshkothari410: Do we have test for this ?

PR #673, opened 2017-11-04
Simplify jwtUtil

shriramshankar:
Simplified `utils/jwtUtil.js` by refactoring `getTokenDetailsFromRequest` and deleting `getTokenDetailsFromString`. It also has other necessary changes to fix the callers of these functions.


api/v1/controllers/lenses.js
```diff
@@ -412,21 +410,9 @@ module.exports = {
...
-        })
-        .catch((err) => {
-          if (err.status === httpStatus.FORBIDDEN) {
-            return createLens();
-          }
-
-          return u.handleError(next, err, helper.modelName);
-        });
+        const user = req.user;
+        seqObj.installedBy = user.id;
```
annyhe: what if user is undefined? can the user be undefined when the returnUser toggle is true?

api/v1/helpers/verbs/doDelete.js
```diff
@@ -37,8 +37,7 @@ function doDelete(req, res, next, props) {
   if (featureToggles.isFeatureEnabled(constants.featureName) &&
   props.modelName === 'Sample') {
     const sampleName = req.swagger.params.key.value.toLowerCase();
-    delPromise = u.getUserNameFromToken(req)
-      .then((user) => redisModelSample.deleteSample(sampleName, user));
+    delPromise = redisModelSample.deleteSample(sampleName, req.user);
```
annyhe: can req.user be undefined here? Samples controller goes directly to doDelete.

jobQueue/jobWrapper.js
```diff
@@ -126,11 +126,10 @@ function logJobOnComplete(req, job) {
       }
 
       logObject.ipAddress = activityLogUtil.getIPAddrFromReq(req);
-      jwtUtil.getTokenDetailsFromRequest(req)
-      .then((resObj) => {
-        logObject.user = resObj.username;
-        logObject.token = resObj.tokenname;
-      }).catch(() => {}); // no-op
+      const tokenDetails = jwtUtil.getTokenDetailsFromRequest(req);
+      logObject.user = tokenDetails.userName;
```
annyhe: getTokenDetailsFromRequest is still asynchronous, correct?
It calls getTokenDetailsFromTokenString, which returns a promise.

tests/api/v1/helpers/utils.js
```diff
@@ -63,46 +43,14 @@ describe('tests/api/v1/helpers/utils.js >', () => {
       .catch(done);
     });
 
-    it('must throw an error for invalid tokens', (done) => {
-      const fakeReq = { headers: { authorization: 'invalidtoken' } };
+    it('with not containing user object in the request', (done) => {
```
annyhe: "without user object in request"

utils/apiLog.js
```diff
@@ -11,9 +11,9 @@
  *
  */
 'use strict'; // eslint-disable-line strict
-const jwtUtil = require('./jwtUtil');
 const featureToggles = require('feature-toggles');
 const activityLogUtil = require('./activityLog');
+const jwtUtil = require('../utils/jwtUtil');
```
annyhe: do we need to change this line?

utils/apiLog.js
```diff
@@ -109,16 +110,11 @@ function logAPI(req, resultObj, retval, recordCountOverride) {
       logObject.request_id = req.headers['x-request-id'];
     }
 
-    // If API token enabled, extract user, token to update log object
-    jwtUtil.getTokenDetailsFromRequest(req)
-    .then((resObj) => {
-      logObject.user = resObj.username;
-      logObject.token = resObj.tokenname;
+    const tokenDetails = jwtUtil.getTokenDetailsFromRequest(req);
+    logObject.user = tokenDetails.userName;
```
annyhe: same comment as above, is getTokenDetailsFromRequest synchronous now?

api/v1/helpers/verbs/doDelete.js
shriramshankar: The user object is always attached to the req by the verifyUserToken in jwtUtil. So, req.user will never be undefined.

api/v1/controllers/lenses.js
shriramshankar: The user object is always attached to the req by the verifyUserToken in jwtUtil. So, req.user will never be undefined.

shriramshankar:
@annyhe @iamigo update the PR with the requested changes. I am also ready for my next PR that is based on these changes.

PR #674, opened 2017-11-07
Add API for reregistering a deregistered collector

annyhe:
- add the api endpoint for re-registering a collector, after it has been de-registered.
- includes tests


tests/api/v1/collectors/re-register.js
```diff
@@ -0,0 +1,131 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/api/v1/collectors/re-register.js
```
iamigo: no dash in file name

api/v1/controllers/collectors.js
```diff
@@ -206,6 +206,37 @@ function deregisterCollector(req, res, next) {
...
 /**
+ * Reregister a collector. Access restricted to Refocus Collector only.
+ *
+ * @param {IncomingMessage} req - The request object
+ * @param {ServerResponse} res - The response object
+ * @param {Function} next - The next middleware function in the stack
+ */
+function reRegisterCollector(req, res, next) {
+
+  //TODO: remove token verification once it's set up in middleware for collectors
```
shriramshankar: The middleware has already been setup. Check for req.headers.IsCollector to make sure that it is a collector.

api/v1/controllers/collectors.js
```diff
@@ -206,6 +206,37 @@ function deregisterCollector(req, res, next) {
 } // deregisterCollector
 
 /**
+ * Reregister a collector. Access restricted to Refocus Collector only.
```
shriramshankar: Wondering if a user can reregister a collector.?

annyhe: What would be the use case? What are your thoughts @iamigo

annyhe:

api/v1/controllers/collectors.js
iamigo: I think just a collector.

annyhe:
updated. Please review @shriramshankar

PR #675, opened 2017-11-08
Update API controllers for Collectors and Samples to use req object instead of jwtUtils

shriramshankar:
fixed collector and sample controllers to use the information added by jwtUtils


api/v1/controllers/collectors.js
```diff
@@ -109,15 +109,9 @@ function attachTemplate(sg) {
  */
 function postCollector(req, res, next) {
   const collectorToPost = req.swagger.params.queryBody.value;
-  const resultObj = { reqStartTime: req.timestamp };
   const toPost = req.swagger.params.queryBody.value;
   helper.model.create(toPost)
   .then((o) => {
-    if (helper.loggingEnabled) {
```
jgraff2: why are you removing this?

api/v1/controllers/samples.js
```diff
@@ -197,7 +196,8 @@ module.exports = {
         u.checkDuplicateRLinks(rLinks);
       }
 
-      redisModelSample.patchSample(req.swagger.params, req.user.name)
+      const userName = req.user ? req.user.name : undefined;
+      redisModelSample.patchSample(req.swagger.params, userName)
```
jgraff2: What happens if the user is undefined? Does patchSample throw an error?
Shouldn't that be impossible since we're always enforcing tokens?

tests/api/v1/collectors/heartbeat.js
```diff
@@ -171,22 +168,6 @@ describe('tests/api/v1/collectors/heartbeat.js >', () => {
         .end(done);
       });
 
-      it('valid collector token, doesnt match collector - by name', (done) => {
-        api.post(`/v1/collectors/${collector1.name}/heartbeat`)
```
jgraff2: Why are you removing this?

api/v1/controllers/collectors.js
shriramshankar: The postCollector is just a stub for your heartbeat test to run for now and this block is not required. Also, this was checked in with defining the `utils` variable below.

tests/api/v1/collectors/heartbeat.js
shriramshankar: This was removed to match the deleted lines from 238 to 243 in the heartbeat code.

jgraff2: This is a legitimate test, right? If deleting those lines causes this test to fail, then they shouldn't be deleted.
There was a TODO in there that they should be removed when it was working in the middleware. If it's not, then they shouldn't be removed yet.

shriramshankar: This is not a legitimate test. There is a story to make sure that this does not happen.

jgraff2: Are you saying it's ok to send a heartbeat for collector A with a token registered to collector B? If that's not allowed then we need this test. If there's a story to move it to the middleware, it's still going to need to be tested.

I don't think we should move it until it's ready, and we shouldn't delete the test now only to add it back later.

api/v1/controllers/samples.js
shriramshankar: I had it has req.user.name in my previous PR but later realized that if at all we were to access any of these endpoints using a collector, req,user will not be defined.

tests/api/v1/collectors/heartbeat.js
shriramshankar: It is not ok for a collector A to send a heartbeat with a token registered to collector B. Lines 238 to 243 and these tests should not have been checked in. I missed it while I was reviewing the PR. There is a story to handle it, the code and its related tests should be added as a part of the story. We should be adding code now to be deleted later.

shriramshankar:
@iamigo @jgraff2  Added the collector spoofing check back to the heartbeat. Please review.

PR #687, opened 2017-11-16
Add cookie with userId to perspective response so the perspective can…

iamigo:
… pass it to the lens.


iamigo:
@annyhe I couldn't find any view tests which checked for a document cookie so I just confirmed this manually by looking at cookies in browser after loading perspective page. If you have a suggestion for an automated test for this, please let me know.

PR #689, opened 2017-11-16
Standardize setting default limit and offset. Enforce max limit.

iamigo:
Update tests.


shriramshankar:
@iamigo this is one of the may PR of the `GET v1/samples` from cache optimization story, correct?

iamigo:
@shriramshankar yes

PR #690, opened 2017-11-17
Encode additional information for user tracking in token payload

pallavi2209:
Large number of updates in test files: most of them are updating the admin token with { isAdmin:true, ProfileName: 'Admin'}.
Plus some added/updated tests in tests/utils/jwtUtil.js
Rest of the changes are related to the code which calls createToken - touches a lot of token code. So, once this is merged,  I will also test all the use cases in staging (esp SSO use case for which we do not have integration tests).

@shriramshankar @iamigo : It would be great if you can test this manually as well, in case I missed some use case. 

Some use cases:
User registers first time.
User authenticates.
An SSO user signs in.
An existing SSO user registers.
A token is created using Refocus UI.
Anything else?

Extra change: Corrected Date.now -> Date.now() in createToken.
Note: This PR does not make any change for createToken corresponding to Bots. I am assuming that will be handled by IMC team. fyi @kamilsmuga

### DIFF
--- a/api/v1/controllers/authenticate.js
+++ b/api/v1/controllers/authenticate.js
@@ -15,6 +15,7 @@ const httpStatus = require('../constants').httpStatus;
 const u = require('../helpers/verbs/utils');
 const apiErrors = require('../apiErrors');
 const jwtUtil = require('../../../utils/jwtUtil');
+const Profile = require('../helpers/nouns/profiles').model;
 
 const resourceName = 'authenticate';
 
@@ -48,15 +49,26 @@ module.exports = {
           return u.handleError(next, _err, resourceName);
         }
 
-        // create token
-        const token = jwtUtil.createToken(req.user.name, req.user.name);
-        req.session.token = token;
-        const retObj = {
-          success: true,
-          message: 'authentication succeeded',
-        };
-        u.logAPI(req, resultObj, retObj);
-        return res.status(httpStatus.OK).json(retObj);
+        return Profile.isAdmin(user.profileId)
+        .then((isAdmin) => { // update in token payload if admin
+          const payloadObj = {
+            ProfileName: user.profile.name,
+            IsAdmin: isAdmin,
+          };
+
+          // create token
+          const token = jwtUtil.createToken(
+            req.user.name, req.user.name, payloadObj
+          );
+
+          req.session.token = token;
+          const retObj = {
+            success: true,
+            message: 'authentication succeeded',
+          };
+          u.logAPI(req, resultObj, retObj);
+          return res.status(httpStatus.OK).json(retObj);
+        });
       });
     })(req, res, next);
   },
--- a/api/v1/controllers/collectors.js
+++ b/api/v1/controllers/collectors.js
@@ -117,7 +117,7 @@ function postCollector(req, res, next) {
      * special token for that collector to use for all further communication
      */
     o.dataValues.token = jwtUtil
-      .createToken(toPost.name, toPost.name);
+      .createToken(toPost.name, toPost.name, { IsCollector: true });
     return res.status(httpStatus.CREATED)
       .json(u.responsify(o, helper, req.method));
   })
--- a/api/v1/controllers/collectors.js
+++ b/api/v1/controllers/collectors.js
@@ -349,8 +349,9 @@ function startCollector(req, res, next) {
          * When a collector registers itself with Refocus, Refocus sends back a
          * special token for that collector to use for all subsequent heartbeats.
          */
-        collector.dataValues.token = jwtUtil
-          .createToken(requestBody.name, requestBody.name);
+        collector.dataValues.token = jwtUtil.createToken(
+          requestBody.name, requestBody.name, { IsCollector: true }
+        );
         return res.status(httpStatus.OK)
           .json(u.responsify(collector, helper, req.method));
       });
@@ -376,7 +377,9 @@ function startCollector(req, res, next) {
   })
   .then(() => returnedCollector.update(requestBody))
   .then((retVal) => {
-    retVal.dataValues.token = jwtUtil.createToken(retVal.name, retVal.name);
+    retVal.dataValues.token = jwtUtil.createToken(
+      retVal.name, retVal.name, { IsCollector: true }
+    );
     return res.status(httpStatus.OK)
       .json(u.responsify(retVal, helper, req.method));
   })
--- a/api/v1/controllers/register.js
+++ b/api/v1/controllers/register.js
@@ -16,6 +16,7 @@ const u = require('../helpers/verbs/utils');
 const helper = require('../helpers/nouns/users');
 const jwtUtil = require('../../../utils/jwtUtil');
 const apiErrors = require('../apiErrors');
+const Profile = require('../helpers/nouns/profiles').model;
 
 const resourceName = 'register';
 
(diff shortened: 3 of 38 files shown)


jobQueue/jobWrapper.js
```diff
@@ -126,10 +126,13 @@ function logJobOnComplete(req, job) {
...
-      const tokenDetails = jwtUtil.getTokenDetailsFromRequest(req);
-      logObject.user = tokenDetails.username;
-      logObject.token = tokenDetails.tokenname;
 
+      /**
+       * we already set UserName and TokenName in req headers when verifying
+       * token
+       */
+      logObject.user = req.headers.UserName;
+      logObject.token = req.headers.TokenName;
```
shriramshankar: this is good. Can we also delete `getTokenDetailsFromRequest` from `utils/jwtUtil.js` ? I do not think `getTokenDetailsFromRequest` is used anywhere else.

utils/jwtUtil.js
```diff
@@ -35,35 +42,32 @@ function assignKeyValue(object, key, value) {
...
+    assignKeyValue(req.headers, 'TokenName', decodedTokenData.tokenname);
+  }
+
+  Object.keys(headersWithDefaults).forEach((headerName) => {
+    let headerValue;
+    if (decodedTokenData.hasOwnProperty(headerName)) {
+      headerValue = decodedTokenData[headerName];
+    } else {
+      headerValue = headersWithDefaults[headerName];
+    }
```
shriramshankar: maybe we can do something like 
`headerValue = decodedToken[headerName] || headersWithDefaults[headerName]`  instead of the else block

utils/jwtUtil.js
```diff
@@ -260,47 +263,42 @@ function verifyToken(req, cb) {
...
     username: userName,
-    timestamp: Date.now,
+    timestamp: Date.now(),
   };
+
+  /*
+    If payload not given, no additional headers are set.
+    If payload is given, set only the headers which matches the headers in
+    headersWithDefaults.
+   */
```
shriramshankar: /*
* 
* 
*/
for multiline comments.

utils/jwtUtil.js
```diff
@@ -35,35 +42,32 @@ function assignKeyValue(object, key, value) {
...
-  assignKeyValue(req.headers, 'UserName', req.user.name);
-  assignKeyValue(req.headers, 'TokenName', tokenName);
-  assignKeyValue(req.headers, 'ProfileName', req.user.profile.name);
-  assignKeyValue(req.headers, 'IsAdmin', isAdmin);
-  assignKeyValue(req.headers, 'IsCollector', false);
-  assignKeyValue(req.headers, 'IsBot', false);
-} // assignUserHeaders
+function assignHeaderValues(req, decodedTokenData) {
+  // username and tokenname should always be there if createToken function is
+  // used for creating token
```
shriramshankar: /*
*
*/
for multiline comments.

config/passportconfig.js
```diff
@@ -60,20 +60,29 @@ module.exports = (passportModule) => {
...
+      }
+
+      return Profile.create({ name: 'RefocusUser' });
+    })
+    .then((profile) =>
+
       // if non-sso register, create new user and token object.
-      let newUserCreated = null;
+      // Note: Used User.findById after creation so that we can get profile
+      // attached to user object.
```
shriramshankar: /*
*
*/
for multiline comments.

api/v1/controllers/collectors.js
```diff
@@ -116,8 +116,9 @@ function postCollector(req, res, next) {
      * When a collector registers itself with Refocus, Refocus sends back a
      * special token for that collector to use for all further communication
      */
-    o.dataValues.token = jwtUtil
-      .createToken(collectorToPost.name, collectorToPost.name);
+    o.dataValues.token = jwtUtil.createToken(
+      collectorToPost.name, collectorToPost.name, { IsCollector: true }
```
iamigo: I think it should include all the other details, too (IsAdmin=false, IsBot=false, etc.)

api/v1/controllers/authenticate.js
```diff
@@ -48,15 +49,26 @@ module.exports = {
...
-          success: true,
-          message: 'authentication succeeded',
-        };
-        u.logAPI(req, resultObj, retObj);
-        return res.status(httpStatus.OK).json(retObj);
+        return Profile.isAdmin(user.profileId)
+        .then((isAdmin) => { // update in token payload if admin
+          const payloadObj = {
+            ProfileName: user.profile.name,
+            IsAdmin: isAdmin,
```
iamigo: I think it should all the other details too, (IsCollector, IsBot, etc.)

api/v1/controllers/register.js
```diff
@@ -43,20 +44,29 @@ module.exports = {
...
-        const userObj = u.responsify(user, helper, req.method);
-        userObj.token = tokenToReturn;
-        req.session.token = tokenToReturn;
-        u.logAPI(req, resultObj, userObj);
-        return res.status(httpStatus.CREATED).json(userObj);
+        return Profile.isAdmin(user.profileId)
+        .then((isAdmin) => { // update in token payload if admin
+          const payloadObj = {
+            ProfileName: user.profile.name,
+            IsAdmin: isAdmin,
```
iamigo: I think it should all the other details too, (IsCollector, IsBot, etc.)

api/v1/controllers/tokens.js
```diff
@@ -80,14 +81,23 @@ module.exports = {
...
     // create token to be returned in response.
     const tokenName = req.swagger.params.queryBody.value.name;
-    const token = jwtUtil.createToken(
-      tokenName, user.name
-    );
+    let token;
+    Profile.isAdmin(user.profileId)
+    .then((isAdmin) => {
+      const payloadObj = {
+        ProfileName: user.profile.name,
```
iamigo: I think it should all the other details too, (IsCollector, IsBot, etc.)

tests/api/v1/admin/rebuildSampleStore.js
```diff
@@ -23,17 +23,14 @@ const Subject = tu.db.Subject;
...
 const initialFeatureState = featureToggles
   .isFeatureEnabled(sampleStore.constants.featureName);
 
 describe('tests/api/v1/admin/rebuildSampleStore.js >', () => {
   describe(`POST ${path} (feature is off) >`, () => {
     let token;
-    const predefinedAdminUserToken = jwtUtil.createToken(
-      adminUser.name, adminUser.name
-    );
+    const predefinedAdminUserToken = tu.createAdminToken();
```
iamigo: nice "cleanup" work here!

tests/api/v1/aspects/postWithCreatedBy.js
```diff
@@ -11,25 +11,16 @@
  */
 'use strict'; // eslint-disable-line strict
 const supertest = require('supertest');
-const adminUser = require('../../../../config').db.adminUser;
```
iamigo: great cleanup work here!

tests/api/v1/bots/post.js
```diff
@@ -47,11 +47,11 @@ describe('tests/api/v1/bots/post.js >', () => {
         return done(err);
       }
 
-      const botToken = jwtUtil
-        .createToken(u.name, u.name);
       const fakeToken = jwtUtil
         .createToken(u.name + 'Fail', u.name + 'Fail');
-      expect(res.body.token).to.equal(botToken);
+
+      // since createToken uses current timestamp, the token value is not fixed.
```
iamigo: Thx for clarifying comment

view/loadView.js
```diff
@@ -76,17 +76,21 @@ function samlAuthentication(userProfile, done) {
       return Profile.findOrCreate({ where: { name: 'RefocusSSOUser' } });
     }
 
+    // profile already attached - default scope applied on find
     return done(null, user);
   })
   .spread((profile) =>
-    User.create({
+    User.create({ // default scope not applied on create, hence no profile
```
iamigo: Can you check for a user story for this, and if not, please create one?

pallavi2209: I think I should make it more clear in the comment. Its just a comment, meaning we do not have profile attached here, so we will use User.find to get the attached profile after this (which I did). No further work needed here.

pallavi2209:
As discussed with @iamigo, I am going to make one more change in addition to the comments in this PR. I will get isAdmin and ProfileName info for the existing tokens and set the header accordingly. Let me update this PR. fyi @shriramshankar

api/v1/controllers/authenticate.js
shriramshankar: Do we have to add it here too? I think when the token is decoded and the headers are set we will have all these information.

pallavi2209: Yeah, we don't need to add the details when creating token. Discussed with Ian.

utils/jwtUtil.js
pallavi2209: @shriramshankar I think this will not work as we expect. Consider we add a new default value in headersWithDefaults['newHeader'] = true
And then I set newHeader to false in decodedToken['newHeader'] = false
Then, I will have req.newHeader set to true (which is opposite of what we set in token in first place).
I am going to add a test for this use case.

pallavi2209:
@iamigo @shriramshankar Updated the PR.

utils/jwtUtil.js
```diff
@@ -208,6 +212,23 @@ function verifyUserToken(req, cb) {
...
+    }
+
+    if (!decodedData.hasOwnProperty('IsAdmin')) {
+      return Profile.isAdmin(req.user.profileId);
+    }
+
+    return decodedData.IsAdmin;
+  })
+  .then((isAdmin) => {
+    decodedData.IsAdmin = isAdmin; // ok to reassign in case of new tokens
```
pallavi2209: @shriramshankar @iamigo Please let me know if there is a better way to do this.

shriramshankar:
Looks good.  We will also have to take care of the tokens created as a part of the start command in the collector.

pallavi2209:
@shriramshankar @iamigo update the PR.

PR #696, opened 2017-11-27
Switch to the 'ratelimiter' package and set it up to allow multiple l…

jgraff2:
…imiter configurations.


docs/docs/01-quickstart.md
```diff
@@ -302,11 +302,13 @@ Refocus lets you limit the number of requests per user--once the user reaches th
 Configure your limits using the following environment variables:
 For more details, see the [express-limiter](https://www.npmjs.com/package/express-limiter) module.
 
-- `EXPRESS_LIMITER_PATH` - the URI to limit; allows wildcards; use "*" to limit  all requests
-- `EXPRESS_LIMITER_METHOD` - the HTTP method to limit; use "all" (in lower case) to limit all HTTP methods
+- `EXPRESS_LIMITER_PATH` - the URI to limit. Can be any valid [express path](http://expressjs.com/en/4x/api.html#path-examples). To specify multiple paths, separate them with commas, e.g. "/v1/aspects,/v1/subjects". Use "*" to limit  all requests
```
annyhe: so the limiter works like this?
EXPRESS_LIMITER_PATH  = "url1, url2"
EXPRESS_LIMITER_METHOD = "post,put"
would limit url1's POST and PUT, and url2's POST and PUT?

what if the user wants to limit only the POST of url1, and the PUT of url2?

tests/limiter/limiter.js
```diff
@@ -37,124 +51,620 @@ describe('tests/limiter/limiter.js >', () => {
...
-  it('(j) Admin user req#3, no ratelimit headers, ok', (done) => {
-    api.post(path)
-    .set('Authorization', predefinedAdminUserToken)
-    .send({ name: tu.namePrefix + 'Limiter7', timeout: '1m' })
-    .expect(constants.httpStatus.CREATED)
-    .expect((res) => {
-      expect(res.headers).to.not.have.property('x-ratelimit-limit');
-    })
-    .end(done);
+  describe.skip('path/method - fork subprocess >', function() {
```
annyhe: is this .skip intentional?

docs/docs/01-quickstart.md
annyhe: what if the user wants to POST url1, PUT url2, and PATCH url3? Does the design allow for that?

jgraff2: Good point. No, that's not possible with the current design.

The configuration could definitely be a lot more flexible. Currently you also can't set different configs for different paths and methods, or have an arbitrary number of limiter configs. We could do that pretty easily, but we'd have to figure out the best way to do all the environment variables.

The purpose of this story was to allow setting up two different configurations for the same path/method so we could restrict it to e.g. no more than 60 requests per hour and also no more than 1 request per 100ms. I think more complicated configuration is outside the scope of this story. The change to allow multiple paths and methods wasn't even really necessary, I just added that because I was already doing it for the lookup. 

Is this something that we need? Should we create a different story for it?

@iamigo

tests/limiter/limiter.js
jgraff2: Yeah, this is the only way I could test the path/method config but it takes a long time to execute. I skipped it so it doesn't make the tests take forever.

rateLimit.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const id = lookupVals.join('_');
+  if (id) {
+    limiterConfig1.id = `1__${id}`;
+    limiterConfig2.id = `2__${id}`;
+  }
+
+  // The order matters. The second one overwrites the limit headers.
+  Promise.resolve()
+  .then(() => applyLimits(limiterConfig2))
```
iamigo: What happens when there is no config for #2?

jgraff2: It will be skipped.
Line 56: return Promise.resolve();

PR #697, opened 2017-11-29
Attach user token to generator added

annyhe:
- include test


tests/api/v1/collectors/heartbeat.js
```diff
@@ -1017,7 +1018,43 @@ describe('tests/api/v1/collectors/heartbeat.js >', () => {
         .then(done)
         .catch(done);
       });
+    });
+
+    describe('added generator >', () => {
+      before(() => tu.toggleOverride('returnUser', true));
+      after(() => tu.toggleOverride('returnUser', false));
```
iamigo: "after" should restore it to whatever it was before, which is not necessarily guaranteed to be "false" -- safer to save the original value in the "before"

annyhe: We are not using the returnUser anywhere else in the test. Even if we are, the other tests ought to setup then teardown. Not sure if it is a good idea to add more code to this section.

jgraff2: don't think it matters much either way.

tests/api/v1/collectors/heartbeat.js
```diff
@@ -1017,7 +1018,43 @@ describe('tests/api/v1/collectors/heartbeat.js >', () => {
...
+          }
 
+          expect(res.body.createdBy).to.equal(userId);
+          const timestamp = Date.now();
+          return Promise.resolve()
+          .then(() => postGenerator(generator2, [collector1]))
+          .then(() => sendHeartbeat(collector1))
+          .then((res) => {
+            const reencryptedSG = res.body.generatorsAdded[0];
+            expect(reencryptedSG.token).to.be.an('string');
```
jgraff2: to.be.a('string')

jgraff2: shouldn't you also check that the value of the token is correct?

annyhe: good catch. Updated

annyhe: do you mean check it is a proper UUID, or the expected UUID? The latter is difficult as the jwtUtil token generation is one way.

jgraff2: ok, makes sense.

jgraff2: If you updated it, doesn't look like it got pushed.

annyhe: @jgraff2 how do I check the value of the token is correct? Do I use jwtUtil.verifyToken(req, callback)? or make a new function in jwtUtil with 
  return jwtVerifyAsync(token, secret, {})
  .then((payLoad) => {
    decodedData = payLoad;
    return User.findOne({ where: { name: decodedData.username } });
  })
  .then((user) => {
    if (!user) {
      throw new apiErrors.ForbiddenError({
        explanation: 'Invalid Token.',
      });
    }

    req.user = user.get();

annyhe: @jgraff2 in retrospect that is a lot of work to check it is a token. Any suggestions?

jgraff2: I think it's fine as-is, given that the token is being created just before this and createToken() is tested elsewhere.

tests/api/interim.js
```diff
@@ -0,0 +1,78 @@
+const expect = require('chai').expect;
```
jgraff2: Looks like you accidentally included code from a different story.

tests/api/v1/collectors/heartbeat.js
iamigo: It's not about whether it's used in *this* test, but rather across all the tests.

PR #698, opened 2017-12-04
Bot Data Socket Emit

tausifmuzaffar:
Create a Bot Data Socket emitter

PR #699, opened 2017-12-04
Events Socket Emit

tausifmuzaffar:
Create a socket emit for Events model


iamigo:
@tausifmuzaffar test failures?

kamilsmuga:
tests fixed @pallavi2209 @pallavisingh9 @iamigo

db/model/event.js
```diff
@@ -68,6 +81,25 @@ module.exports = function event(seq, dataTypes) {
...
+      afterCreate: (instance) => {
+        const changedKeys = Object.keys(instance._changed);
+        const ignoreAttributes = ['isDeleted'];
+        return realTime.publishObject(instance.toJSON(),
+          botEventNames.add, changedKeys, ignoreAttributes, pubOpts);
+      },
+
+      afterUpdate(instance /* , opts */) {
+        return realTime.publishObject(instance.toJSON(),
+          botEventNames.upd, null, null, pubOpts);
```
iamigo: curious why not sending changed keys and ignore attributes here or in afterDelete, like you do in afterCreate?

tests/realtime/utils.js
```diff
@@ -185,8 +192,21 @@ module.exports = {
     return JSON.parse(JSON.stringify(standardBotAction));
   },
 
+  getStandardEvent() {
+    return JSON.parse(JSON.stringify(standardEvent));
+  },
+
   forceDelete(done) {
-    tu.forceDelete(tu.db.Perspective, testStartTime)
+    tu.db.Event.destroy({
```
iamigo: curious why calling destroy directly rather than using tu.forceDelete?

tausifmuzaffar: tu.forcedelete uses name and the event doesn't have so need to use a destroy

db/model/event.js
tausifmuzaffar: overlooked this; ive added the changes

tausifmuzaffar:
@iamigo @kamilsmuga same issue about merging for #698

PR #703, opened 2017-12-08
improve usability of root subject picker in UI for adding/editing perspective

annyhe:
- improves usability by searching rootSubject only when there is text to search for
- filter by whether subject names start with the search text
- this is the bare minimum. Can add tests, change placeholder text, ...


annyhe:
after talking to Jee, these are potential future improvements

annyhe:
@iamigo what is the best way to release this? with ?beta=true url?

annyhe:
do not merge as the result from "edit" perspective differs from when "create" perspective

annyhe:
After upping the upper limit on how many subjects are returned from GET v1/subjects, the specified subject is showing up on https://refocus-sandbox.internal.salesforce.com/perspectives/SR_Grid. Both on create, and edit perspective

view/perspective/utils.js
```diff
@@ -171,10 +180,14 @@ function getConfig(values, key, value) {
   };
 
   if (key === 'subjects') {
-    config.placeholderText = 'Select a Subject...';
+    config.placeholderText = 'Enter a subject name';
```
annyhe: as per our conversation @iamigo 
screen shot 2017-12-20 at 2 41 36 pm

annyhe:
@iamigo are we ok to merge this?
Also, which steps do we need before deploying this into prod?

PR #705, opened 2017-12-12
Sample Generator validation - check user's write permission on Aspects

pallavi2209:


tests/api/v1/generators/post.js
```diff
@@ -121,4 +126,237 @@ describe('tests/api/v1/generators/post.js >', () => {
       });
     });
   });
+
+  describe('validateGeneratorAspectsPermissions >', () => {
```
iamigo: great set of tests!

PR #706, opened 2017-12-12
remove req.query from loadView

annyhe:
The original intention of having req.query in loadView.js was to load an unnamed perspective using the req.query. 
Currently any GET unnamed perspective will redirect to the default perspective. The loading unnamed perspective no longer functions.
Hence we can take out all references to req.query.


annyhe:
Nope. The queryParam are not in the tests. @iamigo

PR #707, opened 2017-12-13
add minimal logging to detect where the invalid sample comes from

annyhe:
- logging only. no throwing errors.

### DIFF
--- a/realtime/utils.js
+++ b/realtime/utils.js
@@ -423,12 +423,12 @@ function attachAspectSubject(_sample, useSampleStore, subjectModel,
   } else {
     const subOpts = {
       where: {
-        absolutePath: subName,
+        absolutePath: { $iLike: subName },
       },
     };
     const aspOpts = {
       where: {
-        name: aspName,
+        name: { $iLike: aspName },
       },
     };
     const getAspectPromise = aspectModel ? aspectModel.findOne(aspOpts) :
--- a/realtime/utils.js
+++ b/realtime/utils.js
@@ -390,16 +390,25 @@ function isIpWhitelisted(addr, whitelist) {
  */
 function attachAspectSubject(_sample, useSampleStore, subjectModel,
   aspectModel) {
-  const sample = _sample.get ? _sample.get() : _sample;
-  let nameParts;
+  let sample;
+  let source;
+  if (_sample.get) {
+    sample = _sample.get();
+    source = 'db';
+  } else {
+    sample = _sample;
+    source = 'redis';
+  }
 
   // check if sample object contains name
   if (!sample.name || sample.name.indexOf('|') < 0) {
-    logger.error('sample object does not contain name', sample);
+    logger.error('sample coming from ', source);
+    logger.error('sample object does not contain name', JSON.stringify(sample));
+    console.trace('from attachAspectSubject');
     return Promise.resolve(null);
   }
 
-  nameParts = sample.name.split('|');
+  let nameParts = sample.name.split('|');
   const subName = nameParts[0];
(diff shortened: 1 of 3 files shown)


realtime/utils.js
```diff
@@ -390,16 +390,25 @@ function isIpWhitelisted(addr, whitelist) {
...
+    sample = _sample;
+    source = 'redis';
+  }
 
   // check if sample object contains name
   if (!sample.name || sample.name.indexOf('|') < 0) {
-    logger.error('sample object does not contain name', sample);
+    logger.error('sample coming from ', source);
+    logger.error('sample object does not contain name', JSON.stringify(sample));
+    console.trace('from attachAspectSubject');
```
annyhe: added stack trace.

realtime/utils.js
```diff
@@ -390,16 +390,25 @@ function isIpWhitelisted(addr, whitelist) {
...
     return Promise.resolve(null);
   }
 
-  nameParts = sample.name.split('|');
+  let nameParts = sample.name.split('|');
```
iamigo: Hmmm. Just noticing down on line 426 that we're doing a where clause with absolutePath but I think that where clause should be using $ilike to find case insensitive match. Same for aspects right below that.

iamigo: @annyhe

annyhe: amended as per your comment.

annyhe:
Ready for you to review @harshkothari410

realtime/utils.js
```diff
@@ -423,12 +423,12 @@ function attachAspectSubject(_sample, useSampleStore, subjectModel,
   } else {
     const subOpts = {
       where: {
-        absolutePath: subName,
+        absolutePath: { $iLike: subName },
```
iamigo: need tests for these where clause changes to confirm case insensitive matches

tests/realtime/realtimeUtils.js
```diff
@@ -16,51 +16,51 @@ const tu = require('../testUtils');
...
+    name: tu.namePrefix + 'temperature',
+    timeout: '30s',
+    isPublished: true,
+    tags: ['temp'],
+  };
   const looksLikeSampleObjNA = {
     value: '10',
-    name: 'NA|temperature',
-    absolutePath: 'NA',
+    name: rootSubjNA + '|' + aspectOne.name,
```
annyhe: create aspect and subject to test query case insensitivity.
add test prefix to aspect and subject.

tests/realtime/utils.js
```diff
@@ -206,6 +206,22 @@ module.exports = {
     return JSON.parse(JSON.stringify(standardBotData));
   },
 
+  forceDeleteAspectsAndSubjects(done) {
+    tu.db.Event.destroy({
```
iamigo: Why Event here?

annyhe: following the convention. The line below at 226: forceDelete says to tu.db.Event...

tests/realtime/utils.js
```diff
@@ -206,32 +206,9 @@ module.exports = {
...
-      force: true,
-    })
-    .then(() => tu.forceDelete(tu.db.Subject, testStartTime))
-    .then(() => tu.forceDelete(tu.db.Aspect, testStartTime))
-    .then(() => done())
-    .catch(done);
-  },
-
   forceDelete(done) {
-    tu.db.Event.destroy({
```
iamigo: Probably should keep deleting Event, but use the tu.forceDelete helper fn instead of calling destroy explicitly here -- I imagine Tausif or someone from IMC added it for a reason

PR #708, opened 2017-12-14
BotData Shared Context

tausifmuzaffar:
We want to be able to sync data between bots when we specified its allowed to do it in the room settings.

The workflow is like this: Room Defines Bot1 must sync with Bot2 -> Bot1 BotData is created/updated -> we determine if we need to sync -> create/update Bot2 BotData


db/helpers/botDataUtils.js
```diff
@@ -0,0 +1,168 @@
...
+const ONE = 1;
+const TWO = 2;
+
+/**
+ * Check if a string is a JSON
+ *
+ * @param {String} item - String to check
+ * @returns {Boolean} - if String is JSON
+ */
+function isJson(item) {
```
iamigo: https://www.npmjs.com/package/is-json?

db/helpers/botDataUtils.js
```diff
@@ -0,0 +1,168 @@
...
+
+  if (typeof itemParse === 'object' && itemParse !== null) {
+    return true;
+  }
+
+  return false;
+} // isJson
+
+/**
+ * Check if a string is a JSON
```
iamigo: Wrong function comment for this function?

iamigo: This function needs some explanation and examples.

db/helpers/botDataUtils.js
```diff
@@ -0,0 +1,168 @@
...
+      outputValue = outputValue
+        .replace(replaceString, instance.value);
+    }
+  }
+
+  return outputValue;
+} // replaceValue
+
+/**
+ * Determine if bot data needs to synced between bots and perform the action.
```
iamigo: "needs to synced"??

db/helpers/botDataUtils.js
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * @param {Object} seq - The array to test
+ * @param {Object} instance - The array to test
```
iamigo: is "instance" really the array to test, again?

db/helpers/botDataUtils.js
```diff
@@ -0,0 +1,168 @@
...
+ * @returns {undefined} - OK
+ * @throws {validationError} - if the array does not contain valid attributes
+ */
+function updateValues(seq, instance) {
+  const Room = seq.models.Room;
+  const Bot = seq.models.Bot;
+  const BotData = seq.models.BotData;
+  const botNames = {};
+  const botsIds = {};
+  return Bot.findAll({
```
iamigo: Bot.findAll with no where clause? I don't understand.

db/helpers/botDataUtils.js
tausifmuzaffar: I tried this package but it doesnt seem to be as good at the code i wrote it return false for this string

{ response: 'this is a {$TestBot1.response$}' }

iamigo: Is `{$varName$}` a standard for doing variable replacement? Where is this used? Why not more standard `${varName}` or "template" style `{{varName}}`?

iamigo:
jscs lint failures

db/helpers/botDataUtils.js
tausifmuzaffar: Some of the bots already use variable placements, like comms bot and pagerduty use handlebars. So we needed something unquie that would conflict with things

iamigo: Could be confusing, hard for bot developers to grok non-standard syntax. What alternatives have you explored?

tausifmuzaffar: Ive decided to switch ${} since handlebars use {{}}; and I dont know another JS template that use it withouth ``

tausifmuzaffar:
@kamilsmuga @BrandonMurnane can yall review this?

PR #709, opened 2017-12-19
added rebuildSubjectHierarchy command

harshkothari410:


db/rebuildSubjectHierarchy.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
iamigo: 2017

db/rebuildSubjectHierarchy.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * db/rebuildSubjectHierarchy.js
+ */
+const utils = require('./utils.js');
+
+utils.doImport()
+.then(() => {
+  utils.seq.models.Subject.rebuildHierarchy()
+  .then(() => {
+    console.log('Successfully rebuild Subject Hierarchy.'); // eslint-disable-line no-console
```
iamigo: since this is printed upon completion it should be in past tense ("rebuilt" rather than "rebuild")

db/rebuildSubjectHierarchy.js
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) 2016, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * db/rebuildSubjectHierarchy.js
+ */
```
iamigo: Can you add some description here about why an admin would ever need to run this? Would it be possible to introspect the db and show the bad data it's going to repair? And have a "read-only" mode where it just shows you the bad data, but doesn't actually rebuild the hierarchy?

db/rebuildSubjectHierarchy.js
harshkothari410: True

db/rebuildSubjectHierarchy.js
harshkothari410: I need to find out that thing if we can find out with the script  about bad data.

harshkothari410:
@iamigo do we have any tests for npm commands ? If yes then I can refer them thats why.

iamigo:
I'm not so concerned about testing how the command is executed--I'd like to see a test where the "corrupt" subject/subjectancestors entries are "repaired"

harshkothari410:
@iamigo added test

PR #711, opened 2018-01-05
Subject caching

jgraff2:
This PR fixes issues with getting subjects from the cache so the behavior is exactly the same as from the db.
The core changes are in the "main fixes" commit. The other commits are for setting up the tests to run with the toggle enabled and making sure errors are consistent when coming from redis or the db.


package.json
```diff
@@ -46,9 +46,13 @@
...
     "test-cors": "CORS_ROUTES=/v1/bots,/v1/botActions,/v1/botData,/v1/events,/v1/rooms mocha -R dot --recursive tests/cors",
-    "test-cache": "mocha -R dot --recursive tests/cache",
+    "test-jobQueue": "mocha --recursive tests/jobQueue",
+    "test-cache": "npm run test-cache-jobQueue && npm run test-cache-models && npm run test-cache-misc",
+    "test-cache-jobQueue": "mocha --recursive tests/cache/jobQueue",
+    "test-cache-models": "mocha --recursive tests/cache/models",
+    "test-cache-misc": "mocha tests/cache",
     "test-view": "NODE_ENV=build mocha -R dot --recursive --compilers js:babel-core/register --require ./tests/view/setup.js tests/view",
-    "test": "istanbul cover ./node_modules/mocha/bin/_mocha --report lcovonly -- -R dot --recursive tests/api tests/cache tests/clock tests/config tests/db tests/enforceToken tests/jobQueue tests/realtime tests/enableCache tests/logging && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage && npm run test-view && npm run test-disablehttp && npm run test-rooms-enabled && npm run test-rooms-disabled && npm run test-cors && npm run test-limiter",
```
jgraff2: This is necessary because these scripts run tests from the same file, and you can't run tests twice in the same mocha process. See comments in test files.

api/v1/apiErrors.js
```diff
@@ -13,126 +13,139 @@
...
 // ----------------------------------------------------------------------------
 // Validation Errors
 // ----------------------------------------------------------------------------
-apiErrors.create({
+errors.create({
+  scope: exports,
   code: 11100,
   status: 400,
   name: 'ValidationError',
-  parent: apiErrors.RefocusApiError,
```
jgraff2: Previously, errors could be overwritten if another error was defined with the same name in a different file, because it defined the error on the global errors object, which is the same whether you require apiErrors, dbErrors, or redisErrors. This was causing discrepancies with error handling with the cache enabled vs disabled. This fixes it by restricting the scope to this module.

api/v1/errorHandler.js
```diff
@@ -122,7 +122,7 @@ module.exports = function errorHandler(err, req, res, next) {
       if (/Route defined in Swagger specification.*/.test(err.message)) {
         err.status = constants.httpStatus.NOT_ALLOWED;
       } else if (err.name === 'SequelizeUniqueConstraintError') {
-        err.status = constants.httpStatus.FORBIDDEN;
```
jgraff2: This commit changes the status for SequelizeUniqueConstraintErrors to 400 so it matches DuplicateResourceError. This is necessary so the same tests will pass when the cache is enabled or disabled.

annyhe:
darn this is a large PR. will start review tomorrow.

package.json
```diff
@@ -46,9 +46,13 @@
     "test-rooms-disabled": "ENABLE_ROOMS=false HIDDEN_ROUTES=/rooms mocha -R dot --recursive tests/roomFlag/roomFlagDisabled",
     "test-db": "npm run checkdb && mocha -R dot --recursive tests/db",
     "test-cors": "CORS_ROUTES=/v1/bots,/v1/botActions,/v1/botData,/v1/events,/v1/rooms mocha -R dot --recursive tests/cors",
-    "test-cache": "mocha -R dot --recursive tests/cache",
+    "test-jobQueue": "mocha --recursive tests/jobQueue",
+    "test-cache": "npm run test-cache-jobQueue && npm run test-cache-models && npm run test-cache-misc",
+    "test-cache-jobQueue": "mocha --recursive tests/cache/jobQueue",
+    "test-cache-models": "mocha --recursive tests/cache/models",
+    "test-cache-misc": "mocha tests/cache",
```
jgraff2: I used the default tree reporter here instead of the dot reporter because I needed to be able to see exactly which tests were being run to fix the issue where they were being skipped. (see below)

Is there a good reason to use the dot reporter? I think it would be more useful to use the default reporter for all the tests. If not, should I change these to use the dot reporter for consistency?

shriramshankar:
@jgraff2 The test coverage went down 11% in the PR.

jgraff2:
I moved the cache and job queue tests out of the main mocha command. This was necessary to run the same tests with different toggles enabled. I didn't realize that would also impact the test coverage.

I'll see if there's a way around this.

shriramshankar:
I thought this PR was to fix the performance degradation when getting subjects from the cache. Is it addressed in the "main fixes" commit?

jgraff2:
It was, originally. When I investigated, I found that the performance degradation was caused because certain requests were causing errors, and one of the integrations was resending the request continuously when it received a 400. The root of the problem was that the get-subjects-from-cache code wasn't working properly. 

As I was fixing it I found several other issues where getting subjects from the cache didn't work correctly, that weren't covered by the existing tests,
so the story turned into making sure the behavior was exactly the same with the toggle enabled or disabled. Part of that was setting up the standard tests to run with the toggle enabled - that's what the changes are for in package.json.

All the functional changes are in the main fixes commit.

shriramshankar:
@jgraff2 Makes sense now. Will take a close look at this later tonight.

package.json
iamigo: default reporter is fine ... I use it locally a lot cuz dot reporter tends to "hide" slow tests

api/v1/controllers/subjects.js
```diff
@@ -33,6 +33,7 @@ const sampleStoreConstants = sampleStore.constants;
 const jobType = require('../../../jobQueue/setup').jobType;
 const jobWrapper = require('../../../jobQueue/jobWrapper');
 const jobSetup = require('../../../jobQueue/setup');
+const common = require('../../../utils/common');
```
shriramshankar: look like this has not been used in this file.

cache/models/subject.js
```diff
@@ -20,6 +20,7 @@ const redisClient = require('../redisCache').client.sampleStore;
 const u = require('../../utils/filters');
 const modelUtils = require('./utils');
 const redisErrors = require('../redisErrors');
+const common = require('../../utils/common');
```
shriramshankar: looks like this has not been used.

api/v1/helpers/nouns/subjects.js
```diff
@@ -115,7 +115,11 @@ function traverseHierarchy(res) {
     }
 
     // filtered array is attached to the children key
-    res.children = filteredChildrenArr;
+    if (filteredChildrenArr.length) {
+      res.children = filteredChildrenArr;
+    } else {
+      res.children = undefined;
+    }
```
shriramshankar: consider using ternary operator?

cache/models/subject.js
```diff
@@ -175,7 +177,12 @@ function traverseHierarchy(res) {
       }
     }
 
-    res.children = filteredChildrenArr;
+    if (filteredChildrenArr.length) {
+      res.children = filteredChildrenArr;
+    } else {
+      res.children = undefined;
+    }
```
shriramshankar: consider using ternary operator?

cache/models/utils.js
```diff
@@ -39,34 +39,49 @@ function applyFieldListFilter(resource, attributes) {
...
   if (opts.order) {
-    const sortedResources = sortByOrder(filteredResources, opts.order);
-    filteredResources = sortedResources;
+    filtered = sortByOrder(filtered, opts.order);
+  }
 
-    const slicedResourceObjs = applyLimitAndOffset(opts, filteredResources);
-    filteredResources = slicedResourceObjs;
+  // apply limits to resources if not already applied in prefilterKeys
+  if (!applyLimitAndOffsetInPrefilter(opts)) {
```
shriramshankar: All the functions that are been used here, were used before it was declared. While I do not have any preference, we have been declaring a function before using it.

tests/cache/models/subjects/runStandardTests.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
```
shriramshankar: 2018

tests/cache/models/subjects/runStandardTests.js
```diff
@@ -0,0 +1,32 @@
...
+'use strict'; // eslint-disable-line strict
+const tu = require('../../../testUtils');
+const requireDir = require('require-dir');
+
+/*
+This test runs the standard subject api tests with the cache enabled.
+Note that this must be run in a separate command from the api tests,
+otherwise these tests will not run because files can't be required
+twice in the same process.
+*/
```
shriramshankar: /*
*
*/  
for multiline comments.

api/v1/apiErrors.js
shriramshankar: Nice.

package.json
jgraff2: Can I go ahead and change it for all the other tests?

cache/models/utils.js
jgraff2: Previously, applyLimitAndOffset was being called before it was declared. Is there something in the style guide about this? 

I definitely prefer being able to declare functions in any order.

package.json
```diff
@@ -46,13 +46,21 @@
...
+    "test-main": "npm run test-with-coverage -- --recursive tests/api tests/clock tests/config tests/db tests/enforceToken tests/realtime tests/enableCache tests/logging",
+    "test-extra": "npm run test-cache && npm run test-jobQueue && npm run test-view && npm run test-disablehttp && npm run test-rooms-enabled && npm run test-rooms-disabled && npm run test-cors && npm run test-limiter",
+    "test-all": "npm run test-main && npm run test-extra",
+    "test": "npm run reset-coverage && npm run test-all && npm run report",
+    "test-travis": "npm run test && npm run coveralls",
+    "test-with-coverage": "istanbul cover ./node_modules/mocha/bin/_mocha -report none --print none --include-pid --",
+    "report": "istanbul report text-summary lcovonly",
+    "reset-coverage": "rm -rf ./coverage",
+    "coveralls": "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && npm run reset-coverage",
```
jgraff2: I made some changes to the test scripts to get the coverage reporting working across multiple mocha runs. Coverage is back up to 88% now.

I could also apply these changes to the test scripts that previously weren't being included in the test coverage. (in test-extra) Any issue with that?

cache/models/utils.js
shriramshankar: Sorry, I did not notice your reply earlier. My linter complains about this every time. Looks like this is something to do with the `no-use-before-define` rule.

package.json
shriramshankar: No issues from my side.

iamigo: Please do it!!!

package.json
```diff
@@ -46,9 +46,21 @@
...
```
iamigo: trying to keep the list of scripts in alphabetical order for readability

api/v1/helpers/nouns/subjects.js
```diff
@@ -115,7 +115,7 @@ function traverseHierarchy(res) {
     }
 
     // filtered array is attached to the children key
-    res.children = filteredChildrenArr;
+    res.children = filteredChildrenArr.length ? filteredChildrenArr : undefined;
```
iamigo: is it better to make this undefined or an empty zero-length array?

cache/models/subject.js
```diff
@@ -175,7 +176,8 @@ function traverseHierarchy(res) {
       }
     }
 
-    res.children = filteredChildrenArr;
+    res.children = filteredChildrenArr.length ? filteredChildrenArr : undefined;
```
iamigo: is it better to make this undefined or an empty zero-length array?

cache/models/subject.js
```diff
@@ -214,22 +216,62 @@ function completeSubjectHierarchy(res, params) {
...
+  // convert the time fields to appropriate format
+  subject.createdAt = new Date(subject.createdAt).toISOString();
+  subject.updatedAt = new Date(subject.updatedAt).toISOString();
+
+  if (opts.attributes) { // delete subject fields
+    modelUtils.applyFieldListFilter(subject, opts.attributes);
+  }
+
+  // add api links
+  subject.apiLinks = utils.getApiLinks(
```
iamigo: the args should all fit on the same line as the fn call itself

cache/models/utils.js
iamigo: This would be an outlier in our codebase (except for some views) -- let's please stick with defining functions before using them.

package.json
jgraff2: @iamigo can you confirm?

api/v1/helpers/nouns/subjects.js
jgraff2: When the hierarchy comes from the db, subjects with no children have an undefined "children" attribute, instead of an empty array. This was changed for consistency with that.

iamigo:
one test failure @jgraff2

PR #713, opened 2018-01-08
migrations to change transform type from text to json

shriramshankar:


shriramshankar:
@iamigo Yes, I have manually tested both up and down migrations. Just deployed this branch to dev-box and it looks good too.

PR #715, opened 2018-01-10
Added user information to room from tracking object

BrandonMurnane:


kamilsmuga:
This `user` object is crucial to exist for the BDK, and Bots. Can we write a test that checks if this object is provided by the server? I'd like avoid a situation where server-side changes related to this specific tracking object break the whole BDK and Bots ecosystem.

PR #717, opened 2018-01-12
Don't upsert a sample if the aspect or subject is unpublished.

jgraff2:


jgraff2:
This was another case where there was a bug in the sample store that was covered in the standard tests but not in the cache-specific tests.

I just added the relevant tests to the existing cache tests for now, but I also created W-4616674 to run all the standard tests with the cache enabled, like in the other recent PR for subject caching.

PR #723, opened 2018-01-26
Collector status in heartbeat

shriramshankar:


tests/api/v1/collectors/heartbeat.js
```diff
@@ -230,569 +233,569 @@ describe('tests/api/v1/collectors/heartbeat.js >', () => {
...
       // reset the tracked changes
       beforeEach((done) => {
-        Promise.resolve()
-        .then(() => sendHeartbeat(collector1))
-        .then(() => sendHeartbeat(collector2))
-        .then(() => sendHeartbeat(collector3))
-        .then(() => done()).catch(done);
+        u.sendHeartbeat(collector1, collectorTokens)
+        .then(() => u.sendHeartbeat(collector2, collectorTokens))
+        .then(() => u.sendHeartbeat(collector3, collectorTokens))
```
jgraff2: I much prefer starting the chain with Promise.resolve() because it makes it more readable and easier to edit.

shriramshankar: Readable ? easier to edit? Why would you create a new promise and not use it? It would be very confusing for someone looking at these tests later.

jgraff2: The promise is being used, it's used to start the chain so the first statement can be wrapped in a then.
Compare these:

u.postGenerator(generator1, userToken, [collector1])
.then(() => u.sendHeartbeat(collector1, collectorTokens))
.then(() => u.patchGenerator(generator1, userToken))
.then(() => u.sendHeartbeat(collector1, collectorTokens))
.then((res) => u.expectLengths({ added: 0, deleted: 0, updated: 1 },
res))

Promise.resolve()
.then(() => u.postGenerator(generator1, userToken, [collector1]))
.then(() => u.sendHeartbeat(collector1, collectorTokens))
.then(() => u.patchGenerator(generator1, userToken))
.then(() => u.sendHeartbeat(collector1, collectorTokens))
.then((res) => u.expectLengths({ added: 0, deleted: 0, updated: 1 },
res))

In the second form, the statements are lined up and consistently formatted. It's easy to see at a glance that the sequence is post, send, patch, send, expect.

And if I want to change the order, I can just copy an entire line and move it. With the first form, if I moved the first line I would also have to wrap it in a then and then unwrap the new first line:

.then(() => u.patchGenerator(generator1, userToken))
.then(() => u.sendHeartbeat(collector1, collectorTokens))
u.postGenerator(generator1, userToken, [collector1])
.then(() => u.sendHeartbeat(collector1, collectorTokens))
.then((res) => u.expectLengths({ added: 0, deleted: 0, updated: 1 },
res))

The tests for the generator changes are all about setting up different permutations of operations. Making it easy to change the order is really important.

shriramshankar: Yes, moving the lines around is easier this way. The first line `postGenerator ` in this example is already a promise and you are wrapping it again with a promise. As I mentioned in my earlier comment, it can get confusing for someone looking at these tests later. When I looked at these tests I was not sure why this was done.

iamigo: I like the readability of the way Jon had it organized, so maybe just add comments explaining why it's structured this way... ?

PR #724, opened 2018-01-29
Run sample tests with cache enabled

jgraff2:


tests/api/v1/samples/put.js
```diff
@@ -39,25 +42,29 @@ describe(`tests/api/v1/samples/put.js, PUT ${path} >`, () => {
...
+      sampleName1 = samp.name;
+      subjectId1 = samp.subjectId;
+      aspectId1 = samp.aspectId;
       return u.doCustomSetup('TEA', 'TOFFEE');
     })
     .then((samp) => Sample.create(samp))
     .then((samp) => {
-      subjectId = samp.subjectId;
-      aspectId = samp.aspectId;
+      sampleName2 = samp.name;
```
shriramshankar: You are not using sampleName2 anywhere else other than L25 and L52.

tests/api/v1/samples/getWithLimit.js
```diff
@@ -15,6 +15,10 @@ const u = require('./utils');
 const getWithLimit = require('../common/getWithLimit.js').getWithLimit;
 const Sample = tu.db.Sample;
 const path = '/v1/samples';
+const supertest = require('supertest');
+const api = supertest(require('../../../../index').app);
+const constants = require('../../../../api/v1/constants');
+const expect = require('chai').expect;
```
shriramshankar: Looks like L18 to L21 is not required for these tests to run.

jgraff2: Good catch. I had set these up to debug some errors I was getting and I forgot to remove them.

tests/api/v1/samples/put.js
jgraff2: I added it for consistency. I needed to access both aspect1 and aspect2, and it's clearer to call it sampleName1 and sampleName2 than to just have one sampleName.

PR #725, opened 2018-02-06
remove "publishPartialSample" short term toggle

iamigo:


realtime/utils.js
```diff
@@ -471,6 +461,5 @@ module.exports = {
   isIpWhitelisted,
   parseObject,
   shouldIEmitThisObj,
-  isThisSample,
```
shriramshankar: I think we do not need the `attachAspectSubject` function too.

iamigo: attachAspectSubject is used a lot in cache/model/samples

iamigo: It's used by redisPublisher.pubishSample

api/v1/controllers/samples.js
```diff
@@ -422,17 +416,9 @@ module.exports = {
...
-                publisher.publishPartialSample(sample);
-              } else {
-                publisher.publishSample(sample, subHelper.model);
-              }
-            }
-          });
-        });
+        .then((samples) =>
+          samples.filter((s) => !s.isFailed)
+          .forEach((s) => publisher.publishSample(s, subHelper.model)));
```
shriramshankar: We could potentially have problems with this when the number of samples from `bulkupsert` is a lot. I think here, we are iterating over the array twice and creating also a copy of it.

realtime/utils.js
shriramshankar: Ok, got it.

PR #728, opened 2018-02-08
Redo collector start endpoint

shriramshankar:
The `startCollector` function in `api/v1/controllers/collectors.js` has been fully redone (except adding the generators to the response).

### DIFF
--- a/api/v1/controllers/collectors.js
+++ b/api/v1/controllers/collectors.js
@@ -317,8 +317,7 @@ function heartbeat(req, res, next) {
  * POST /collectors/start
  *
  * Starts a collector by setting the status to "Running". If the collector is
- * not found, a new collector is created. A collector token is sent back as a
- * response.
+ * not found, a new collector is created.
  * @param {IncomingMessage} req - The request object
  * @param {ServerResponse} res - The response object
  * @param {Function} next - The next middleware function in the stack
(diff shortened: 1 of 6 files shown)


api/v1/swagger.yaml
```diff
@@ -11274,16 +11277,16 @@ definitions:
...
+        $ref: "#/definitions/CollectorConfigResponse"
       generatersAdded:
         $ref: "#/definitions/GeneratorsAddedResponse"
       generatersUpdated:
         $ref: "#/definitions/GeneratorsUpdatedResponse"
       generatorsRemoved:
         $ref: "#/definitions/GeneratorsRemovedResponse"
 
-  CollectorConfigUpdatesResponse:
+  CollectorConfigResponse:
```
shriramshankar: The swagger definition of collectorConfig includes many properties which are probably not needed. We can talk about it tomorrow and remove the ones that not required.

api/v1/controllers/collectors.js
```diff
@@ -329,73 +314,67 @@ function heartbeat(req, res, next) {
...
 
-    if (returnedCollector.status === 'Running' ||
-      returnedCollector.status === 'Paused') {
-      throw new apiErrors.ForbiddenError({ explanation:
-        'Cannot start--only stopped collectors can start.',
-      });
-    }
+        if (collector.status === 'Running' || collector.status === 'Paused') {
+          throw new apiErrors.ForbiddenError({ explanation:
+            'Cannot start--this collector is not registered.',
```
iamigo: This explanation doesn't appear to match the if(...)

tests/api/v1/collectors/utils.js
```diff
@@ -44,9 +44,9 @@ function startCollector(collector, collectorTokens, userToken) {
   });
 }
 
-function stopCollector(collector, collectorTokens) {
+function stopCollector(collector, userToken) {
```
iamigo: thx for renaming all these token vars!

PR #732, opened 2018-02-12
Added external id to room

BrandonMurnane:
added external id migration aswell as tests

### DIFF
--- a/migrations/20180212145410-add-external-id.js
+++ b/migrations/20180212145410-add-external-id.js
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2017, salesforce.com, inc.
+ * Copyright (c) 2018, salesforce.com, inc.
  * All rights reserved.
  * Licensed under the BSD 3-Clause license.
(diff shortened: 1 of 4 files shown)


migrations/20180212145410-add-external-id.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
```
iamigo: 2018

migrations/20180212145410-add-external-id.js
```diff
@@ -0,0 +1,22 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+'use strict'; // eslint-disable-line strict
+
+module.exports = {
+  up: function (qi, Sequelize) {
```
iamigo: don't need to do it here but when there's more than one step in the migration, please wrap it in a transaction

PR #733, opened 2018-02-13
track generator changes in redis instead of memory.

jgraff2:


api/v1/controllers/collectors.js
```diff
@@ -285,6 +283,24 @@ function heartbeat(req, res, next) {
     return o.save();
   })
 
+  // get the ids for changed generators
+  .then(() => heartbeatUtils.getChanges(collectorName))
+
+  // find the generator objects for the changes ids
+  .then((changed) => Promise.join(
+      Generator.findAll({ where: { id: changed.added } }),
```
iamigo: why three separate db queries here?

jgraff2: I need to be able to distinguish between the added, deleted, and updated generators sent back from the db.

iamigo: So your assumption is that three simultaneous db queries is "cheaper" than using one query and then iterating over the results to assign each result to the right set based on the id?

jgraff2: No, just that it's cleaner this way. I can do that if you're concerned about the performance.

iamigo: I'm okay with doing it either way, but please document your reasoning in a code comment.

jgraff2: Note that in practice it will be rare for multiple generators to be updated in different ways for the same collector within the same heartbeat cycle so most of the time there will only be one query sent anyway.

api/v1/controllers/collectors.js
```diff
@@ -285,6 +283,24 @@ function heartbeat(req, res, next) {
...
+      Generator.findAll({ where: { id: changed.deleted } }),
+      Generator.findAll({ where: { id: changed.updated } }),
```
shriramshankar: Is changed.[added/updated/deleted] a list of ids?

config/redisConfig.js
```diff
@@ -29,6 +29,12 @@ module.exports = {
     cache: pe.REDIS_CACHE && pe[pe.REDIS_CACHE] ?
       pe[pe.REDIS_CACHE] : PRIMARY_REDIS,
 
+    /*
+     * Track generator changes to be sent in collector heartbeat response.
+     */
+    heartbeat: pe.REDIS_HEARTBEAT && pe[pe.REDIS_HEARTBEAT] ?
+      pe[pe.REDIS_HEARTBEAT] : PRIMARY_REDIS,
+
```
shriramshankar: Do we want to use a separate instance just to track the collectors?

iamigo: We don't need to provision a separate instance for it, but I think it makes sense to allow an admin to assign it where they want.

api/v1/helpers/verbs/heartbeatUtils.js
```diff
@@ -11,20 +11,15 @@
...
-];
-const ADDED = 1;
-const DELETED = 2;
-const UPDATED = 3;
-const NOCHANGE = 4;
+const redisClient = require('../../../../cache/redisCache').client.heartbeat;
+const ADDED = 'added';
+const DELETED = 'deleted';
+const UPDATED = 'updated';
+const NOCHANGE = 0;
```
shriramshankar: When we inspect the redis instance, the key of the unchanged generators (for collector1)will look like `heartbeat::generatorChanges::collector1::0`. Can we change it to something more meaningful?

api/v1/helpers/verbs/heartbeatUtils.js
```diff
@@ -35,76 +30,106 @@ const NOCHANGE = 4;
...
+      // update redis
+      promises.push(removeFromSet(collectorName, undoChange, genId));
+      promises.push(addToSet(collectorName, change, genId));
+    });
 
+    return Promise.all(promises);
   });
+} // trackGeneratorChanges
+
+function getChanges(collectorName) {
```
shriramshankar: Can we please add comments to the functions that do not have them and declare them before they are used?

api/v1/controllers/collectors.js
jgraff2: Yes. I will change it to make it more clear.

api/v1/helpers/verbs/heartbeatUtils.js
jgraff2: Redis will only be updated when there is a change. NOCHANGE is only there to make the logic clearer.

api/v1/controllers/collectors.js
```diff
@@ -284,20 +284,24 @@ function heartbeat(req, res, next) {
...
-      }
+  .then(() => heartbeatUtils.getChangedIds(collectorName))
+
+  // find the generator objects for the changed ids
+  .then((changedIds) => Promise.all(
+    ['added', 'deleted', 'updated'].map((changeType) => {
+      const ids = changedIds[changeType];
+      const where = { where: { id: ids } };
+      return ids.length ? Generator.findAll(where) : Promise.resolve([]);
+    })
```
jgraff2: Changed to only query the db when there are changes.

api/v1/helpers/verbs/heartbeatUtils.js
shriramshankar: Ok. So we will never see a key like `heartbeat::generatorChanges::collector1::0` in redis, correct?

jgraff2: correct.

api/v1/helpers/verbs/heartbeatUtils.js
```diff
@@ -17,6 +17,91 @@ const DELETED = 'deleted';
...
+/**
+ * Get the changes from redis for a given collector
+ *
+ * @param {String} collectorName - The collector name
+ * @returns {Object} - An object with lists of added, deleted, and updated
+ * generator ids
+ */
+function getChangedIds(collectorName) {
+  return new Promise((resolve, reject) =>
+    redisClient.multi()
```
iamigo: why callback over promise here?

jgraff2: Good point. Updated.

PR #735, opened 2018-02-14
Add generators start endpoint

shriramshankar:


api/v1/swagger.yaml
```diff
@@ -11267,8 +11267,9 @@ definitions:
           The ipAddress where the collector is installed.
         maxLength: 60
       collectorConfig:
-        type: object
         $ref: "#/definitions/CollectorConfigResponse"
+      generatersAdded:
```
iamigo: "generatersAdded" should be "generatorsAdded"

tests/api/v1/collectors/get.js
```diff
@@ -75,7 +75,7 @@ describe('tests/api/v1/collectors/get.js >', () => {
   });
 
   it('get with limit ok', (done) => {
-    const c2 = u.toCreate;
+    const c2 = JSON.parse(JSON.stringify(u.toCreate));
```
iamigo: I think it's cleaner to put the parse+stringify logic in the utils module in a "getCollectorToCreate()" fn--that way every test which uses it doesn't have to repeat the parse+stringify pattern

tests/db/model/collector/create.js
```diff
@@ -19,11 +19,11 @@ describe('tests/db/model/collector/create.js >', () => {
   let userId;
   let collectorObj;
   beforeEach((done) => {
+    collectorObj = JSON.parse(JSON.stringify(u.getCollectorObj()));
```
iamigo: same as comment above

tests/db/model/collector/find.js
```diff
@@ -56,8 +88,32 @@ describe('tests/db/model/collector/find.js >', () => {
     .catch(done);
   });
 
+  it('Collector Instnace with related generators', (done) => {
```
jgraff2: typo

tests/db/model/collector/find.js
```diff
@@ -56,8 +88,32 @@ describe('tests/db/model/collector/find.js >', () => {
...
+      const sg2 = generators.filter((gen) =>
+        gen.dataValues.name === generator2.name)[0].dataValues;
+      expect(sg1.id).to.include(generator1.id);
+      expect(sg2.id).to.include(generator2.id);
+      done();
+    })
+    .catch(done);
+  });
+
+  it('Collector Instnace without generators ', (done) => {
```
jgraff2: typo

PR #739, opened 2018-02-26
Check for sample missing name

iamigo:
- api/v1/controllers/samples.js
  - add logging to sample find/post/patch/put/upsert to track invalid sample name problems, e.g. missing name
- api/v1/helpers/verbs/utils.js
  - noop (comments and formatting changes only)
- cache/models/samples.js
  - consolidate all the sample name splitting and validation into helper fn "parseName"
  - call parseName where we used to split and validate the sample name, and call it again on the sample we're returning *after* performing the operation (upsertOneSample/patchSample/putSample/getSample/findSamples)
  - checkWritePermission does not need sample arg so simplified the fn signature and its callers
  - add more details to error explanations, e.g. the subject/aspect/sample name or id
- tests/cache/jobQueue/bulkUpsert.js
  - comments and formatting changes only
- tests/cache/models/samples/get.js
  - updated to match more detailed error message
- tests/cache/models/samples/post.js
  - updated to match more detailed error message
- tests/cache/models/samples/upsert.js
  - updated to match more detailed error message
- utils/common.js
  - comments and logging changes only

### DIFF
--- a/cache/models/samples.js
+++ b/cache/models/samples.js
@@ -205,7 +205,7 @@ function handleUpsertError(objectType, isBulk) {
   }
 
   throw err;
-}
+} // handleUpsertError
 
 /**
  * Upsert a sample. If subject exists, get aspect and sample. If aspect exists,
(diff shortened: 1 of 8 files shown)


cache/models/samples.js
```diff
@@ -56,26 +58,54 @@ const ONE = 1;
...
+    aspect: { name: undefined },
+  };
+  const arr = (name || '').split(sampleNameSeparator);
+  if (arr.length === 2) {
+    retval.subject.absolutePath = arr[0];
+    retval.aspect.name = arr[1];
+    return retval;
+  }
+
+  logger.error(`cache/models/samples.parseName|Invalid sample name "${name}"`);
```
shriramshankar: Do we also want to log the function from which the error was logged to identify the action (`/post/put / patch / delete`) that caused this to happen?

iamigo: That will get logged from the extra logging in the api controller

PR #740, opened 2018-02-28
Deactivating Rooms

kriscfoster:
- Added a toggle that can deactivate rooms form the UI.

screen shot 2018-02-28 at 08 08 00
screen shot 2018-02-28 at 08 08 09


view/rooms/app.js
```diff
@@ -258,6 +265,13 @@ function setupSocketIOClient(bots) {
           detail: events,
         }));
       });
+
+      if (events.context.type === 'RoomState') {
```
kamilsmuga: check if `events.context` exist before extracting `type` value to prevent `Can't ready property type of undefined` errors

view/rooms/app.js
```diff
@@ -258,6 +265,13 @@ function setupSocketIOClient(bots) {
...
+        document.getElementById('activeToggle')
```
kamilsmuga: can you reuse the existing `activeToggle` const reference instead of doing document.getElementId every time?

view/rooms/app.js
```diff
@@ -266,7 +280,85 @@ function setupSocketIOClient(bots) {
...
+    };
+
+    const events = {
+      log: message,
+      context: eventType,
+      userId: _user.id,
+      roomId: parseInt(ROOM_ID, 10)
+    };
+
+    return u.postPromiseWithUrl('/v1/events', events);
```
kamilsmuga: can you define events path as a constant on top of this file? similar to the existing room and bots routes

iamigo:
Also, what does it mean that a room is "Active"? 
And the "Confirmation Required" dialog asks "Would you like to close this room?" -- is that what comes up when you toggle that new "Active" toggle? Is the toggle "Active/Closed"? That would be weird.

kriscfoster:
@iamigo I will write these tests. When an incident/room is opened first it will be active. When the incident is finished the room will be deactivated/closed from this toggle on the UI. This will just let people in the room know that the incident is finished.

iamigo:
Seems like a lot of confusing language here.

1. `When an incident/room is opened...` - Mixing "incident" with "room" -- which is it? Are they the same thing? Are we calling it one thing here and another thing somewhere else? What do we want end-users calling it?
2. `When an incident/room is opened first it will be active.` - so "opened" === "active"? Need to pick the right work and stick with it everywhere, otherwise too confusing for everyone.
3. `When the incident is finished the room will be deactivated/closed` - so "incident is finished" === "room is closed" === "room is deactivated"?

PR #744, opened 2018-03-08
Fix put behavior

jgraff2:
While working on PR 745, I realized that the PUT behavior is incorrect. It behaves like a PATCH - it only updates the fields that were specified in the request, and does not update blank fields to null/false. These are the changes required to fix it and get the tests passing.

This PR should be merged before the other one.

### DIFF
--- a/api/v1/helpers/verbs/utils.js
+++ b/api/v1/helpers/verbs/utils.js
@@ -59,6 +59,7 @@ function updateInstance(o, puttableFields, toPut) {
 
 /**
  * Sorts the array field of an object or an array of objects alphabetically.
+ *
  * @param {Object} props - The helpers/nouns module for the given DB model
  * @param  {Object|Array} instArrayOrObject  - The instance object or an array
  *   of instance object with fields containing an array of objects that needs
@@ -80,8 +81,7 @@ function sortArrayObjectsByField(props, instArrayOrObject) {
 } // sortArrayObjectsByField
 
 /**
- * Sends the udpated record back in the json response
- * with status code 200.
+ * Sends the udpated record back in the json response with status code 200.
  *
  * @param {Object} resultObj - For logging
  * @param {Object} req - The request object
@@ -92,13 +92,12 @@ function sortArrayObjectsByField(props, instArrayOrObject) {
  */
 function handleUpdatePromise(resultObj, req, retVal, props, res) {
   const returnObj = retVal.get ? retVal.get() : retVal;
-
   sortArrayObjectsByField(props, returnObj);
 
   // publish the update event to the redis channel
   if (props.publishEvents) {
-    publisher.publishSample(
-      returnObj, props.associatedModels.subject, realtimeEvents.sample.upd);
+    publisher.publishSample(returnObj, props.associatedModels.subject,
+      realtimeEvents.sample.upd);
   }
 
   // update the cache
@@ -111,10 +110,9 @@ function handleUpdatePromise(resultObj, req, retVal, props, res) {
 
   resultObj.dbTime = new Date() - resultObj.reqStartTime;
   logAPI(req, resultObj, returnObj);
-
   return res.status(constants.httpStatus.OK)
     .json(responsify(returnObj, props, req.method));
-}
+} // handleUpdatePromise
 
 /**
  * In-place removal of certain keys from the input object
--- a/api/v1/helpers/verbs/utils.js
+++ b/api/v1/helpers/verbs/utils.js
@@ -41,13 +41,11 @@ function updateInstance(o, puttableFields, toPut) {
       }
 
       o.set(key, nullish);
-
-      // take nullified fields out of changed fields
-      o.changed(key, false);
     } else {
       /*
-       * value may have changed. set changed to true to
-       * trigger checks in the model
+       * value may be set but not changed. set changed to true to
+       * trigger checks in the model. This is the only way to differentiate between
+       * PUT and PATCH in certain cases.
        */
       o.changed(key, true);
       o.set(key, toPut[key]);
--- a/db/model/subject.js
+++ b/db/model/subject.js
@@ -519,13 +519,6 @@ module.exports = function subject(seq, dataTypes) {
           inst.parentAbsolutePath == false;
         const pidEmpty = inst.parentId == null || inst.parentId == false;
 
-        console.log('parentId', inst.parentId)
-        console.log('previous parentId', inst._previousDataValues['parentId'])
-        console.log('papChanged', papChanged)
-        console.log('pidChanged', pidChanged)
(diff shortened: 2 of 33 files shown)


api/v1/helpers/verbs/utils.js
```diff
@@ -41,13 +41,11 @@ function updateInstance(o, puttableFields, toPut) {
       }
 
       o.set(key, nullish);
-
-      // take nullified fields out of changed fields
-      o.changed(key, false);
```
jgraff2: This is the line that breaks the PUT behavior. The DB will only update fields with changed=true, so setting it to false for missing fields means fields that are left blank in the request will not be updated to null/false. This makes the PUT behave like a PATCH, only updating the fields that were specified and ignoring all others.

This line was added about a year ago. I think it was added to get tests passing without realizing the implications.

Can you confirm this is not the intended behavior?

PUT has been acting like PATCH for the last year. We should make sure no one is relying on this before we change it back. (I believe the original bug that triggered this investigation was caused by a request that was relying on PUT acting like a PATCH)

api/v1/helpers/verbs/utils.js
```diff
@@ -41,13 +41,11 @@ function updateInstance(o, puttableFields, toPut) {
...
     } else {
       /*
-       * value may have changed. set changed to true to
-       * trigger checks in the model
+       * value may be set but not changed. set changed to true to
+       * trigger checks in the model. This is the only way to differentiate between
+       * PUT and PATCH in certain cases.
        */
       o.changed(key, true);
```
jgraff2: I spent a lot of time trying to get it working without this line too, but wasn't able to. 

There are certain cases where a PUT and a PATCH can have the same "changed" values, but the behavior should be different. Setting changed to true here allows us to differentiate them. I don't think there's any way around this as long as we are using the same logic in the model update hooks for both PUT and PATCH.

Leaving in this line should be fine, since it's setting it to true. This just means it will always be updated in the db even if the value is the same.

api/v1/helpers/verbs/utils.js
iamigo: @jgraff2 FTW!!!!!! THIS IS HUGE!!!!
Almost all the PUTs in our logs from the last 30 days are to /v1/aspects, from the integration with the most "sample missing name" problems. @shriramshankar 
Just make sure this doesn't break anything with PUT/PATCH lens and bot (which are different cuz of file upload).

db/model/subject.js
```diff
@@ -249,52 +249,7 @@ module.exports = function subject(seq, dataTypes) {
...
-              }
-
-              resolve(inst);
-            })
-            .catch((err) => reject(err));
-          });
-        }
-
-        inst.setDataValue('absolutePath', inst.name);
+        return inst.setupParentFields();
```
iamigo: yessss!

tests/api/v1/generators/put.js
```diff
@@ -69,7 +69,7 @@ describe('tests/api/v1/generators/put.js >', () => {
           description: 'An ok sample\'s value, e.g. \'0\'',
         },
       },
-      subjects: ['US'],
+      subjectQuery: '?subjects',
```
iamigo: `?subjects` doesn't seem like a valid subjectQuery... shouldn't it be something like `?absolutePath=Foo.*` or `?tags=Bar`

api/v1/helpers/verbs/utils.js
shriramshankar: Does it mean this was causing the "sample with missing name" problem?

iamigo: I'm not sure I can claim that with any confidence, but it is surely related.

tests/api/v1/generators/put.js
jgraff2: I was just using the fields that were already defined in utils.getGenerator. I will change it.

api/v1/helpers/verbs/utils.js
jgraff2: PUT for lens and bots doesn't use doPut, or call updateInstance directly, so this won't have any effect on those.

db/model/subject.js
```diff
@@ -720,6 +675,53 @@ module.exports = function subject(seq, dataTypes) {
...
+            key = 'absolutePath';
+            value = this.getDataValue('parentAbsolutePath');
+          }
+
+          const whereObj = { where: {} };
+          if (key === 'absolutePath') {
+            whereObj.where[key] = { $iLike: value };
+          } else {
+            whereObj.where[key] = value;
+          }
```
shriramshankar: The scopes `id` and `absolutePath` were created to handle L691 to L697. Can you use them instead?

db/model/subject.js
```diff
@@ -558,30 +513,30 @@ module.exports = function subject(seq, dataTypes) {
           }
         }
 
+        const papChanged = inst.changed('parentAbsolutePath');
+        const pidChanged = inst.changed('parentId');
+        const papEmpty = inst.parentAbsolutePath == null ||
+          inst.parentAbsolutePath == false;
+        const pidEmpty = inst.parentId == null || inst.parentId == false;
```
shriramshankar: Lines 516 to 520 and Lines 531 to 535 looks the same to me. Do you need both?

iamigo:
@jgraff2 - did you see Shriram's comments?

db/model/subject.js
jgraff2: It was using those originally - I changed it because there was an issue with case insensitivity. 
I changed it back and updated the scope to be case insensitive instead. Any objection to having the scope be case-insensitive for all queries?

db/model/subject.js
jgraff2: Yeah, they are both necessary because setupParentFields can change the values and that needs to be reflected in these variables for the logic to be correct.

db/model/subject.js
jgraff2: @shriramshankar @iamigo

db/model/subject.js
```diff
@@ -187,7 +187,7 @@ module.exports = function subject(seq, dataTypes) {
         }));
         Subject.addScope('absolutePath', (value) => ({
           where: {
-            absolutePath: value,
+            absolutePath: { $iLike: value },
```
jgraff2: I updated the scope to be case insensitive. Any objection to having this scope be case-insensitive for all queries?

db/model/subject.js
shriramshankar: Makes sense using the case-insensitive comparison.

db/model/subject.js
iamigo: no objection from me--doesn't break any existing tests, right?

jgraff2: right.

PR #745, opened 2018-03-08
Fix put aspect is published

jgraff2:
These are the fixes for W-4645848, which was a bug where aspects were not getting updated with PUT requests. The root cause was that the “changed” property for updated fields was being set artificially in the doPut, but the aspect hook logic relied on the real “changed” values.  I updated it to use the actual changed values, and added tests for several related bugs that were happening before this fix.

While working on this I found a bigger issue with PUTs and fixed it in PR 744. That one should be merged first - there is a test that won't pass until it is merged.


tests/api/v1/aspects/put.js
```diff
@@ -435,4 +440,160 @@ describe('tests/api/v1/aspects/put.js >', () => {
...
+      .expect(constants.httpStatus.OK)
+      .endAsync()
+      .then((res) => expectInResponse(res, expected))
+      .then(() => expectInDB(expected))
+      .then(() => expectInSampleEmbed(0))
+      .then(() => done())
+      .catch(done);
+    });
+
+    it('updating aspect with isPublished missing deletes its samples,',
```
jgraff2: This is the case that caused the original bug. It was returning the correct response but not actually updating it in the db.

tests/cache/models/aspects/runStandardTests.js
```diff
@@ -0,0 +1,30 @@
...
+ * twice in the same process.
+ */
+
+describe('tests/cache/models/aspects/runStandardTests.js, ', () => {
+  before(() => tu.toggleOverride('enableRedisSampleStore', true));
+  after(() => tu.toggleOverride('enableRedisSampleStore', false));
+
+  describe('run standard tests with cache enabled', () => {
+    requireDir('../../../api/v1/aspects');
+  });
```
jgraff2: Some of the related bugs I found with updating aspects were cache-specific, and weren’t being covered in the existing tests.
This is to set up the aspect tests to run with the cache enabled, like I did recently with subjects and samples. 
 I thought this wouldn’t be necessary before because all the tests were passing, but I realized it was because all the sample access in the aspect tests was going directly to the db. I changed it to use the api instead for all sample access.

iamigo:
@jgraff2 small merge conflict?

iamigo:
@jgraff2 one test failing?

jgraff2:
The failing test should pass once the other PR is merged.

db/model/aspect.js
```diff
@@ -318,17 +322,14 @@ module.exports = function aspect(seq, dataTypes) {
...
+            .forEach(key => instChanged[key] = inst[key]);
             redisOps.hmSet(aspectType, inst.name, instChanged);
           }
         }
 
-        if (inst.changed('isPublished') &&
-          inst.previous('isPublished') &&
-          !inst.getDataValue('isPublished') ||
-          inst.changed('name')) {
+        if (isPublishedChanged && !inst.isPublished || nameChanged) {
```
iamigo: Generally I find lines like this easier to read and understand if they're written like this (explicit grouping), and we only need to evaluate the AND'ed expressions if the OR expression is true, so put that first?

`if (nameChanged || (isPublishedChanged && !inst.isPublished)) {`

PR #746, opened 2018-03-09
Upsert Bot Data

tausifmuzaffar:
Upsert BotData where if values is json merge the new json with the old json


api/v1/swagger.yaml
```diff
@@ -1796,6 +1796,68 @@ paths:
...
 
+  # ---------------------------------------------------------------------------
+  /botData/upsert:
+    x-swagger-router-controller: botData
+    post:
+      security:
+        - jwt: []
+      summary: Upsert botData
+      tags: [ botData ]
+      description: Creates or updates bot data for a rooms specific bots
```
iamigo: room apostrophe s

db/helpers/botDataUtils.js
```diff
@@ -39,6 +39,28 @@ function isJson(item) {
...
+ *
+ * @param {String} originalValue - Original bot data value
+ * @param {String} newValue - New bot data value
+ * @returns {String} - Bot data where values are properly merged
+ */
+function combineValue(originalValue, newValue) {
+  let output = newValue;
+
+  // If values are json, then merge values
+  if ((isJson(originalValue)) && (isJson(newValue))) {
```
iamigo: Need to be pretty explicit about what gets overwritten, e.g. if original is { "a": [1, 2, 3] } and I send { "a": [2, 3, 4] } then, the way this is written, the new value is [2, 3, 4], i.e. it isn't *merging* arrays, it's overwriting them.

tests/api/v1/botData/upsert.js
```diff
@@ -0,0 +1,199 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
```
iamigo: New file -- 2018

api/v1/controllers/botData.js
```diff
@@ -13,12 +13,14 @@
...
 const helper = require('../helpers/nouns/botData');
 const Bot = require('../../../db').Bot;
+const BotData = require('../../../db').BotData;
 const doPost = require('../helpers/verbs/doPost');
 const doFind = require('../helpers/verbs/doFind');
 const doGet = require('../helpers/verbs/doGet');
 const doPatch = require('../helpers/verbs/doPatch');
 const doDelete = require('../helpers/verbs/doDelete');
 const u = require('../../../utils/common');
+const bdUtils = require('../../../db/helpers/botDataUtils');
```
iamigo: I don't *love* calling into a db helper module from api layer.
That being said, I don't want you to have to jump through hoops to move stuff and I don't want you to duplicate code.
Use your best judgement :)

api/v1/swagger.yaml
```diff
@@ -1796,6 +1796,68 @@ paths:
...
+              - value
+              - roomId
+              - botId
+      responses:
+        201:
+          description: >-
+            Bot Data Created
+          schema:
+            $ref: "#/definitions/BotDataResponse"
+        200:
```
iamigo: Nice use of response codes for the upsert :)

db/helpers/botDataUtils.js
```diff
@@ -39,6 +39,30 @@ function isJson(item) {
...
+ * JSON. If either parameter is not JSON strings we replace originalValue
+ * with the newValue. If both values are JSON strings we extend the
+ * originalValue by adding any missing key-value pairs and replacing any
+ * exsiting key-value pairs with new value
+ *
+ * @param {String} originalValue - Original bot data value
+ * @param {String} newValue - New bot data value
+ * @returns {String} - Bot data where values are properly merged
+ */
+function combineValue(originalValue, newValue) {
```
iamigo: One of the big reasons to pull a function like this out to a separate module is that you get the opportunity to unit test the heck out of it separately from the context in which it will be called. Where are the tests?

api/v1/controllers/botData.js
tausifmuzaffar: Its called "botDataUtils" i think it feels okay, if i have to use the file again ill probably move it to utils more directly accessible

shriramshankar:
@tausifmuzaffar Sorry, I missed this request for review.

api/v1/controllers/botData.js
```diff
@@ -90,6 +92,37 @@ module.exports = {
...
+  upsertBotData(req, res, next) {
+    const queryBody = req.swagger.params.queryBody.value;
+    BotData.findOne({
+      where: {
+        name: queryBody.name,
+        botId: queryBody.botId,
+        roomId: queryBody.roomId,
+      },
+    })
+    .then((bd) => {
```
shriramshankar: Do you want to do this in a scope in the botData model?

api/v1/controllers/botData.js
```diff
@@ -13,12 +13,14 @@
 
 const helper = require('../helpers/nouns/botData');
 const Bot = require('../../../db').Bot;
+const BotData = require('../../../db').BotData;
```
shriramshankar: Can you use the BotData model that has already been required in `nouns/botData`?

api/v1/controllers/botData.js
tausifmuzaffar: I dont think it needs to? Do you think it would function better?

shriramshankar: It has got nothing to do with performance, just that converting it to a scope makes it more modular and that you are not writing any DB logic in the controller.

tausifmuzaffar: ive made some changes --- is this what you mean?

shriramshankar: This is fine too. I was thinking we could do something like https://github.com/salesforce/refocus/blob/master/db/model/sample.js#L222. Maybe for another day? I have approved this PR anyways.

PR #749, opened 2018-03-14
Sandboxing Javascript into iFrames

tausifmuzaffar:
Security wanted to make sure bots cant inject JS into other bots so we moved the bots from div containers to iFrame containers


view/rooms/app.js
```diff
@@ -35,9 +35,9 @@ let _io;
 let _user;
 let _roomName;
 let _isActive;
-let botInfo = {};
+const botInfo = {};
 const DEBUG_REALTIME = window.location.href.split(/[&\?]/)
-  .includes('debug=REALTIME');
+  .includes('log=REALTIME');
```
iamigo: I liked the consistency of knowing the same query parameter works in room and perspectives... why changing it?

tausifmuzaffar: In our bdk we implemented a  "Winston" like console logging 
https://git.soma.salesforce.com/Refocus-Bots/refocus-bdk/pull/24

There will be another PR that will introduce the same logging we use for the bots to refocus; this was a start to move them closer to each other.

tests/view/rooms/room/app.js
```diff
@@ -0,0 +1,80 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/view/rooms/room/app.js
+ */
+
```
shriramshankar: Turn on the strict mode here.

view/rooms/app.js
```diff
@@ -127,48 +139,178 @@ function createFooter(bot) {
...
+  iframe.style.width = '100%';
+  iframe.frameBorder = 0;
+
+  headerSection.appendChild(iframe);
+  headerSection.appendChild(footerSection);
+  botContainer.appendChild(headerSection);
+  botsContainer.appendChild(botContainer);
+
+  const parsedBot = parseBot(bot);
+  iframeBot(iframe, bot, parsedBot, user);
```
kriscfoster: @tausifmuzaffar just wondering where user is defined, should this be _user?

BrandonMurnane: +1

tausifmuzaffar: user is defined in the index.pug --- cant use _user because it sterilizes  the data, and the bot needs the unsterilize data

kriscfoster: @tausifmuzaffar makes sense now, thank you

iamigo: I have found it to be very useful to add a comment saying it's defined in the pug when you use a var defined in the pug, otherwise I find myself searching around for where it's declared, where value is assigned, etc.

PR #751, opened 2018-03-14
Add db/model/generator instance fn "updateForHeartbeat"

iamigo:
- Replaces array of subject absolutePaths with array of subject records.
- Replaces array of aspect names with array of aspect records.
- Tests

(Next PR will actually use this fn from Collectors API.)

### DIFF
--- a/db/model/generator.js
+++ b/db/model/generator.js
@@ -342,8 +342,12 @@ module.exports = function generator(seq, dataTypes) {
        */
       updateForHeartbeat() {
         const g = this.get();
-        const aspectPromises = g.aspects.map((a) =>
-          seq.models.Aspect.findOne({ where: { name: { $iLike: a } } }));
+        const aspectPromises = g.aspects.map((a) => seq.models.Aspect.findOne({
+          where: {
+            name: { $iLike: a },
+            isPublished: true,
+          },
+        }));
         let subjectPromises;
         if (g.subjects && g.subjects.length) {
           subjectPromises = g.subjects.map((absPath) =>
@@ -363,7 +367,8 @@ module.exports = function generator(seq, dataTypes) {
 
         return seq.Promise.all(aspectPromises)
         .then((records) => {
-          this.aspects = records.map((rec) => rec.get());
+          this.aspects = records.filter((found) => found)
+            .map((rec) => rec.get());
         })
         .then(() => seq.Promise.all(subjectPromises))
         .then((records) => {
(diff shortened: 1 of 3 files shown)


db/model/generator.js
```diff
@@ -332,6 +332,45 @@ module.exports = function generator(seq, dataTypes) {
...
+       * Replaces array of subject absolutePaths with array of subject records.
+       * Replaces array of aspect names with array of aspect records.
+       * TODO: set array of subjects when generator provides "subjectQuery"
+       *
+       * @returns {Generator} - the updated Generator
+       */
+      updateForHeartbeat() {
+        const g = this.get();
+        const aspectPromises = g.aspects.map((a) =>
+          seq.models.Aspect.findOne({ where: { name: { $iLike: a } } }));
```
shriramshankar: We have to find only published aspects , correct?

iamigo: Correct... and I'll make that change here... but we should never actually get into that state where the Generator references an unpublished aspect because we will be preventing unpublish/delete of aspects which are referenced by a SG.

iamigo:
@shriramshankar updated after comments

db/model/generator.js
```diff
@@ -332,6 +332,50 @@ module.exports = function generator(seq, dataTypes) {
...
+              where: {
+                absolutePath: {
+                  $iLike: absPath,
+                },
+                isPublished: true,
+              },
+            })
+          );
+        } else {
+          // TODO find subjects using subjectQuery
```
jgraff2: Separate story?

iamigo: It keeps growing in complexity so yeah I think I'm going to implement the "subjectQuery" stuff in a separate story. I will still have one more PR to actually use this new model fn from the api controller.

db/model/generator.js
```diff
@@ -332,6 +332,50 @@ module.exports = function generator(seq, dataTypes) {
...
+          subjectPromises = [];
+        }
+
+        return seq.Promise.all(aspectPromises)
+        .then((records) => {
+          this.aspects = records.filter((found) => found)
+            .map((rec) => rec.get());
+        })
+        .then(() => seq.Promise.all(subjectPromises))
+        .then((records) => {
```
jgraff2: I think it would be better to do these in parallel by wrapping the "all"s in a "join".

iamigo: Yes!

iamigo:
@jgraff2 updated after feedback

PR #754, opened 2018-03-16
fix heartbeat response

iamigo:
- update the generator db model to add a class fn called "findForHeartbeat" (which calls the recently-added "updateForHeartbeat")
- update the heartbeat response model definition in swagger: generatorsAdded/Updated have arrays of subject records and aspect records
- update the collector api to use the new "updateForHeartbeat" and "findForHeartbeat" functions when preparing the generatorsAdded/Updated to return

### DIFF
--- a/tests/db/model/generator/find.js
+++ b/tests/db/model/generator/find.js
@@ -141,10 +141,9 @@ describe('tests/db/model/generator/find.js >', () => {
     .then((res) => {
       expect(res).to.be.an('array').to.have.lengthOf(2);
       expect(res[0].aspects[0]).to.contain.property('name', 'Temperature');
-      console.log('findForHeartbeat', res);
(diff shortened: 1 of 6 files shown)


tests/db/model/generator/find.js
```diff
@@ -110,27 +110,40 @@ describe('tests/db/model/generator/find.js >', () => {
...
+      const sub = g.subjects;
       expect(sub).to.be.an('array');
       expect(sub[0]).to.contain.property('absolutePath', 'foo.bar');
       expect(sub[1]).to.contain.property('absolutePath', 'foo.baz');
       done();
     })
     .catch(done);
   });
+
+  it('findForHeartbeat', (done) => {
```
shriramshankar: Do we need a test on the api layer to make sure the heartbeat response returns a subject/aspect object?

iamigo: Added.

PR #757, opened 2018-03-19
Don't cause error when sample has no aspect.

jgraff2:


cache/models/samples.js
```diff
@@ -137,8 +137,11 @@ function cleanAddAspectToSample(sampleObj, aspectObj) {
...
   const aspect = sampleStore.arrayObjsStringsToJson(aspectObj,
     constants.fieldsToStringify.aspect);
-  modelUtils.cleanQueryBodyObj(aspect, embeddedAspectFields);
-  sampleStore.convertAspectStrings(aspect);
+  if (aspect) {
+    modelUtils.cleanQueryBodyObj(aspect, embeddedAspectFields);
+    sampleStore.convertAspectStrings(aspect);
+  }
+
   sampleRes.aspect = aspect;
```
shriramshankar: Can we attach the aspect only when it is not null to match what we have right now in prod?

jgraff2: What do you mean? That's what this is doing.
Do you mean it should be "if (aspect != null)" instead?

shriramshankar: I meant adding L145 inside the if block above.

shriramshankar: @jgraff2 I am already working on this file, I can make this change later. I am going to merge this PR so that I can do a deploy.

jgraff2: Oh ok, I didn't read that closely enough. Makes sense, I will change it.

jgraff2: oh. ok.

PR #759, opened 2018-03-20
fullName field in User table

kriscfoster:
- Added the fullName field to the user table
- Added fullName field to the "createdBy" association of the other db models


db/model/user.js
```diff
@@ -52,6 +52,10 @@ module.exports = function user(seq, dataTypes) {
       type: dataTypes.BOOLEAN,
       defaultValue: false,
     },
+    fullName: {
+      type: dataTypes.STRING,
```
iamigo: let's use one of the db constants for string length -- a name should not be unlimited length!

migrations/20180315131351-add-column-user-fullName.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up(qi, Sequelize) {
+    return qi.addColumn('Users', 'fullName', {
+      type: Sequelize.STRING,
+      defaultValue: null,
```
iamigo: specify string length, see comment above

kriscfoster:
@iamigo This is something that I don't know 100% yet but when we get to the stage of doing it we may reach out for guidance

PR #763, opened 2018-03-23
added generatorTemplate to swagger docs in put patch and post

harshkothari410:


harshkothari410:
Do not review it I found out one more thing is missing. I need to ask about that @iamigo

tests/api/v1/generators/post.js
```diff
@@ -89,6 +89,21 @@ describe('tests/api/v1/generators/post.js >', () => {
     });
   });
 
+  it('post without Generator Teamplate error', (done) => {
```
iamigo: typo in "Template"

harshkothari410:
Yeah locally this test is not failing. Let me check.

api/v1/swagger.yaml
```diff
@@ -3614,9 +3635,20 @@ paths:
...
+              generatorTemplate:
+                type: object
+                description: Details of Generator Template.
+                properties:
+                  name:
+                    type: string
+                    description: The Generator Template's name.
+                  version:
+                    type: string
+                    description: The Generator Template's version.
```
iamigo: description should be:

`A Generator Template version number (MAJOR.MINOR.PATCH) or a valid x-range, tilde range or caret range. Any of X, x, or * may be used to "stand in" for one of the numeric major, minor or patch values. A tilde range allows patch-level changes if a minor version is specified on the comparator, or allows minor-level changes if not. A caret range allows changes that do not modify the left-most non-zero major, minor or patch number. See https://github.com/npm/node-semver for more details on semver matching.`

api/v1/swagger.yaml
```diff
@@ -2040,7 +2040,6 @@ paths:
           in: formData
           description: The unique readable name for the bot
           type: string
-          required: true
```
iamigo: careful! you don't want to change the bot stuff in this pr!

harshkothari410: Oops I haven't change anything. Maybe because I pulled and pushed. This is strange.

harshkothari410:
@iamigo fixed that.

harshkothari410:
But we are not validating version at swagger level. So writing tests about version validation should not be part of this PR, and not even part of generators. It should be part of generatorTemplate. I can create another PR for these tests as part of GenetatorTemplate. Whats you thought? @iamigo

iamigo:
ok we're doing that validation in the db layer

harshkothari410:
Cool. So any other changes ?

iamigo:
not from me... one more reviewer needed

PR #764, opened 2018-03-23
New Bot field: version 

kamilsmuga:
Added a new field - version - for Bot model. This will allow better tracking of Bot releases as well as implementing a solid caching strategy


iamigo:
please don't add reviewers until tests are passing :(

kamilsmuga:
@iamigo tests fixed - they caught a lot of cases. Nice!

api/v1/swagger.yaml
```diff
@@ -2058,6 +2058,11 @@ paths:
           in: formData
           type: file
           format: binary
+        -
+          name: version
```
iamigo: required field?

migrations/20180323141450-add-column-bot-version.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  up(qi, Sequelize) {
```
iamigo: do you need to backfill existing rows in the migration too?

PR #765, opened 2018-03-23
Bots are in columns and can be moved around

kriscfoster:
- Bots are in 3 columns
- Bots can be moved between columns if in 3 column mode
- Fixed too much whitespace issue
- Test for writing content into iframe

screen shot 2018-03-23 at 16 20 53


kriscfoster:
@iamigo do you suggest we use something like Selenium to test this functionally?

iamigo:
We have to do **something** -- selenium is certainly a popular option. Just to be clear, I'm not talking about testing specific bots--that's outside the scope of this repo--but rather all the stuff that's provided as part of the "platform", esp. the interactive stuff.

kriscfoster:
Yeah that makes sense. I'll create a story to spike on this sometime in the near future.

PR #766, opened 2018-03-23
fullName in swagger

kriscfoster:
- Updated swagger to contain user fullName

### DIFF
--- a/api/v1/swagger.yaml
+++ b/api/v1/swagger.yaml
@@ -9151,12 +9151,14 @@ paths:
               - email
               - id
               - name
+              - fullName
               - createdBy
               - createdAt
               - updatedAt
               - -email
               - -id
               - -name
+              - -fullName
               - -createdBy
               - -createdAt
               - -updatedAt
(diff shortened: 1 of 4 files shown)


api/v1/swagger.yaml
```diff
@@ -9222,6 +9224,11 @@ paths:
                 maxLength: 256
                 description: >
                   First and last name.
+              fullName:
+                type: string
+                maxLength: 60
+                description: >
+                  The user's full name.
```
shriramshankar: Should we make the maxLength of the fullName >= the maxLength of the username on L9222 ? Same comment for the other swagger changes.

kriscfoster: From my understanding I made the maxLength 60 because this is the max length of the fullName string in the model. Do you think this isn't long enough?

shriramshankar: 60 is good too, just that when I looked at this PR  it kind of felt odd to see a full name field having a length that is less than the username field.  If there are other important things for you to do, we can talk about this later.  I have no problems with approving this PR. Let me know.

kriscfoster:
@shriramshankar I added a migration to change the string length of fullName to 256 and changed this in swagger too.. Is there a reason why the length of name is 254 here (https://github.com/salesforce/refocus/blob/master/db/model/user.js#L44) but the maxLength of name is 254 in swagger? I'm just curious

shriramshankar:
@kriscfoster that is must be a  mistake, both should match.

PR #770, opened 2018-03-25
find subjects - cache the responses using url as key

iamigo:
return from cache if available
(uses the standard functionality already provided by our doFind implementation)
allow admin to override the general CACHE_EXPIRY_IN_SECS with SUBJECT_CACHE_EXPIRY_IN_SECS

also: get rid of "getSubjectFromCache" while we're here... let GET /v1/subjects/[QUERY_PARAMS] and GET /v1/subjects/[KEY] use the db


config/toggles.js
```diff
@@ -111,10 +111,6 @@ const shortTermToggles = {
   logInvalidHmsetValues: environmentVariableTrue(pe,
     'LOG_INVALID_HMSET_VALUES'),
 
-  // Enable GET from cache for /v1/subjects, /v1/subjects/{key}
-  getSubjectFromCache: environmentVariableTrue(pe,
-    'GET_SUBJECT_FROM_CACHE'),
-
```
shriramshankar: This means will not be using the GET_SUBJECT_FROM_CACHE flag and we will no longer get /v1/subjects and v1/subject/{key} from sample store correct?

shriramshankar: Just read your comment :)

tests/api/v1/subjects/getWithLimit.js
```diff
@@ -41,5 +41,7 @@ describe('tests/api/v1/subjects/getWithLimit.js >', () => {
   after(u.forceDelete);
   after(tu.forceDeleteUser);
 
-  getWithLimit(path);
+  // Skip tests which modify the deafult limit since we're caching the results.
```
jgraff2: typo - default

api/v1/controllers/subjects.js
```diff
@@ -198,26 +199,20 @@ module.exports = {
...
+   * cached response if env var SUBJECT_CACHE_EXPIRY_IN_SECS > 0 and cached
+   * response is available.
    *
    * @param {IncomingMessage} req - The request object
    * @param {ServerResponse} res - The response object
    * @param {Function} next - The next middleware function in the stack
    */
   findSubjects(req, res, next) {
     validateTags(null, req.swagger.params);
-    if (featureToggles.isFeatureEnabled('getSubjectFromCache')) {
```
jgraff2: If we're removing the existing subject caching, I think there's also code in cache/models that can be removed.

tests/api/v1/subjects/getWithLimit.js
```diff
@@ -41,5 +41,7 @@ describe('tests/api/v1/subjects/getWithLimit.js >', () => {
...
+  const skipTestsWhichModifyDefaultLimit = true;
+  getWithLimit(path, false, skipTestsWhichModifyDefaultLimit);
```
jgraff2: I think it would be better to run the tests with SUBJECT_CACHE_EXPIRY_IN_SECS=0. There's no reason to run them with it set because most of the tests set up the data in the db directly.

And we also need specific tests for the find caching where it is enabled.

iamigo: Thx for all the thoughtful comments on this PR, @jgraff2 -- I'm going to take another stab at this with a broader brush.

api/v1/controllers/subjects.js
iamigo: yes, will remove that fn in next pr

PR #771, opened 2018-03-25
persp async request only isPublished resources (subj, asp, lens)

iamigo:


jgraff2:
Is it ok to release this without a flag?

iamigo:
@jgraff2 I feel like this change is really low risk so I didn't think releasing behind a flag would add value. What do you think?

jgraff2:
I agree, just wanted to double-check. Approving.

PR #773, opened 2018-03-26
Don't allow unpublish/rename/delete for aspects referenced by Sample Ge…

jgraff2:
…nerators

This is just for aspects now - I removed the changes for subjects since we triaged the story last week to only use the subjectQuery field.


db/model/aspect.js
```diff
@@ -341,6 +353,28 @@ module.exports = function aspect(seq, dataTypes) {
             resolve(found.length === 1);
           }));
       }, // isWritableBy
+
+      checkGeneratorReferences(action) {
+        const aspectName = this.previous('name');
+        const where = { where: { aspects: { $contains: [aspectName] } } };
```
iamigo: "contains" is case sensitive so we may have to use a more expensive query to check for reference :(

db/model/aspect.js
```diff
@@ -192,19 +193,30 @@ module.exports = function aspect(seq, dataTypes) {
...
        * @param {Aspect} inst - The instance being destroyed
        * @returns {Promise}
        */
       beforeDestroy(inst /* , opts */) {
-        return new seq.Promise((resolve, reject) =>
-          common.setIsDeleted(seq.Promise, inst)
-          .then(() => resolve(inst))
-          .catch((err) => reject(err))
-        );
+        return inst.checkGeneratorReferences('delete')
```
shriramshankar: Ideally, something like this should be a db association between the model and the generator model like the ones between L141 and L154.  Considering that the use case is slightly different does it make sense to create an association on the db level?

iamigo: I don't see any reason to NOT make it an association... I can't remember why we didn't make it one in the first place.

jgraff2:
I figured out how to allow case-insensitivity without setting up an association. I'm enforcing that the aspect names on the generator are stored as lower case in the db, so the aspect hooks can find referenced generators using the lowercase aspect name. Since the Generator aspects array isn't guaranteed to match the case of the actual aspects anyway, we don't lose anything by doing this.

jgraff2:
I figured out how to allow case-insensitivity without setting up an association. I'm enforcing that the aspect names on the generator are stored as lower case in the db, so the aspect hooks can find referenced generators using the lowercase aspect name. Since the Generator aspects array isn't guaranteed to match the case of the actual aspects anyway, we don't lose anything by doing this.

jgraff2:
A related question: Since we made the decision to only send aspect names and not the whole aspect object, and since the aspect names on the generator can be case insensitive, the aspect names the transform gets won’t necessarily match the case of the actual aspects.

So we either need to: 
1. specify that the aspect names passed to the transform are case-insensitive, and require the transform author to convert them if necessary 
2. get the aspect from the db anyway, but send only the name (more reason for association?)

iamigo:
aspect names in general *are* case-insensitive and the primary purpose of them in the transform is to generate the sample name, which is also case-insensitive, so... i think you're right that it's worth noting the docs and in the function comments in the example

db/model/aspect.js
```diff
@@ -341,6 +353,30 @@ module.exports = function aspect(seq, dataTypes) {
             resolve(found.length === 1);
           }));
       }, // isWritableBy
+
+      checkGeneratorReferences(action) {
+        const aspectName = this.previous('name');
+
+        // the aspect names are stored in lowercase on the generator
+        const where = { aspects: { $contains: [aspectName.toLowerCase()] } };
+        return seq.models.Generator.findAll({ where })
```
shriramshankar: Was the lower casing done just to use `contains` instead of `like`?

jgraff2: This queries on the Generator's "aspects" field, which is an array, so like doesn't work. The lower casing was done because it was the only way I found to make the query case-insensitive without setting up an association or iterating over all generators.

shriramshankar: Makes sense.

PR #778, opened 2018-03-29
update socket.io to current version

iamigo:


shriramshankar:
I think we should do a release just with all these version changes to see if it has any effect on the overall performance.

iamigo:
agreed

PR #779, opened 2018-03-29
Client side caching for Bots

kamilsmuga:
Route: /v1/bots/[key]

Client requests
- cache-control: max-age = 1 year
- cache-control: private - don't want CDNs to cache these as custom Javascript might leak stuff

Server response
- contains Etag
- if Etag is unchanged -> returns 304
- if Etag changed -> fetches a new version from DB and returns 200 (hopefully!!!)

More on HTTP caching: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching


view/utils.js
```diff
@@ -99,10 +99,10 @@ function removeSpinner(spinnerID) {
  * @param {String} url The url to get from
  * @returns {Promise} For use in chaining.
  */
-function getPromiseWithUrl(url) {
+function getPromiseWithUrl(url, headers) {
```
iamigo: Can you please add a function-level description for this function and note that when you do not pass in any headers, caching is off, but that you can send in your own url-specific headers if you want to override that and turn caching on

kamilsmuga:
@iamigo jsdoc improved as of e4cca3f

PR #781, opened 2018-03-29
added a toggle to get aspects and subjects from db instead of sample store

shriramshankar:
This is a minor change to get the subject and aspect of the sample from Postgres instead of the `sampleStore`. All of it is just undoing a tiny portion of the changes that was done to remove the `sampleStore` flag. This could possibly speed up other operations that access the `sampleStore`.


shriramshankar:
`tests/realtime/realtimeUtils.js` is getting bigger. I created a story to separate out the rooms part from this file.

PR #783, opened 2018-03-30
set up collector status endpoint

jgraff2:


api/v1/controllers/collectors.js
```diff
@@ -32,6 +32,7 @@ const GlobalConfig = require('../helpers/nouns/globalconfig').model;
 const config = require('../../../config');
 const GeneratorTemplate = require('../../../db/index').GeneratorTemplate;
 const Generator = require('../../../db/index').Generator;
+const Collector = require('../../../db/index').Collector;
```
shriramshankar: Can you use the collector model required in the noun folder? Looks like it has already been required in L14.

jgraff2: Actually, looks like I don't need it at all. Will remove.

PR #784, opened 2018-03-30
Use npm ci for Travis builds.

jgraff2:
This PR sets up our Travis builds to use the new npm ci command. It relies on a checked-in package-lock.json that specifies the exact packages to install.
From now on anytime a package is added or updated we will also need to run "npm install" locally and check in the updated package-lock.json.

The “npm install” step was previously taking 140s.
Now it takes 80s. 
We can potentially get it down to 30s:

npm ci was added in version 5.7. Travis uses an older version by default. So in order to use npm ci we need to install the latest npm in before_install. This adds 25s to the total time. If Travis updates the default version in the future or if there's a way to cache it we could save another 25s.

We could also save 25s by removing the post-install script. Currently it is:
"postinstall": "NODE_ENV=production gulp browserifyViews && gulp movecss && node scripts/moveSocket.js",
There's probably a better way to do this that doesn't take 25s on every build. Thoughts?

Timing Details:
- npm install: 140s
- npm ci (no cache): 65s + 25s = 90s
- npm ci (with cache): 55s + 25s = 80s
- npm ci (with cache, no post-install): 30s + 25s = 55s
- npm ci (with cache, no post-install, no npm install): 30s


.travis.yml
```diff
@@ -44,3 +44,13 @@ script:
   - npm run jscs-enforce
   - npm run checkdb
   - npm run test-travis
+
+before_install:
+ - npm install -g npm@^5.7.0
```
iamigo: line 49 should have two spaces before the dash
line 55 should have two spaces before the word "directories"
line 56 should have four spaces before the dash

iamigo: did you generate the package lock file *after* pulling all the recent module updates?
did you run `npm update` *before* generating the package lock file?

shriramshankar:
I thought the package-lock.json file was checked in by mistake. Do we have to do npm ci  considering this
> From now on anytime a package is added or updated we will also need to run "npm install" locally and check in the updated package-lock.json.

PR #786, opened 2018-04-03
Delay making api calls in perspective

shriramshankar:


shriramshankar:
@iamigo Changed the model to load the filter section only after the data becomes available.

meynet:
Can we make the button say "saving..." and gray out after being clicked? I think with that change I will be ok with this update. However, there may be some issues with how we are sending request since it is stalling which means that more than 6 tcp requests are active https://developers.google.com/web/tools/chrome-devtools/network-performance/reference#timing

Is there maybe another request that isn't finished?

PR #789, opened 2018-04-04
SSO to populate User fullName

kriscfoster:


view/loadView.js
```diff
@@ -108,6 +110,12 @@ function samlAuthentication(userProfile, done) {
       });
     }
 
+    if (!user.fullName) {
+      user.updateAttributes({
```
iamigo: why non-standard "updateAttributes"? shouldn't it just be "update"?

view/loadView.js
```diff
@@ -108,6 +110,12 @@ function samlAuthentication(userProfile, done) {
       });
     }
 
+    if (!user.fullName) {
+      user.update({
```
iamigo: The "done" on line 120 is going to be called *before* the user.update call executes unless you use promise semantics i.e. update(...).then(...)

kriscfoster:
@iamigo thanks, I think this is better now

kriscfoster:
@iamigo thanks, I think this is better now

view/loadView.js
```diff
@@ -110,14 +110,16 @@ function samlAuthentication(userProfile, done) {
...
+    if (user.fullName) {
+      // profile already attached - default scope applied on find
+      return done(null, user);
     }
 
-    // profile already attached - default scope applied on find
-    return done(null, user);
+    user.update({
+      fullName: userFullName,
+    }).then(() => {
```
iamigo: should be `.then(() => done(null, user));`

view/loadView.js
```diff
@@ -108,8 +110,14 @@ function samlAuthentication(userProfile, done) {
...
-    // profile already attached - default scope applied on find
-    return done(null, user);
+    if (user.fullName) {
+      // profile already attached - default scope applied on find
+      return done(null, user);
+    }
+
+    user.update({ // user.fullName doesn't exist, update user
+      fullName: userFullName,
+    }).then(() => done(null, user));
```
iamigo: Does the "user" which is passed on in the `done(null, user)` actually include the new fullName attribute? Test?

kriscfoster: @iamigo This test was here but I updated some wording to make this clearer https://github.com/salesforce/refocus/pull/789/files#diff-3036175005cd1bfda68f7b60355102bdR129 .. The test uses the user that was created in the beforeHook (without a fullName), and then checks that the fullName is there after

PR #790, opened 2018-04-06
Bot Object -> Version field -> PUT request bug

kamilsmuga:
Issue: PUT requests aren't updating `version` field for Bots
This PR fixes the bug

### DIFF
--- a/tests/api/v1/bots/put.js
+++ b/tests/api/v1/bots/put.js
@@ -68,6 +68,23 @@ describe('tests/api/v1/bots/put.js >', () => {
     });
   });
 
+  it('Pass, put bot version & url', (done) => {
+    api.put(`${path}/${testBot.id}`)
+    .set('Authorization', token)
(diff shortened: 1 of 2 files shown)


tests/api/v1/bots/put.js
```diff
@@ -68,6 +68,23 @@ describe('tests/api/v1/bots/put.js >', () => {
     });
   });
 
+  it('Pass, put bot version & url', (done) => {
```
tausifmuzaffar: Does it fail if the version or url is invalid?

kamilsmuga: There isn't any version validation (we're expecting a string, and suggesting to follow semantic versioning 2.0.0).
For the url - it will fail if a malformed url is passed.

iamigo: take a look at /db/constants.versionRegex - Refocus uses that validation for the sample generator template version field

PR #792, opened 2018-04-11
Use seq raw queries

shriramshankar:


shriramshankar:
screen shot 2018-04-11 at 6 26 37 pm
Check out the memory usage before and after the latest deploy in sandbox.
Before deploy metrics : bulk upserts with 1000 samples every 1 second. 
After release metrics: bulk upserts with 1000 samples every 200 ms

realtime/utils.js
```diff
@@ -389,20 +393,16 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
...
-      },
-    };
-    const aspOpts = {
-      where: {
-        name: { $iLike: aspName },
-      },
-    };
     const getAspectPromise = sample.aspect ? Promise.resolve(sample.aspect) :
-      aspectModel.findOne(aspOpts);
+    sequelize.query('SELECT * FROM "Aspects" WHERE id = :aspectId and ' +
```
iamigo: Please add a comment here describing why we're using raw query.

realtime/utils.js
```diff
@@ -376,7 +378,9 @@ function isIpWhitelisted(addr, whitelist) {
  * @returns {Promise} - which resolves to a complete sample with its subject and
  *   aspect.
  */
-function attachAspectSubject(sample, subjectModel, aspectModel) {
+function attachAspectSubject(sample) {
+  const sequelize = require('../db').sequelize;
```
iamigo: If this is necessary to require *here* inside the fn (as opposed to doing the require at the module level), please add a comment explaining why.

realtime/utils.js
iamigo: `SELECT *` -- is the real-time sample event going to contain the same list of aspect attributes as before?

shriramshankar: I did some spot checking in sandbox and prod. The aspect attributes are the same in both sandbox and prod.

PR #793, opened 2018-04-12
Log an error when a sample does not have aspectId and subjectId field 

shriramshankar:


realtime/utils.js
```diff
@@ -401,14 +401,15 @@ function attachAspectSubject(sample) {
...
+      '"isDeleted" = 0 LIMIT 1', { replacements: { name: aspName },
       type: sequelize.QueryTypes.SELECT, });
 
     const getSubjectPromise = sample.subject ? Promise.resolve(sample.subject) :
     sequelize.query('SELECT id, "absolutePath", name, tags FROM "Subjects" ' +
-      'WHERE id = :subjectId and "isDeleted" = 0', { replacements:
-        { subjectId: sample.subjectId }, type: sequelize.QueryTypes.SELECT, });
+      'WHERE "absolutePath" iLike :subName and "isDeleted" = 0 LIMIT 1',
+      { replacements: { subName: subName },
+      type: sequelize.QueryTypes.SELECT, });
```
shriramshankar: Or, should we throw an error if subjectId and aspectId is not there, like L387?

shriramshankar: Log the error and move on, like we have for samples without name.

iamigo: Yeah I think we should log them - that way we at least have some way of knowing that there's a data problem

PR #794, opened 2018-04-16
removed webpack

shriramshankar:


shriramshankar:
@iamigo  Do we want to support "react hot module reload" ? Not sure if we are still using it.

annyhe:
Hello @shriramshankar 
webpack makes the front-end development easier with hot reload. 
To get hot reload working, you'll need to
1. `export NODE_ENV=development` in both the terminal in which the server runs, and the terminal in which webpack runs
2. add `--watch` to the build command in package.json
3. start the webpack with build by running `npm run build` while the server is on.

You are free to take it out if you don't want hot reload.

PR #795, opened 2018-04-17
Make the Event Log longer

tausifmuzaffar:


tausifmuzaffar:
@iamigo @shriramshankar @kamilsmuga I deployed on devbox looks to work

PR #797, opened 2018-04-18
use seq object from model

shriramshankar:


shriramshankar:
@iamigo This does not work. Consumes 2x memory to the one we have in master.

realtime/utils.js
```diff
@@ -389,20 +389,15 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
...
-      aspectModel.findOne(aspOpts);
+      aspectModel.sequelize.query('SELECT * FROM "Aspects" WHERE ' +
+      'name iLIKE :aspName and "isDeleted" = 0 Limit 1', { replacements:
+        { aspName: aspName }, type: aspectModel.sequelize.QueryTypes.SELECT, });
     const getSubjectPromise = sample.subject ? Promise.resolve(sample.subject) :
-      subjectModel.findOne(subOpts);
+      subjectModel.sequelize.query('SELECT id, "absolutePath", name, tags ' +
+      'FROM "Subjects" WHERE "absolutePath" iLIKE :absPath and "isDeleted" = 0 Limit 1',
+      { replacements: { absPath: subName }, type:
+      subjectModel.sequelize.QueryTypes.SELECT, });
```
shriramshankar: @iamigo This does not work. Consumes 2x memory compared to the one in master.

shriramshankar:
This consumes less memory than the one running in prod (the rate of increase of memory is much slower)

PR #805, opened 2018-04-19
change heap size

harshkothari410:


harshkothari410:
Yes true @iamigo . we need to find way for this.

PR #806, opened 2018-04-23
Subject delete realtime

shriramshankar:
This is a change to send a subject delete event to the perspective when subjects abspath is changed.

### DIFF
--- a/db/model/subject.js
+++ b/db/model/subject.js
@@ -19,11 +19,9 @@ const constants = require('../constants');
 const dbErrors = require('../dbErrors');
 const redisOps = require('../../cache/redisOps');
 const subjectType = redisOps.subjectType;
-const eventName = {
-  add: 'refocus.internal.realtime.subject.add',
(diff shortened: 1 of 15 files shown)


db/model/subject.js
```diff
@@ -326,10 +313,20 @@ module.exports = function subject(seq, dataTypes) {
...
+        } else if (inst.isPublished && inst.changed('absolutePath')) {
+          /*
+           * When an absolutePath is changed, send a subject delete event with
+           * the old subject instance, followed by a subject add event with the
+           * new subject instance
+           */
+          publishObject(inst._previousDataValues, eventName.del,
+            changedKeys, ignoreAttributes);
+          publishObject(inst, eventName.add, changedKeys,
+              ignoreAttributes);
```
shriramshankar: L320 to L329 were the only changes required. The rest is just clean. (I went down a wrong path before realizing this)

PR #811, opened 2018-04-25
Pass URL Parameters on Redirect

tausifmuzaffar:


view/rooms/new/app.js
```diff
@@ -34,7 +34,7 @@ function getPathVariables(adr){
   }
 
   const paramType = qdata.roomType || '';
-  const paramActive = qdata.active === 'true';
+  const paramActive = qdata.active !== 'false';
```
iamigo: so the intended behavior here is for any value other than the string "false" to be treated as true?

tausifmuzaffar: yeah, default everything to active unless specifically false

kamilsmuga:
@tausifmuzaffar build failure

tausifmuzaffar:
@kamilsmuga @iamigo fixed build

view/rooms/new/app.js
iamigo: I wouldn't object to a comment on this line saying just that ;)

PR #819, opened 2018-04-27
force garbage collection in the bulkUpsert worker after publishing the samples

shriramshankar:


worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -17,6 +17,13 @@ const cacheSampleModel = require('../../cache/models/samples');
 const publisher = require('../../realtime/redisPublisher');
 
 module.exports = (job, done) => {
+  if (global.gc) {
+    global.gc();
+  } else {
+    logger.info('Garbage Collection unavailable. Pass --expose-gc when ' +
+      'starting the worker');
+  }
```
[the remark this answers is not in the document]
shriramshankar: This fixes the memory issue and this will not effect the response time too. So, this is one solution but may not be the most optimal. I want to merge #820, before deploying this in prod.

worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -60,6 +60,14 @@ module.exports = (job, done) => {
         }
       }
 
+      // force garbage collection after samples are published
+      if (global.gc) {
+        global.gc();
+      } else {
+        logger.info('Garbage Collection unavailable. Pass --expose-gc when ' +
```
iamigo: I don't think we need this logged on every job, esp. if this is just experimental. If we like the way this works, I'd feel better putting it behind a toggle, e.g. `if (featureToggles.isFeatureEnabled(...) && global.gc) { ... }`

shriramshankar: I was thinking we would run a clock job just for this if this works. Not sure what is the right way to do it though.

Procfile
```diff
@@ -1,4 +1,4 @@
 clock: IS_HEROKU=true npm run start-clock
 release: IS_HEROKU=true npm run on-release
 web: IS_HEROKU=true npm start
-worker: IS_HEROKU=true node --optimize_for_size --max_old_space_size=12000 worker/jobProcessor.js
+worker: IS_HEROKU=true node --optimize_for_size --expose-gc worker/jobProcessor.js
```
shriramshankar: @iamigo @harshkothari410 I made this change to use the default max_old_space instead of 12GB. We will fail fast if we know this does not work.

harshkothari410: True

PR #822, opened 2018-04-30
reject requests with multiple x-forwarded-for ip addresses

iamigo:
(behind feature toggle)


config/rejectMultipleXForwardedFor.js
```diff
@@ -0,0 +1,29 @@
...
+ * When the config toggle "rejectMultipleXForwardedFor" is true, reject with
+ * status 401 if there are multiple IP address values in the "X-Forwarded-For"
+ * header.
+ */
+'use strict'; // eslint-disable-line strict
+const FWD = 'x-forwarded-for';
+
+module.exports = (req, res, next) => {
+  if (req.headers && req.headers[FWD] && req.headers[FWD].indexOf(',') !== -1) {
+    const err = new Error('Access denied');
```
meynet: It seems more user friendly to say that it is related to IP address but is that bad security practice?

iamigo: I thought about that this morning and I was starting to craft a question to Joseph Trimble about it but as I was writing the question out it just became clear that we don't give attackers more details about why the attack failed.

iamigo:
Updated.

PR #824, opened 2018-05-01
Polyfill Resize Bug

tausifmuzaffar:
The resizing feature didnt work in Firefox

PR #825, opened 2018-05-01
Fix Firefox Polyfill

tausifmuzaffar:
Couldn't clear branch from 
https://github.com/salesforce/refocus/pull/824


view/rooms/app.js
```diff
@@ -406,24 +406,61 @@ function iframeBot(iframe, bot, parsedBot, currentUser) {
...
+        }
+      );
+    }        
+  </script>`;
 
   const iframeContent = iframeCss +
       `<script>var user = "${currentUser}"</script>
       ${contentSection}
-      <script>${botScript}</script>` +
+      <script>${botScript}</script>`+
```
shriramshankar: Looks like this change was not intentional ?

PR #826, opened 2018-05-02
Log out the number of events per key for real-time events published, …

iamigo:
…for perspectives and bots.

- (config/toggles.js) Define new toggle "enablePubStatsLogs" based on env var ENABLE_PUB_STATS_LOGS.
- (config.js) Define new config var pubStatsLogsIntervalMillis, default to 60000 (1 min.), override with env var PUB_STATS_LOGS_INTERVAL_MILLIS. When enablePubStatsLogs is enabled, we will write out "activity=pubStats" log lines every minute (unless admin overrides the default).
- (config/activityLog.js) Define new "activity=pubStats" log type
- (clock/scheduledJobs/pubStatsLogs.js) New scheduled job to generate the "activity=pubStats" log lines and reset the counters every <pubStatsLogsIntervalMillis>.
- (clock/index.js) When enablePubStatsLogs is enabled, add the new pubStatsLogs scheduled job to the list of jobs managed by the clock dyno.
- (realtime/constants.js) - Define a constant to use as the key to the redis cache hash object where we will store the published event counts.
- (realtime/redisSubscriber.js) - When enablePubStatsLogs is enabled, track the number of real-time events in redis cache.
- (tests/clock/pubStatsLogs.js) - Unit test the clock/pubStatsLogs.js functions

### DIFF
--- a/clock/scheduledJobs/pubStatsLogs.js
+++ b/clock/scheduledJobs/pubStatsLogs.js
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2018, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * clock/scheduledJobs/pubStatsLogs.js
+ *
+ * Activity logging to track the number of real-time events published per key
+ * since the last logging interval.
+ */
+const aType = 'pubStats';
+const prototype = require('../../config/activityLog').activityType[aType];
+const activityLog = require('../../utils/activityLog');
+const PUB_STATS_HASH = require('../../realtime/constants').pubStatsHash;
+const redis = require('../../cache/redisCache').client.cache;
+
+/**
+ * Generate a pubStats logging object.
+ *
+ * @param {String} key - The "publish" event key
+ * @param {Integer} count - The number of events published to that key since
+ *  the last logging interval
+ * @returns {Object} a new pubStats logging object
+ */
+function toLogObj(key, count) {
+  const obj = JSON.parse(JSON.stringify(prototype));
+  if (typeof key === 'string' && key.length > 0) obj.key = key;
+  if (typeof +count === 'number' && count > 0) obj.count = +count;
+  return obj;
+} // toLogObj
+
+/**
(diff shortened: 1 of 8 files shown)


config/toggles.js
```diff
@@ -60,6 +60,8 @@ const longTermToggles = {
   enableKueStatsActivityLogs:
     environmentVariableTrue(pe, 'ENABLE_KUESTATS_ACTIVITY_LOGS'),
 
+  enablePubStatsLogs: environmentVariableTrue(pe, 'ENABLE_PUB_STATS_LOGS'),
```
harshkothari410: Can you make one line comment over here about this toggle variable?

iamigo: done

clock/scheduledJobs/pubStatsLogs.js
```diff
@@ -0,0 +1,67 @@
...
+ * Generate the logging objects, one per publish "key". Look up the pubStats
+ * hash in redis and "reset" it by deleting it; generate and return an array of
+ * pubStats activity type logging objects.
+ *
+ * @returns {Array} array of pubStats activity type logging objects
+ */
+function generateLogObjects() {
+  return Promise.all([
+    redis.hgetallAsync(PUB_STATS_HASH),
+    redis.delAsync(PUB_STATS_HASH),
```
shriramshankar: Your expectation is `hgetall` will run before `del`, correct ? I thought the order of execution is not guaranteed.

realtime/redisSubscriber.js
```diff
@@ -11,54 +11,53 @@
...
-     * io emitter to send data to the browser clients.
-     */
-    emitter(io, key, parsedObj);
-  });
+  // Broadcast messages to Perspectives and Bots
+  [subBot, subPerspective].forEach((s) => {
+    s.on('message', (channel, mssgStr) => {
+      if (featureToggles.isFeatureEnabled('enableRealtimeActivityLogs')) {
+        logger.info('Size of message received by subscriber', mssgStr.length);
+      }
```
shriramshankar: We don't need L44. This was from the POC we did to reduce the sample size.

clock/scheduledJobs/pubStatsLogs.js
iamigo: The order of the results is the order of the array of promises, regardless of the order of execution/completion, but it is true that there is no guarantee that the first will start before the second, so I'll pull these apart and make them sequential

PR #828, opened 2018-05-03
Auto create Room

tausifmuzaffar:
When a room has all the settings needed to create a room it automatically creates it

### DIFF
--- a/view/rooms/new/app.js
+++ b/view/rooms/new/app.js
@@ -16,6 +16,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import FormController from './FormController';
+const request = require('superagent');
 const url = require('url');
 const address = window.location.href;
 const NAME_PATH = 3;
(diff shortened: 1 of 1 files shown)


view/rooms/new/app.js
```diff
@@ -54,6 +55,28 @@ function getPathVariables(adr){
...
+    active: paramaters.active,
+  };
+  req
+    .send(obj)
+    .end((error, res) => {
+      if (error) {
+        if (error.response.text.includes('SequelizeUniqueConstraintError')) {
+          window.location.href = `/rooms/${paramaters.name}`;
+        }
+        console.log('Error: ', error.response.text);
```
iamigo: use console.error?

view/rooms/new/app.js
```diff
@@ -54,6 +55,28 @@ function getPathVariables(adr){
   };
 }
 
+function createRoom(paramaters){
+  const req = request.post('/v1/rooms');
+  const obj = {
+    name: paramaters.name,
+    type: paramaters.type,
```
kamilsmuga: `roomType` instead of `type` is used previously in this file. Autoredirect doesn't work when specifying `type`

view/rooms/new/app.js
```diff
@@ -67,10 +90,17 @@ window.onload = () => {
     />,
     formContainer
   );
-  uPage.removeSpinner();
+  if (paramaters.name &&
+    paramaters.type &&
```
kamilsmuga: roomType instead of type

view/rooms/new/app.js
```diff
@@ -54,6 +55,28 @@ function getPathVariables(adr){
...
+  req
+    .send(obj)
+    .end((error, res) => {
+      if (error) {
+        if (error.response.text.includes('SequelizeUniqueConstraintError')) {
+          window.location.href = `/rooms/${paramaters.name}`;
+        }
+        console.error(error.response.text);
+      } else {
+        window.location.replace(`/rooms/${res.body.id}`);
```
kamilsmuga: can we keep URL params? maybe keeping params should be explicitily requested by adding `&keepParams=true`?

tausifmuzaffar:
@kamil what you think now

view/rooms/new/app.js
tausifmuzaffar: replaced type with roomType

kamilsmuga:
@tausifmuzaffar functionality wise- I like it. 

Build fails, though. 

```var pathName = url.parse(address, true).pathname.split('/');
                                                ^
TypeError: Cannot read property 'split' of null```

PR #829, opened 2018-05-03
allow client-side caching when perspective loads lens

iamigo:
by overriding the default headers which specify no cache


iamigo:
@kamilsmuga we would have to generate the ETag header when returning the lens, which we could do deterministically, i.e. from a hash of the lens, or even just a hash of the last updated date, but then it doesn't actually save use anything because we have to retrieve the lens and generate the hash for the browser to compare it to... so... never mind

PR #831, opened 2018-05-03
Bulk publish

shriramshankar:


shriramshankar:
@iamigo @harshkothari410 This is a another branch I want to deploy to prod and test. The worker memory does not seem to increase but there is a bit increase in the response times.

PR #832, opened 2018-05-03
Validate subject query

harshkothari410:


iamigo:
Looks ok so far but I'll take another look once tests are passing

db/model/generator.js
```diff
@@ -373,5 +378,48 @@ module.exports = function generator(seq, dataTypes) {
...
+
+    splitSubjectQuery.forEach((sq) => {
+      const splitSQ = sq.split('=');
+
+      if (splitSQ.length !== 2) {
+        throw new ValidationError('subjectQuery ValidationError',
+          'Format of subjectQuery must be "?{key}={value}"');
+      }
+
+      if (splitSQ[0] == 'tags' && splitSQ[1].indexOf('*') > -1) {
```
shriramshankar: Wondering if we also decided to prevent queries like absolutePath=* and name=* . I think we talked about it but I could not remember what we decided on.

iamigo: Since "absolutePath=*" (or "name=*") are the same as not including that clause at all, we should treat it the same as if it wasn't there at all, and send back validation error if there are no other query params

harshkothari410:
@iamigo @shriramshankar I updated code.Now added code into helper file instead of model file.

PR #835, opened 2018-05-04
Fixed apos issue

BrandonMurnane:


view/loadView.js
```diff
@@ -151,7 +151,7 @@ function loadView(app, passport) {
       (req, res) => {
         const trackObj = {
           trackingId: viewConfig.trackingId,
-          user: JSON.stringify(req.user),
+          user: JSON.stringify(req.user).replace(/'/g,"apos;"),
```
iamigo: shouldn't it be `"&apos;"`?

view/rooms/app.js
```diff
@@ -911,8 +911,8 @@ window.onload = () => {
 
   // Note: this is declared in index.pug:
   _io = io;
-  _user = JSON.parse(user.replace(/&quot;/g, '"'));
-  let room;
+ _user = JSON.parse(user.replace(/&quot;/g, '"')
+    .replace(/apos;/g, "'"));  let room;
```
iamigo: `&apos;` ?

view/loadView.js
iamigo: Can you add a comment about why you're not escaping it "normally" so people don't keep asking questions like this.

BrandonMurnane:
@tausifmuzaffar fixed

PR #836, opened 2018-05-04
allow admin to prioritize/deprioritize jobs by user name and/or token

iamigo:
in addition to just ip address

modified the calculateJobPriority function signature to get rid of args we aren't using, PLUS added args to pass in conf.prioritizeJobsFrom and conf.deprioritizeJobsFrom for better testability


pallavi2209:
@iamigo Build failing - I think some flapping test.

PR #823, opened 2018-04-30
currentCollector to Generator - db migration

pallavi2209:
db validation for collector and tests coming in next PRs

PR #837, opened 2018-05-04
currentCollector field api changes for generator endpoints

pallavi2209:


shriramshankar:
@iamigo, The "collector reassignment" spike that was done, assumes that a current collector has already been assigned by the user. If we want to make it read-only, we have also have to revert #823.

pallavi2209:
I am holding off the changes until we decide.

iamigo:
Refocus should set it to the first one in the list. I don’t think we want to let people do it--we should always be free to reassign as needed.

pallavi2209:
Closing this Pr and creating a new one with reverted and new changes.

PR #839, opened 2018-05-10
upgrade to node 8

shriramshankar:


shriramshankar:
@iamigo Upgrading to node 8 seems to help with the memory issue. Should we deploy this branch and see how this performs in prod?

iamigo:
Wow! Love it! Yeah, let's see how it looks in prod--but obviously prepare for roll back

PR #841, opened 2018-05-10
Add "intervalSecs" to Generator db+api

iamigo:


tests/api/v1/generators/post.js
```diff
@@ -58,6 +58,7 @@ describe('tests/api/v1/generators/post.js >', () => {
 
       expect(res.body.apiLinks).to.be.an('Array');
       expect(res.body.name).to.include(generator.name);
+      expect(res.body).to.have.property('intervalSecs', 60);
```
jgraff2: Should also have a test case where the value is actually set.

iamigo: yes, done

api/v1/swagger.yaml
```diff
@@ -11559,6 +11585,10 @@ definitions:
         readOnly: true
         description: >
           The collector id.
+      intervalSecs:
```
pallavi2209: Do we need to add this field to GeneratorFieldsParam, GeneratorResponse and HeartbeatGeneratorResponse ? I added currentCollector to those.

tests/db/model/generator/create.js
```diff
@@ -169,44 +170,6 @@ describe('tests/db/model/generator/create.js >', () => {
     });
   });
 
-  it('not ok, with both subjects and subjectQuery present in ' +
```
pallavi2209: Do we need to remove these 2 tests?

PR #848, opened 2018-05-20
Filter generatorsAdded for "currentCollector"

iamigo:
- "startCollector" should only return a generator in "generatorsAdded" if the collector is the "currentCollector" for that generator


harshkothari410:
Any test needs to be updated or added ? @iamigo otherwise looks good for me.

iamigo:
@harshkothari410 we'll be able to properly test this change via api once we close out two more stories we have about assigning "currentCollector" to generators

api/v1/controllers/collectors.js
```diff
@@ -362,75 +362,78 @@ function heartbeat(req, res, next) {
...
+  })
+  /* Update or create */
+  .then((coll) => coll ? coll.update(body) : helper.model.create(body))
+  .then((coll) => {
+    name = coll.name;
+    collToReturn = coll;
+    return coll.getCurrentGenerators();
+  })
+  /* Filter: only want generators marked as "current" for this collector. */
+  .then((gens) => gens.filter((g) => g.currentCollector = name))
```
jgraff2: Shouldn't coll.getCurrentGenerators already be returning the correct list of current generators?

iamigo: Yeah, that's in the "trackGenerators" story -- this filter shouldn't be necessary if that is working.

jgraff2: So this is just a stop-gap until that gets fixed?

iamigo: Er, I wasn't thinking of it that way but I think you're right.

jgraff2: Ok. Can you add a TODO to remove the filter once getCurrentGenerators is working?

jgraff2: @iamigo can you see my comment in this old pr?

There was a typo in this line that we didn't catch in the review: "=" instead of "==". So this wasn't actually filtering. If it had been correct, the trust collector would have been broken when this went out about a month ago.

This is why I was confused earlier. I was pretty sure I hadn't made any functional changes in the reassign generators PR. I had replaced this filter with a find by collector name, which is functionally the same. But the filter wasn't actually filtering, so my change made it work correctly, which broke the generator assignment since we hadn't fully thought through how to make the transition.

I updated the next story to change this back once all the assignment is in place.

iamigo: Good catch @jgraff2

PR #850, opened 2018-05-21
Job logging

iamigo:
- define new 'job' activity log type
- update each of the job types to emit 'job' log before calling done
- get rid of unnecessary nesting in config/activityLog.js
- modify feature toggles for all activity log types - use a single "ENABLE_ACTIVITY_LOGS" env var instead of a separate env var for each activity log type
- wrap "unauthorized" activity logging in feature toggle


tests/config/toggles.js
```diff
@@ -31,4 +31,17 @@ describe('tests/config/toggles.js >', () => {
...
+    const env = {
+      myList: ' Abc,def ,GHi,',
+    };
+    expect(toggles.envVarIncludes(env, 'myList', 'abc')).to.equal(true);
+    expect(toggles.envVarIncludes(env, 'myList', 'ABC')).to.equal(true);
+    expect(toggles.envVarIncludes(env, 'myList', 'def')).to.equal(true);
+    expect(toggles.envVarIncludes(env, 'myList', 'DEF')).to.equal(true);
+    expect(toggles.envVarIncludes(env, 'myList', 'ghi')).to.equal(true);
+    expect(toggles.envVarIncludes(env, 'myList', 'GHI')).to.equal(true);
+    done();
```
jgraff2: Should have at least one false test. 
Also null and wildcards?

config/toggles.js
```diff
@@ -49,28 +76,24 @@ function environmentVariableTrue(processEnv, environmentVariableName) {
  * things from getting out of hand and keeping tons of dead unused code around.
  */
 const longTermToggles = {
-  // Enable api activity logging
-  enableApiActivityLogs:
-    environmentVariableTrue(pe, 'ENABLE_API_ACTIVITY_LOGS'),
+  // Activity logging
+  enableApiActivityLogs: envVarIncludes(pe, 'ENABLE_ACTIVITY_LOGS', 'api'),
+  enableJobActivityLogs: envVarIncludes(pe, 'ENABLE_ACTIVITY_LOGS', 'job'),
+  enableKueStatsActivityLogs: envVarIncludes(pe, 'ENABLE_ACTIVITY_LOGS',
```
jgraff2: Good idea, this is much better!

tests/config/toggles.js
```diff
@@ -31,4 +31,22 @@ describe('tests/config/toggles.js >', () => {
     expect(toggles.environmentVariableTrue(env, 'hello')).to.equal(false);
     done();
   });
+
+  it('envVarIncludes', (done) => {
+    const env = {
+      myList: ' Abc,def ,GHi,',
+      withWildcard: '*',
+    };
+    expect
```
jgraff2: code style

iamigo: sorry about that :(

PR #855, opened 2018-05-21
fix aspect permission test flapper

iamigo:


tests/api/v1/generators/post.js
```diff
@@ -269,7 +269,7 @@ describe('tests/api/v1/generators/post.js >', () => {
       .catch((e) => done(e));
     });
 
-    it('error, all aspects do not have user permission', (done) => {
+    it.only('error, all aspects do not have user permission', (done) => {
```
jgraff2: left in by mistake?

iamigo: i hate when i do that :(

tests/api/v1/generators/post.js
```diff
@@ -295,9 +295,9 @@ describe('tests/api/v1/generators/post.js >', () => {
       })
       .then((asp2) => { // created aspect2
         aspects.push(asp2.name);
-        asp2.addWriter(user2); // assign user2 as writer to aspect2
-        return validateGeneratorAspectsPermissions(aspects, req);
+        return asp2.addWriter(user2); // assign user2 as writer to aspect2
       })
+      .then(() => validateGeneratorAspectsPermissions(aspects, req))
```
jgraff2: Good catch. Should it be the same on line 291?

iamigo: changed a few more, yeah

iamigo: tests passing

PR #862, opened 2018-05-24
Upgrade sequelize

jgraff2:


api/v1/constants.js
```diff
@@ -43,12 +44,12 @@ module.exports = {
   BASE_SCOPE: 'baseScope',
   SEQ_DEFAULT_SCOPE: 'defaultScope',
   SEQ_DESC: 'DESC',
-  SEQ_LIKE: '$iLike',
-  SEQ_CONTAINS: '$contains',
-  SEQ_OVERLAP: '$overlap',
-  SEQ_IN: '$in',
-  SEQ_OR: '$or',
-  SEQ_NOT: '$not',
+  SEQ_LIKE: Op.iLike,
```
iamigo: don't need these defined here in constants anymore?

jgraff2: There are some places in the code where these are used, and some places where the operators are used directly. I left it as-is and just changed the value of the constants variable here.

I think these were originally defined here because they were using string values, so yeah since we're using symbols now there's no reason to have separate constants values. I will remove these and use them directly.

venkateshperneedy:
I am looking for a salesforce administrator at Tustin, CA
Full-time Role 
if you know anyone reaches me: Venkatesh@itminds.net

PR #863, opened 2018-05-24
fix associations bug

jgraff2:
This is a fix for some of the failing associations tests. When I saw these were failing, I assumed it was the same cause as the issue that I had already investigated and determined was a sequelize bug, so I skipped them as well. It was actually an unrelated bug in my changes, which is fixed here.

The original issue is still occurring, it was not fixed by the sequelize upgrade.
I created W-5026622 to investigate further.

### DIFF
--- a/api/v1/helpers/verbs/utils.js
+++ b/api/v1/helpers/verbs/utils.js
@@ -641,16 +641,12 @@ function findByKey(props, params, extraAttributes) {
   opts.where = {};
   opts.where[props.nameFinder || 'name'] = keyClause;
 
-  const attrArr = [];
-  if (opts.attributes && Array.isArray(opts.attributes)) {
-    for (let i = 0; i < opts.attributes.length; i++) {
-      attrArr.push(opts.attributes[i]);
-    }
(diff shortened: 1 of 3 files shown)


api/v1/helpers/verbs/utils.js
```diff
@@ -641,16 +641,12 @@ function findByKey(props, params, extraAttributes) {
...
-
-  if (extraAttributes && Array.isArray(extraAttributes)) {
-    for (let i = 0; i < extraAttributes.length; i++) {
-      attrArr.push(extraAttributes[i]);
+  let attrArr = opts.attributes;
+  if (extraAttributes && Array.isArray(extraAttributes) && extraAttributes.length) {
+    if (attrArr) {
+      attrArr.push(...extraAttributes);
+    } else {
+      attrArr = extraAttributes;
```
jgraff2: My changes last time to getScopedModel relied on the opts.attributes array being passed in. That was the case when called from doFindAndCountAll, but I missed that it was passing in a cloned array here.

PR #866, opened 2018-05-28
Fix for wobbly iframe bug

kriscfoster:
- Solution for the iframe shaking/wobbling which was caused by in infinite loop of the height changing
- Generalised the outputSize function so it isn't browser specific

### DIFF
--- a/view/rooms/app.js
+++ b/view/rooms/app.js
@@ -410,6 +410,7 @@ function iframeBot(iframe, bot, parsedBot, currentUser) {
   <script>
     function outputSize() {
       const el = document.getElementById("${bot.name}");
+      // Sometimes clientHeight was > scrollHeight so we need to get the max
       const botHeight =  Math.max(el.clientHeight, el.scrollHeight);
    
       parent.postMessage(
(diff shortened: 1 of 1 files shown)


view/rooms/app.js
```diff
@@ -408,50 +408,37 @@ function iframeBot(iframe, bot, parsedBot, currentUser) {
...
+      // Sometimes clientHeight was > scrollHeight so we need to get the max
+      const botHeight =  Math.max(el.clientHeight, el.scrollHeight);
+   
+      parent.postMessage(
         {
           "name": "${bot.name}",
-          "height": document.getElementById("${bot.name}").clientHeight < 100 ?
-            600 :
-            document.getElementById("${bot.name}").clientHeight
+          "height": botHeight < 100 ? 100 : botHeight
```
iamigo: This can written more simply as `Math.max(botHeight, 100)` (and use a named constant for 100)

tausifmuzaffar:
@kriscfoster update branch

package.json
```diff
@@ -109,16 +109,16 @@
     "html-webpack-plugin": "^2.16.0",
     "if-env": "^1.0.4",
     "ip": "^1.1.5",
-    "joi": "^13.4.0",
+    "joi": "^13.3.0",
```
iamigo: are all these package version changes intentional?

kriscfoster: @iamigo I'm only seeing changes to view/rooms/app.js

PR #867, opened 2018-05-29
Add logging rate limit

sameesiddiqui:
* Added new activityLog object for limiter
* Create & print logObject if user hits rate limit


tests/limiter/limiter.js
```diff
@@ -182,14 +183,28 @@ describe('tests/limiter/limiter.js >', () => {
     });
 
     it('First user, 429', (done) => {
+      tu.toggleOverride('enableLimiterActivityLogs', true);
+      logger.on('logging', testLogMessage);
```
jgraff2: It would be better to add this in a new "describe" specifically for logging to keep it separate from the functionality tests.

tests/limiter/limiter.js
```diff
@@ -182,14 +183,28 @@ describe('tests/limiter/limiter.js >', () => {
...
       })
       .catch(done);
+
+      function testLogMessage (transport, level, msg, meta) {
+        const logObj = {};
+        logObj['activity'] = msg.split(' ')[0].split('=')[1] //gets activity param from log
+        try {
+          expect(logObj.activity).to.equal('limiter');
+          logger.removeListener('logging', testLogMessage);
+          tu.toggleOverride('enableLimiterActivityLogs', false);
```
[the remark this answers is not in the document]
sameesiddiqui: Don't think so. Tested some print statements and it doesn't reach them if expect throws an error.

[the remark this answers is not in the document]
iamigo: So everything *after* the expect should be in an "after"?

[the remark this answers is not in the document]
sameesiddiqui: Ah I get it now. 

I was referencing the way logging was done in the bulkUpsert tests (lines 159-163) and the toggleOverride to disable the activity logs was in the "try" block. Is there a reason why the "after" block doesn't disable the api and worker activity logs?

tests/limiter/limiter.js
sameesiddiqui: done

tests/limiter/limiter.js
```diff
@@ -302,6 +303,67 @@ describe('tests/limiter/limiter.js >', () => {
...
+      doneCopy = done
+      makeRequest('/v1/aspects', 'post', token1)
+      .then((res) => {
+        expect(res.status).to.equal(constants.httpStatus.TOO_MANY_REQUESTS);
+        expect(res.header['x-ratelimit-limit']).to.equal('3');
+        expect(res.header['x-ratelimit-remaining']).to.equal('0');
+      })
+      .catch(done);
+    });
+  })
```
iamigo: missing ;

tests/limiter/limiter.js
```diff
@@ -74,28 +75,28 @@ describe('tests/limiter/limiter.js >', () => {
...
+        } else {
+          resolve(res);
+        }
       });
-    }
+    });
+  }
+
+  describe('limit functionality >', () => {
+    afterEach((done) => setTimeout(done, 100));
```
iamigo: must be careful with timeouts -- don't want to set it higher than necessary, but it might work ok locally and fail in travis

tests/limiter/limiter.js
```diff
@@ -302,6 +303,67 @@ describe('tests/limiter/limiter.js >', () => {
...
+        expect(res.status).to.equal(constants.httpStatus.CREATED);
+        expect(res.header['x-ratelimit-limit']).to.equal('3');
+        expect(res.header['x-ratelimit-remaining']).to.equal('0');
+        done();
+      })
+      .catch(done);
+    });
+
+    it('limit over longer period', (done) => {
+      doneCopy = done
```
iamigo: missing ;

tests/limiter/limiter.js
sameesiddiqui: strange. ran it before and it didn't catch either of these. (ran it again and still doesn't complain)

iamigo: lol please update the list of directories to enforce on in the "jscs-enforce" script in package.json to include this directory

tests/limiter/limiter.js
sameesiddiqui: Does travis normally require higher timeouts than local testing? Is there some metric to use or is it experimental with how long to set them

iamigo: trial and error :(

package.json
```diff
@@ -151,7 +151,7 @@
     "redux-logger": "^2.0.2",
     "redux-thunk": "^0.1.0",
     "semver-regex": "^1.0.0",
-    "sequelize": "^4.37.0",
+    "sequelize": "^4.37.10",
```
iamigo: must update package-lock when changing this

jgraff2: You shouldn't need to change this. The "^" means npm will use the latest version of 4.x.x. You generally only need to update it when upgrading the major version or adding a new package.

And for future reference: when you do need to change it here, you would also need to run npm install and check in the updated package-lock.json. (The build is failing right now because package.json doesn't match package-lock.json.)

jgraff2: Was there a reason you changed this?

sameesiddiqui: must have updated automatically when I git pulled or ran npm install. I'll change it back

jgraff2: If you run "npm install sequelize" it will install the latest sequelize version and update package.json. Generally you should just run "npm install", and it will install the latest version for all packages without changing package.json.

And generally, package-lock.json will be updated any time you run "npm install", but you should only check it in when necessary (new package or major version update).

It doesn't really matter, functionally it's the same, but I think it's better to keep the PRs clean and only check in necessary changes.

I think you should revert the changes to package-lock.json as well.

@iamigo make sense?

PR #872, opened 2018-06-02
Track generator changes for current collector only

jgraff2:
Previously the changes were tracked based on Generator.collectors. This uses Generator.currentCollector instead.

Because currentCollector can't be updated directly through the API the calls to trackGeneratorChanges are moved to db hooks, and the tests are updated to not go through the api.

Actually setting the currentCollector field will be done in the other story I have this sprint.

### DIFF
--- a/api/v1/helpers/verbs/heartbeatUtils.js
+++ b/api/v1/helpers/verbs/heartbeatUtils.js
@@ -102,28 +102,27 @@ function getKey(collectorName, changeType) {
  * collector.
  *
  * @param {Object} generator - The Generator that has been changed
- * @param {Array} oldCollectors - The names of the collectors that were associated
- *  with this generator before the update.
- * @param {Array} newCollectors - The names of the collectors that are associated
- *  with this generator now, after the update has been applied.
+ * @param {Array} oldCollector - The name of the Collector this Generator was
+ *  assigned to before the update.
+ * @param {Array} newCollector - The name of the Collector this Generator is
+ *  assigned to after the update.
  */
-function trackGeneratorChanges(generator, oldCollectors, newCollectors) {
-  const genId = generator.id;
-  const allCollectors = Array.from(new Set([...oldCollectors, ...newCollectors]));
+function trackGeneratorChanges(generator, oldCollector, newCollector) {
+  if (oldCollector === newCollector) {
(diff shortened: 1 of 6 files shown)


db/model/generator.js
```diff
@@ -191,16 +192,36 @@ module.exports = function generator(seq, dataTypes) {
...
+              return inst.addWriter(inst.createdBy);
+            }
+          }),
+        ]);
 
         return inst;
       }, // afterCreate
+
+      afterUpdate(inst) {
+        let oldCollector = inst.previous('currentCollector');
```
iamigo: check `inst.changed()` for whether currentCollector or isActive were changed before doing anything here... may be able to short circuit and exit without doing anything

jgraff2: If currentCollector wasn't changed we still need to track the changes based on whether isActive changed. (line 216 & 218)
And if isActive wasn't changed we still need to check if it's false, in which case we don't track anything. (line 220)

api/v1/helpers/verbs/heartbeatUtils.js
```diff
@@ -102,28 +102,26 @@ function getKey(collectorName, changeType) {
...
+  }
 
-      // determine the change type
-      let change;
-      const inOld = oldCollectors.includes(collectorName);
-      const inNew = newCollectors.includes(collectorName);
-      if (inOld && !inNew) change = DELETED;
-      if (!inOld && inNew) change = ADDED;
-      if (inOld && inNew) change = UPDATED;
+  function trackChangesForCollector(collectorName, change, generator) {
```
pallavi2209: Defining trackChangesForCollector function inside trackGeneratorChanges function is resulting in multiple eslint errors on my sublime - I would like to know if there is an advantage in doing so.

jgraff2: Personally I much prefer using internal functions in a case like this where it isn't used anywhere else and makes no sense outside the context of the calling function. It keeps it self-contained and makes it easier to understand.

jgraff2: You already approved so I will go ahead and merge if there are no objections.

pallavi2209: Please do. For some reason, squash and merge is not enabled for me right now.

PR #874, opened 2018-06-04
flapper fix

jgraff2:


tests/db/model/aspect/find.js
```diff
@@ -19,20 +19,18 @@ const Op = require('sequelize').Op;
...
-        tags: ['jedi', 'boring'],
-      });
-      Aspect.create({
-        name: 'leia',
-        timeout: '1m',
-        isPublished: true,
-        tags: ['princess', 'jedi'],
-      });
-    })
+    .then(() => Aspect.create({
```
iamigo: `.then(() => Aspect.bulkCreate([{ ... }, { ... }]))` would also work here

PR #877, opened 2018-06-05
Seralize BotData instead of using JSON.stringify()

kriscfoster:
- Bots have been having some trouble with botData that contains special characters. This solves the problem and makes the shared context/sync more reliable


kriscfoster:
@iamigo @tausifmuzaffar @kamilsmuga Could we review this now but wait to merge as we need to do a scheduled release with some bots.

PR #880, opened 2018-06-06
update jwtUtils token verification for generators and collectors

iamigo:
**utils/jwtUtil.js**
- add new "IsGenerator" header
- minor formatting and var name cleanup
- do user check for user/generator/collector token types
- unnest the nested "try to verify me as a person first and if that fails then try to verify me as a collector and so on..." -- identify the type of user but fall back to trying to verify as a person if the type is not available (e.g. from an old token generated before we were adding that information to the token payload)
- when verifying a collector token, the token name must match a collector name and the user must exist
- when verifying a generator token, the token name must match a generator name and the user must exist
- simplify error message when token is invalid to just "Forbidden" -- we shouldn't be giving a potential attacker more details about why it was forbidden

**api/v1/controllers/collectors.js**
- update the patch algorithm when patching restricted fields to just use the "IsCollector" header which is set by the authentication process
- collector tokens will have collector name as the token name and user name as user name
- create collector tokens with { IsCollector: true }
- generator tokens will have generator name as the token name and user name as user name
- create generator tokens with { IsGenerator: true }

**realtime/setupSocketIO.js**
- minor cleanup and var name changes for clarification

**tests/api/v1/collectors/heartbeat.js**
- minor formatting and test name clarification only

**tests/api/v1/collectors/patch.js**
- create collector tokens with collector name, user name and { IsCollector: true }
- rewrite tests to expect "Forbidden" message
- minor formatting

**tests/api/v1/collectors/start.js**
- rewrite tests to expect "Forbidden" message

**tests/enforceToken/createToken.js**
- rewrite tests to expect "Forbidden" message

**tests/utils/jwtUtil.js**
- bots require version
- minor formatting, var name cleanup, test name cleanup
- create collector tokens with collector name, user name and { IsCollector: true }
- add tests for generator tokens
- update tests to check for IsGenerator header

### DIFF
--- a/utils/jwtUtil.js
+++ b/utils/jwtUtil.js
@@ -151,7 +151,9 @@ function verifyCollectorToken(payload, req, cb) {
   .then(() => Collector.findOne({ where: { name: payload.tokenname } }))
   .then((coll) => {
     if (!coll) {
-      throw new apiErrors.ForbiddenError({ explanation: 'Forbidden' });
+      throw new apiErrors.ForbiddenError({
+        explanation: 'Authentication Failed',
+      });
     }
 
     assignHeaderValues(req, payload);
@@ -159,7 +161,7 @@ function verifyCollectorToken(payload, req, cb) {
   })
   .catch((err) => {
     throw new apiErrors.ForbiddenError({
-      explanation: 'Invalid/No Token provided.',
+      explanation: 'Authentication Failed',
     });
   });
 } // verifyCollectorToken
@@ -181,7 +183,9 @@ function verifyGeneratorToken(payload, req, cb) {
   .then(() => Generator.findOne({ where: { name: payload.tokenname } }))
   .then((gen) => {
     if (!gen) {
-      throw new apiErrors.ForbiddenError({ explanation: 'Forbidden' });
+      throw new apiErrors.ForbiddenError({
+        explanation: 'Authentication Failed',
+      });
     }
 
     assignHeaderValues(req, payload);
@@ -189,7 +193,7 @@ function verifyGeneratorToken(payload, req, cb) {
   })
   .catch(() => {
     throw new apiErrors.ForbiddenError({
-      explanation: 'Invalid/No Token provided.',
+      explanation: 'Authentication Failed',
     });
   });
 } // verifyGeneratorToken
@@ -209,7 +213,7 @@ function setUser(payload, req) {
   .then((user) => {
     if (!user) {
       throw new apiErrors.ForbiddenError({
-        explanation: 'Invalid Token.',
+        explanation: 'Authentication Failed',
       });
     }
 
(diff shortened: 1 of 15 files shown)


tausifmuzaffar:
Looks good to me @BrandonMurnane can you confirm?

pallavi2209:
"simplify error message when token is invalid to just "Forbidden" -- we shouldn't be giving a potential attacker more details about why it was forbidden" - this makes sense, the only con in my opinion is when testing -
There can be forbidden error for various reasons. When testing for a specific use case, we check for forbidden error only - so how do we make sure that the error is because of the use case we are testing (and not something else) ?

iamigo:
You raise a good point, @pallavi2209--I don't have an answer for that. Do you have any suggestions?

pallavi2209:
@iamigo I have seen this happening before in our application and that is why I made sure that I add some specific text related to the error. We can probably use resourceType: 'Token' (e.g.) to narrow down the Forbidden Error, but I am not sure how can we narrow down to a specific error, except for using message or explanation attributes with the minimal text we are comfortable with.
Just some googling-
[https://stackoverflow.com/questions/942951/rest-api-error-return-good-practices](url)
".....the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity. If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead."

iamigo:
I am comfortable with distinguishing between a couple of broad categories but without providing details which could help an attacker. So maybe 403 with explanation "Authentication Failed" (e.g. login or token authentication failed) vs. 403 with explanation "Insufficient Permissions" (e.g. you are trying to edit/delete a resource you do not have permission to edit or delete). Is there maybe one or two other broad categories of 403 FORBIDDEN we could call out in the error explanation?

utils/jwtUtil.js
```diff
@@ -263,17 +274,15 @@ function verifyUserToken(req, cb) {
 function verifyToken(req, cb) {
   const token = req.session.token || req.headers.authorization;
   if (token) {
-    return verifyUserToken(req, cb)
-    .then((ret) => ret)
+    return jwtVerifyAsync(token, secret, {})
+    .then((payload) => {
+      if (payload.IsCollector) return verifyCollectorToken(payload, req, cb);
+      if (payload.IsGenerator) return verifyGeneratorToken(payload, req, cb);
+      return verifyUserToken(payload, req, cb);
```
jgraff2: I think this would be clearer if it was an if-else.

jgraff2:
I agree that it's important to distinguish between an invalid token and insufficient permissions, not just for testing but also for usability.

pallavi2209:
Sounds good to me. Seems like currently the mentioned are the only broad categories where we do not want to provide too much details. Let me know if you want to  go ahead with this PR and I can approve it.

iamigo:
I'm going to make those changes and resubmit... hopefully by EOD... so hopefully we can get it reviewed and merged on Monday

iamigo:
@pallavi2209 @jgraff2 updated the PR per your comments

iamigo:
image

user name and token name now reporting correctly from collector and generator

PR #882, opened 2018-06-07
Send sample realtime events when aspect tags change

pallavi2209:
Here is how I tested this:
Created three aspects:
TooHot with tags ['too', 'hot', 'asp1']
TooCold with tags ['too', 'hot', 'asp1']
Aspect3 with tags ['asp3', 'random', 'tag3']

Then created a perspective with filter - include tag "too"
Result - shows samples only for TooHot and TooCold

PATCH /v1/aspects/Aspect3
{
	"tags": ["too"]
}
Result - Perspective shows all samples (Aspect3 is added)

DELETE /v1/aspects/Aspect3/tags/too
Result - shows samples only for TooHot and TooCold (Aspect3 is removed)

Let me know if I missed some use case.

### DIFF
--- a/cache/redisOps.js
+++ b/cache/redisOps.js
@@ -422,6 +422,29 @@ module.exports = {
     return redisClient.batch(cmds).execAsync();
   },
 
+  /**
+   * Get samples for a given aspect
+   * @param  {String} aspName - Aspect Name
+   * @returns {Promise} - Resolves to array of samples
+   */
+  getSamplesFromAspectName(aspName) {
+    const aspsubmapKey = redisStore.toKey(
+      redisStore.constants.objectType.aspSubMap, aspName
+    );
+    const promiseArr = [];
+    return redisClient.smembersAsync(aspsubmapKey)
+    .then((subjNames) => {
+      subjNames.forEach((subjName) => {
+        const sampleName = subjName + '|' + aspName;
+        promiseArr.push(redisClient.hgetallAsync(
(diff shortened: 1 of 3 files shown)


cache/redisOps.js
```diff
@@ -422,6 +422,29 @@ module.exports = {
     return redisClient.batch(cmds).execAsync();
   },
 
+  /**
+   * Get samples for a given aspect
+   * @param  {String} aspName - Aspect Name
+   * @returns {Promise} - Resolves to array of samples
+   */
+  getSamplesFromAspectName(aspName) {
```
pallavi2209: @iamigo I wanted to create a test for this function, but I do not have aspSubjMap in redis to test this. Looks like we are using an independent utility [aspectSubjectMap](https://github.com/salesforce/refocus-utilities/tree/master/src/sampleStore/aspectSubjectMap) to create this set in redis. Any way to work around this?

iamigo: Looks like we're implementing this out of order--that separate utility is to populate the aspsubmap from scratch when there is none, but we do need to update all the places where we will be keeping that map up to date: aspect unpublish, aspect delete, sample post, sample put, sample patch, sample upsert, sample bulk upsert, sample delete. If that doesn't exist, @pallavi2209, can you check for an existing story for it, and if one does not exist, create it?

pallavi2209:
@iamigo Yes, there was a story, pulled that in - We can merge this and I can continue working on that story.

PR #883, opened 2018-06-07
Use short-term feature toggle to control whether we call publishSampl…

iamigo:
…e within the sample upsert's promise chain

or just resolve the promise before we publish.

### DIFF
--- a/worker/jobs/bulkUpsertSamplesJob.js
+++ b/worker/jobs/bulkUpsertSamplesJob.js
@@ -56,17 +56,16 @@ module.exports = (job, done) => {
           return Promise.resolve();
         } else {
           successCount++;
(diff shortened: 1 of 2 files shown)


worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -56,17 +56,16 @@ module.exports = (job, done) => {
           return Promise.resolve();
         } else {
           successCount++;
+          if (featureToggles.isFeatureEnabled('publishSampleInPromiseChain')) {
+            // Wait for publish to complete before resolving the promise.
```
jgraff2: Technically this only waits for the db request to complete, the actual publish is still done asynchronously. I think it would be good to be clear about this in the comments.

Approving since this is a minor concern.

PR #884, opened 2018-06-08
Reassign generators

jgraff2:


api/v1/controllers/collectors.js
```diff
@@ -405,16 +404,10 @@ function startCollector(req, res, next) {
   /* Update or create */
   .then((coll) => coll ? coll.update(body) : helper.model.create(body))
   .then((coll) => {
-    name = coll.name;
     collToReturn = coll;
-    return coll.getCurrentGenerators();
+    /* TODO: change to use currentGenerators once that includes current gens only */
+    return Generator.findAll({ where: { currentCollector: coll.name } });
```
jgraff2: changed because filtering getCurrentGenerators doesn't currently work for Generators that don't don't specify collectors at all.

PR #885, opened 2018-06-08
Reassign generators

jgraff2:
Organized by commit.

This sets up the functions to be used for assigning and reassigning generators.
They aren't actually called anywhere right now, that will come in future stories.

There are also loose ends with the Collector.currentGenerators field and the Generator.isActive field. For now I am doing a find by name instead of using currentGenerators, and I am ignoring isActive until we decide how to handle it.

Details in the "Further Changes" section of the "Reassigning Generators" Quip doc.

### DIFF
--- a/api/v1/controllers/collectors.js
+++ b/api/v1/controllers/collectors.js
@@ -163,11 +163,9 @@ function getCollectorStatus(req, res, next) {
  * @param {Function} next - The next middleware function in the stack
  */
 function patchCollector(req, res, next) {
-  // verify controller token if atleast one field is writable by collector
   let verifyCtrToken = false;
   const reqBodyKeys = Object.keys(req.body);
   const cltrWritableFields = helper.fieldsWritableByCollectorOnly;
-
   for (let i = 0; i < cltrWritableFields.length; i++) {
     const fieldName = cltrWritableFields[i];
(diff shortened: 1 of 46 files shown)


api/v1/controllers/collectors.js
```diff
@@ -405,16 +404,10 @@ function startCollector(req, res, next) {
   /* Update or create */
   .then((coll) => coll ? coll.update(body) : helper.model.create(body))
   .then((coll) => {
-    name = coll.name;
     collToReturn = coll;
-    return coll.getCurrentGenerators();
+    /* TODO: change to use currentGenerators once that includes current gens only */
+    return Generator.findAll({ where: { currentCollector: coll.name } });
```
jgraff2: I changed this because filtering on currentGenerators doesn't work for Generators that don't specify collectors at all.

tests/db/model/collector/methods.js
```diff
@@ -0,0 +1,257 @@
...
+      });
+
+      it('undefined', () => {
+        Collector.build({})
+        .isRunning().should.be.false;
+      });
+
+    });
+
+    describe('isAlive', () => {
```
iamigo: easier to read the test results when every describe ends with ` >`, e.g. `describe('isAlive >'), () => {`

jgraff2:
@harshkothari410 please review.

PR #890, opened 2018-06-18
Postinstall improve performance

harshkothari410:


gulpfile.js
```diff
@@ -86,36 +83,38 @@ gulp.task('browserifyViews', () => {
...
-  rebundle(appFiles.pop());
-});
+  return rebundle(appFiles);
+}
+
+/*
+ * Browserifies the views, generating output under the designated destination
+ * directory.
+ */
+gulp.task('browserifyViews', () =>
```
iamigo: all on one line

gulpfile.js
```diff
@@ -130,9 +129,6 @@ gulp.task('default', conf.tasks.default, () => {
 gulp.task('movecss', () =>
   gulp.src(conf.view.dir + conf.view.cssFiles)
     .pipe(gulp.dest(conf.view.dest))
-    .on('end', () => {
```
iamigo: what do we lose by not calling exit explicitly?

harshkothari410: Because gulp is already calling it. We are abruptly exiting. So we return to gulp then gulp will print log line that task xys is completed in xyz second and it will exit after that.

harshkothari410: screen shot 2018-06-19 at 12 30 06 pm
see this.

harshkothari410:
@iamigo Updated PR.

gulpfile.js
```diff
@@ -86,53 +83,48 @@ gulp.task('browserifyViews', () => {
    * @return {Object} bundle Bundled dependencies.
    */
   function rebundle(pathToApp) {
```
jgraff2: The param name and type should be changed since this is now an array of paths.

jgraff2:
So the improvement comes from doing all the tasks in parallel, instead of waiting for each to finish before moving on to the next, correct?

How long is it taking now compared to before?

harshkothari410:
True. Initially it was recursion after end event so it was taking too much amount of time plus its not written in standard format. Currently browserify done in 13 - 14 second instead of 22 - 25 second. so entire build is done in less then 50 seconds.

jgraff2:
Great!
Did you also look into whether there is any way around running this on every build? Seems like overkill since the result will be the same most of the time.

harshkothari410:
The thing is our project structure is not that good. So I cant use gulp watch and enbale cache things. So once we finalize our UI then we can start with proper react project structure and then thing will be much more faster and easier.

PR #891, opened 2018-06-19
send sample del events when aspect is unpublished or deleted

harshkothari410:


harshkothari410:
@pallavi2209 I have changed your function as well. I verified the behavior but still please do double check.

db/model/aspect.js
```diff
@@ -158,13 +158,28 @@ module.exports = function aspect(seq, dataTypes) {
...
        *
        * @param {Aspect} inst - The instance being destroyed
        * @returns {Promise}
        */
       beforeDestroy(inst /* , opts */) {
-        return inst.checkGeneratorReferences('delete')
-        .then(() => common.setIsDeleted(seq.Promise, inst));
+        const promiseArr = [];
+
+        promiseArr.push(redisOps.getSamplesFromAspectName(inst.name)
```
iamigo: After this operation, if we're pushing arrays onto promiseArr, isn't promiseArr an array of arrays rather than an array of promises? How does seq.Promise.all(arr) deal with an array of arrays?

harshkothari410: I didnt get you. We are pushing promise only.

harshkothari410: redisOps.getSamplesFromAspectName(inst.name) this is promise.

iamigo: oh yeah sorry i was misreading the `.each`

db/model/aspect.js
```diff
@@ -158,13 +158,28 @@ module.exports = function aspect(seq, dataTypes) {
...
+       * samples associated with that aspect for perspectives.
        *
        * @param {Aspect} inst - The instance being destroyed
        * @returns {Promise}
        */
       beforeDestroy(inst /* , opts */) {
-        return inst.checkGeneratorReferences('delete')
-        .then(() => common.setIsDeleted(seq.Promise, inst));
+        const promiseArr = [];
+
```
iamigo: no empty line here

pallavi2209:
@harshkothari410 Have you checked this manually? If yes, on localhost or heroku? Can you share heroku deployment and the steps you did for us to verify?

harshkothari410:
No I checked on Local machine. With changing aspect tags. @pallavi2209

PR #893, opened 2018-06-19
Get config get options

sameesiddiqui:
* use set in getOptions to reduce to O(n) runtime
* only call getOptions in getConfig when key == statusFilter or none of the specified keys
* add test for getOptions functionality


pallavi2209:
@sameesiddiqui Can you please check manually that perspectives are loading and realtime events are coming in.

sameesiddiqui:
@iamigo updated 'getOptions' function naming and added tests per your suggestion. Review when you get the chance

PR #896, opened 2018-06-20
Added response

BrandonMurnane:


api/v1/controllers/bots.js
```diff
@@ -197,8 +197,10 @@ module.exports = {
     u.findByKey(helper, req.swagger.params)
     .then((o) => {
       o.set('lastHeartbeat', timestamp);
+      res.status(httpStatus.OK);
```
iamigo: what if the call to o.save() on the next line results in an error? is it still ok that you already returned "OK"?

BrandonMurnane: we dont react in anyway to the response this is mainly to stop the heroku timeout errors

PR #897, opened 2018-06-20
add new realtime event type for samples with no changes

iamigo:


cache/models/samples.js
```diff
@@ -145,6 +146,23 @@ function cleanAddAspectToSample(sampleObj, aspectObj) {
...
+ * that should not be treated as a change.)
+ *
+ * @param {Object} newSample - received from query body
+ * @param {Object} oldSample - from redis
+ * @returns {Boolean} true if any sample attributes are being modified
+ */
+function isSampleChanged(newSample, oldSample) {
+  const delta = Object.keys(newSample)
+    .filter((key) => key !== 'name')
+    .filter((key) =>
```
pallavi2209: @iamigo We can also try using just one filter instead of two here - it would still be O(n) - but would not hurt to try.

PR #900, opened 2018-06-21
Add heartbeat response

BrandonMurnane:


api/v1/controllers/bots.js
```diff
@@ -197,7 +197,7 @@ module.exports = {
     u.findByKey(helper, req.swagger.params)
     .then((o) => {
       o.set('lastHeartbeat', timestamp);
-      res.status(httpStatus.OK);
+      res.status(httpStatus.OK).json();
```
iamigo: Where's the test which shows that this change makes it work?

BrandonMurnane: @iamigo test added

PR #901, opened 2018-06-22
Automatically deactivate rooms if no recent activity

kriscfoster:
- A lot of rooms are being left active as it is a manual process to deactivate them
- Rooms are automatically deactivated if there has been no events in the last 2 hours
- Enabled with a feature toggle (AUTO_DEACTIVATE_ROOMS)

### DIFF
--- a/clock/index.js
+++ b/clock/index.js
@@ -62,11 +62,8 @@ if (featureToggles.isFeatureEnabled('enablePubStatsLogs')) {
   setInterval(pubStatsLogs.execute, conf.pubStatsLogsIntervalMillis);
 }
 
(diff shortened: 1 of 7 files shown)


clock/index.js
```diff
@@ -61,6 +62,11 @@ if (featureToggles.isFeatureEnabled('enablePubStatsLogs')) {
   setInterval(pubStatsLogs.execute, conf.pubStatsLogsIntervalMillis);
 }
 
+// If enableAutoDeactivateRooms is true then automatically deactivate rooms
```
cassiodias: this comment might be redundant since the if speaks itself

PR #905, opened 2018-06-25
Auto deactivate rooms if no recent activity

kriscfoster:
- A lot of rooms are being left active as it is a manual process to deactivate them
- Rooms are automatically deactivated if there has been no events in the last 2 hours
- Checks every 5 minutes if any active rooms should be deactivated
- Enabled with a feature toggle (AUTO_DEACTIVATE_ROOMS)


view/rooms/list/ListController.js
```diff
@@ -27,7 +27,6 @@ class ListController extends React.Component {
       tableHeaders,
       tableRows,
     } = this.props;
-    tableRows.sort((a, b) => moment(b.updatedAt) - moment(a.updatedAt));
```
tausifmuzaffar: why are you removing this?

clock/scheduledJobs/deactivateRooms.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+}
+
+/**
+ * Execute the call to deactivate rooms with no recent activity
+ *
+ * @returns {Promise} - Promise that rooms were deactivated
+ */
+function execute() {
+  return dbRoom.findAll({ where: { active: true } })
```
tausifmuzaffar: Can you filter by rooms that have been created greater than 2 hours ago?

view/rooms/list/ListController.js
kriscfoster: @tausifmuzaffar updatedAt only gets updated when the room is patched. If we leave this in, all the old rooms that become deactivated will jump to the top of the list. Do we just want to order by id (when it was created), or something else?

kriscfoster:
@tausifmuzaffar done

clock/scheduledJobs/deactivateRooms.js
```diff
@@ -0,0 +1,93 @@
...
+const featureToggles = require('feature-toggles');
+const moment = require('moment');
+const Op = require('sequelize').Op;
+
+const dbRoom = require('../../db/index').Room;
+const dbEvent = require('../../db/index').Event;
+const dbBotAction = require('../../db/index').botAction;
+
+const TWO = 2;
+const THRESHOLD_IN_MINUTES = 120; // 2 hours
```
iamigo: should be coming from config - 2 hrs might seem like a reasonable duration now, but who knows six months from now - you can hard-code a reasonable default in the config file but allow admin to override using env var

clock/scheduledJobs/deactivateRooms.js
```diff
@@ -0,0 +1,93 @@
...
+                isPending: true,
+                parameters: [],
+              };
+
+              dbBotAction.create(syncBotAction);
+            }
+          });
+        }
+
+        return room.update({ active: false });
```
iamigo: Add a message to the timeline that this is being auto-closed?

config.js
```diff
@@ -73,6 +73,8 @@ const DEFAULT_PERSIST_REDIS_SAMPLE_STORE_MILLISECONDS = 120000; // 2min
 
 const botEventLimit = pe.BOT_EVENT_LIMIT || 100;
 
+const deactivateRoomsInterval = pe.DEACTIVATE_ROOMS_INTERVAL || 300000; // 5min
```
iamigo: should be `+pe.DEACTIVATE_ROOMS_INTERVAL` which coerces that env var string to number

config/toggles.js
```diff
@@ -179,6 +179,10 @@ const shortTermToggles = {
   requireHelpEmailOrHelpUrl: environmentVariableTrue(
     pe, 'REQUIRE_HELP_EMAIL_OR_HELP_URL'
   ),
+
+  autoDeactivateRooms: environmentVariableTrue(
```
iamigo: Create a user story to follow up on this short-term toggle to either remove it or make it a long-term toggle?

kriscfoster: I have created the story to remove the short term toggle

tests/clock/deactivateRooms.js
```diff
@@ -0,0 +1,115 @@
...
+    })
+    .then((room) => {
+      testEvent.roomId = room.id;
+      testEvent.createdAt = MOON_LANDING;
+      return Event.create(testEvent);
+    })
+    .then((event) => {
+      deactivateRooms.execute()
+      .then((deactivatedRooms) => {
+        Room.findAll({ where: { active: true } })
```
cassiodias: Any chance of asynchronously a different test create and look just for room active = true? Is there a way to make sure you are checking for a specific room, like ID instead of table size?

kriscfoster: @cassiodias I like this, changed the tests to check whether specific rooms are active/in-active instead of relying on table size

tests/clock/deactivateRooms.js
```diff
@@ -72,15 +72,20 @@ describe('tests/clock/deactivateRooms.js >', () => {
...
-      return Event.create(testEvent);
+      return (Event.create(testEvent));
     })
     .then((event) => {
       deactivateRooms.execute()
       .then((deactivatedRooms) => {
-        Room.findAll({ where: { active: true } })
+        Room.findOne({ where: { id: event.roomId } })
         .then((r) => {
+          /**
```
iamigo: the convention for multiline comments *inside* a function is just one asterisk to open

```
/*
 * my multiline comment
 * goes here
 */
```

You only use two asterisks for module- and function-level multiline comments.

... and thanks for adding the comment here--very helpful!

PR #906, opened 2018-06-25
Bulk Subject Delete

cassiodias:
I have created a new branch and PR to this feature - Tried to make sure that I didn't add any side effect. 

Just added my methods, test, and changes leaving even the variable names on controller/subject.js as the original.

test/subject/bulkDelete.js test now is working locally and on Travis - apparently the order that I was creating the subjects and using before / after utils.populateRedis, utils.forceDelete , testUtils.forceDeleteUser was affecting different tests.


api/v1/controllers/subjects.js
```diff
@@ -535,9 +535,6 @@ module.exports = {
    * DELETE /v1/subjects/{key}/relatedLinks/
    * DELETE /v1/subjects/{key}/relatedLinks/{name}
    *
-   * Deletes specified/all related links from the subject and sends updated
```
cassiodias: Already reverted that, did this just to check a new branch as a copy of master works on Travis

api/v1/controllers/subjects.js
```diff
@@ -24,7 +24,7 @@ const doGetHierarchy = require('../helpers/verbs/doGetHierarchy');
 const doPatch = require('../helpers/verbs/doPatch');
 const doPost = require('../helpers/verbs/doPost');
 const doPut = require('../helpers/verbs/doPut');
-const u = require('../helpers/verbs/utils');
```
cassiodias: Added my renaming stuffs, but next commit I have changed to be like the original master convention (using the variable as 'u');
Just added my stuff on next commit, leaving the code as much close to the original master branch

tests/api/v1/subjects/bulkDelete.js
```diff
@@ -0,0 +1,159 @@
...
+
+describe('tests/api/v1/subjects/bulkDelete.js', () => {
+  let token;
+  const AUTHORIZATION = 'Authorization';
+  const DELETE_PATH = '/v1/subjects/delete/bulk';
+
+  const parentAndChildIdStore = [];
+  const fooAndBlahIdStore = [];
+
+  const TIME_OUT = 100;
```
iamigo: "timeout" is one word - no underscore needed

worker/jobs/bulkDeleteSubjectsJob.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/**
+ * Executes bulk subject deletion.
+ *
+ * @param {Array} - List of subject keys, meaning that key is
+ *  a string either subject id or absolutePath.
+ * @param {Array} - nouns/subjects/readOnlyFields
+ * @param {Object} - user
+ * @returns {[Promise]} - Indicates success or failure for each
```
iamigo: if this is returning an array of promises, then it would be `[{Promise}]`, but also worth spelling out, e.g. `an array of Promises where each promise resolves to indicate the success or failure deleting one individual subject` (or something like that)

PR #908, opened 2018-06-27
Saving Bot-Layout as a cookie

kriscfoster:
- Saving bot-layout as a cookie so when users move the bots around and then reload the page, their changes will stay


package.json
```diff
@@ -110,7 +110,7 @@
     "html-webpack-plugin": "^2.16.0",
     "if-env": "^1.0.4",
     "ip": "^1.1.5",
-    "joi": "^13.4.0",
+    "joi": "^13.3.0",
```
cassiodias: Downgrade? Should we do that? I guess the latest version from master should be kept.

PR #909, opened 2018-06-28
Getting the layout of bots from a cookie

kriscfoster:
- After someone moves the bots around in a room, this layout is saved as a cookie in the browser.
- When they reload the page, the layout will be determined from this cookie.


view/rooms/utils/page.js
```diff
@@ -75,4 +76,63 @@ module.exports = {
...
+      if (layout[param] && layout[param].length
+        && layout[param].every(bot => bots.indexOf(bot) === NEG_ONE)) {
+        return false;
+      }
+    }
+
+    return true;
+  },
+
+  saveLayoutAsCookie() {
```
iamigo: This feels like both a "serialize" function for bot layout and a function to save in cookie. Would it be useful to actually have serialize and deserialize utility functions for bot layout?

tests/view/rooms/utils/page.js
```diff
@@ -0,0 +1,46 @@
...
+      rightColumn: ['BOT D'],
+    };
+    const botsInRoom = ['BOT A', 'BOT B'];
+    expect(utils.botLayoutIsValid(botLayout, botsInRoom)).to.equal(false);
+  });
+
+  it('fail, incorrect format of Bot Layout obj', () => {
+    const botLayout = {
+      leftCol: ['BOT A', 'BOT B'],
+      middleCol: [],
```
cassiodias: What happens with numbers or empty string?

view/rooms/app.js
```diff
@@ -935,8 +936,18 @@ window.onload = () => {
...
+      u.getCookie(`${window.location.pathname}-bots-layout`);
+    if (layoutCookie) {
+      layoutCookie = JSON.parse(layoutCookie);
+      // Checking if the layout that came from the cookie is valid
+      if (uPage.botLayoutIsValid(layoutCookie, room.bots)) {
+        _botsLayout = layoutCookie;
+        // Reordering bots so they will be in the correct layout
+        room.bots = _botsLayout.leftColumn.concat(_botsLayout.middleColumn)
+          .concat(_botsLayout.rightColumn);
+      }
```
cassiodias: what happens if the boot is not valid?

view/rooms/utils/page.js
```diff
@@ -75,4 +76,73 @@ module.exports = {
...
+
+  /**
+   * Checks whether a layout object is valid for an array of Bots
+   * @param {Object} layout - Layout object
+   * @param {Array} bots - Bots in the room
+   * @returns {boolean}
+   */
+  botLayoutIsValid(layout, bots) {
+    // Layout does not contain required field
+    if (!layout || !layout.leftColumn || !layout.middleColumn ||
```
cassiodias: sometimes is better having a boolean function with a meaningful name (isValidLayout(layout)) instead of the comment to explain the long IF.

Or: if (layout.isvalid()) being more OO.

view/rooms/utils/page.js
```diff
@@ -75,4 +76,73 @@ module.exports = {
...
+    }
+
+    for (const param in layout) {
+      // If a bot in the layout does not actually exist in the room
+      if (layout[param] && layout[param].length
+        && layout[param].every(bot => bots.indexOf(bot) === NEG_ONE)) {
+        return false;
+      }
+    }
+
```
cassiodias: I feel that all those validations shouldn't be page.js responsibility, don't know if by design would be OK add into a layout or create something like layout validator. In that way, we would have a more cohesive object, more testable and easy to reuse.

view/rooms/app.js
kriscfoster: It will just use the default layout and wont use the one defined in the cookie

view/rooms/utils/page.js
kriscfoster: @cassiodias That makes sense incase we ever add new functionality to 'layout'. Are you suggesting I create another file in the utils folder called "layout.js" to do all of this or something else?

BrandonMurnane:
i asked my questions in person

view/rooms/utils/layout.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    return true;
+  },
+
+  /**
+   * Gets the bot layout from the page and returns as an object.
+   * @returns {{leftColumn: Array, middleColumn: Array, rightColumn: Array}}
+   */
+  getBotLayoutFromPage() {
```
iamigo: no unit tests for this utility fn?

view/rooms/utils/layout.js
```diff
@@ -0,0 +1,86 @@
...
+      Array.from(rightColumn.children).forEach((c) => {
+        botsLayout.rightColumn.push(c.id.replace('-section', ''));
+      });
+    }
+
+    return botsLayout;
+  },
+
+  // Gets bot layout from page and saves the stringified version as a cookie
+  getLayoutAndSaveAsCookie() {
```
iamigo: no unit tests for this utility fn?

PR #911, opened 2018-06-29
Generator isActive changes

jgraff2:
- change the swagger description to more clearly express what isActive means.
- validate: add validation so that if a generator is active, it must have collectors specified.
- createWithCollectors/updateWithCollectors: restructure to make it work with the new validation
- beforeUpdate: when isActive is changed, assign/unassign the collector.
- assignToCollector:
  - only assign if isActive
  - don't actually save the generator, just update the currentCollector field
  - remove generic assignment (if no collectors specified, set to null)


db/model/generator.js
```diff
@@ -446,22 +447,12 @@ module.exports = function generator(seq, dataTypes) {
    * @returns {Promise<Generator>}
    */
   Generator.prototype.assignToCollector = function () {
```
jgraff2: Note that this doesn't return a promise anymore, or save to the db. I figured this would make more sense for being called from generator patch/put and collector start where it's already saving the generator later, so we're not saving to the db multiple times.

iamigo: this should be comment in the code

db/model/collector.js
```diff
@@ -253,7 +242,10 @@ module.exports = function collector(seq, dataTypes) {
   Collector.prototype.reassignGenerators = function () {
     /* TODO: change to use currentGenerators once that includes current gens only */
     return seq.models.Generator.findAll({ where: { currentCollector: this.name } })
-    .then((gens) => Promise.all(gens.map((g) => g.assignToCollector())));
+    .map((g) => {
```
iamigo: is the TODO on line 243 still valid?

db/model/generator.js
```diff
@@ -167,6 +167,10 @@ module.exports = function generator(seq, dataTypes) {
       beforeUpdate(inst /* , opts */) {
         const gtName = inst.generatorTemplate.name;
         const gtVersion = inst.generatorTemplate.version;
+        if (inst.changed('isActive')) {
```
iamigo: should this `if` also be checking whether the new value of `isActive` is `true`?

db/model/collector.js
jgraff2: Yes. Once @sameesiddiqui 's change is done we can replace the findAll with this.currentGenerators

db/model/generator.js
jgraff2: No. If isActive was turned off we also need to call assignToCollector to set currentCollector to null.

jgraff2:
Don't merge this yet.
I realized there's a problem with the validation. I'm working on fixing it now.

jgraff2:
Updated.

jgraff2:
@sameesiddiqui

PR #912, opened 2018-06-29
Call generator.assignToCollector on gen post

pallavi2209:


api/v1/controllers/generators.js
```diff
@@ -178,6 +178,7 @@ module.exports = {
     validateGeneratorAspectsPermissions(toPost.aspects, req)
     .then(() =>
       helper.model.createWithCollectors(toPost))
+    .then((o) => o.assignToCollector())
```
jgraff2: Note that in my PR I changed assignToCollector so it no longer saves the changes - it just sets the currentCollector field and expects the calling function to do the saving later. I was thinking that would make more sense for patch and put where it's already saving the generator later, so we're not saving to the db multiple times.

So after merging my changes you would also need to add a save here after calling assignToCollector. If after figuring out the logic for the other methods you think it's better the way it was before, I can change it back easily.

jgraff2:
I think it's important that all your changes are in the same PR, since this will be "turning on" the generator assignment and each piece is incomplete without the others.

api/v1/controllers/generators.js
iamigo: Make sure there are comments in the code describing all those choices

pallavi2209:
Okay sounds good

PR #914, opened 2018-06-29
Multiple associations Sequelize "bug"

cassiodias:


cassiodias:
Fixed by PR https://github.com/salesforce/refocus/pull/921

PR #916, opened 2018-07-02
Bot display name

kriscfoster:
- We want to be able to give bots display names that can be displayed in their header instead of their official name
- On install/update of a bot, this display name can be updated
- If a nickname exists for a bot, it will be shown in the header
- In the below example, Communications-Bot has a display name ("Communications Bot Nickname") but Timeline-Bot does not
screen shot 2018-07-02 at 12 07 24

### DIFF
--- a/api/v1/swagger.yaml
+++ b/api/v1/swagger.yaml
@@ -2068,8 +2068,9 @@ paths:
         -
           name: displayName
           in: formData
-          description: The display name of the bot (Name displayed in the header).
+          description: The name displayed in the bot header.
           type: string
+          maxLength: 60
           required: false
         -
           name: actions
@@ -2220,7 +2221,8 @@ paths:
                 description: The version of the bot. Use Semantic Versioning 2.0.0
               displayName:
                 type: string
-                description: The display name of the bot. Displayed in the header of the bot.
+                maxLength: 60
+                description: The name displayed in the bot header.
               actions:
                 type: array
                 description: Array of objects that contain the name of actions a bot can take and an array of parameters needed for that action
@@ -2288,8 +2290,9 @@ paths:
         -
           name: displayName
           in: formData
-          description: The display name of the bot. Displayed in the header.
+          description: The name displayed in the bot header.
           type: string
+          maxLength: 60
         -
           name: ui
           description: The zipped HTML and Javascript needed to run the UI of the bot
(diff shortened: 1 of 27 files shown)


api/v1/swagger.yaml
```diff
@@ -2065,6 +2065,12 @@ paths:
           description: The version of the bot. Use Semantic Versioning 2.0.0
           type: string
           required: true
+        -
+          name: displayName
+          in: formData
+          description: The display name of the bot (Name displayed in the header).
```
iamigo: `The name displayed in the bot header`

api/v1/swagger.yaml
```diff
@@ -2212,6 +2218,9 @@ paths:
               version:
                 type: string
                 description: The version of the bot. Use Semantic Versioning 2.0.0
+              displayName:
+                type: string
+                description: The display name of the bot. Displayed in the header of the bot.
```
iamigo: `The name displayed in the bot header`

api/v1/swagger.yaml
```diff
@@ -2276,6 +2285,11 @@ paths:
           in: formData
           description: The version of the bot. Use Semantic Versioning 2.0.0
           type: string
+        -
+          name: displayName
+          in: formData
+          description: The display name of the bot. Displayed in the header.
```
iamigo: `The name displayed in the bot header`

db/model/bot.js
```diff
@@ -94,6 +94,11 @@ module.exports = function bot(seq, dataTypes) {
       allowNull: false,
       comment: 'Bot Version. Use Semantic Versioning ^2.0.0',
     },
+    displayName: {
+      type: dataTypes.STRING,
+      allowNull: true,
+      comment: 'Displayed in the header of the Bot',
```
iamigo: `The name displayed in the bot header`

db/model/bot.js
```diff
@@ -94,6 +94,11 @@ module.exports = function bot(seq, dataTypes) {
       allowNull: false,
       comment: 'Bot Version. Use Semantic Versioning ^2.0.0',
     },
+    displayName: {
+      type: dataTypes.STRING,
```
iamigo: should have max length assigned here using one of the constants
... and use the length in swagger regex validation too
... and add tests for max length

PR #918, opened 2018-07-03
Fixing Sequelize error when specifying multiple associations as field params

cassiodias:
* Changed method **buildFieldList** to add unique list of FK avoiding Sequelize failures;
** Unique because sometimes the FK is already as attribute of option object (req.params);
** Created list of attributes to remove as well before returning back to client (as @jgraff2 suggested);
* Changes on doFind/doGet and some other controllers to generate the list of attributes and cleaning up them before sending back to the client keeping the API as expected;
* Minor clean up and lint warnings;


api/v1/controllers/subjects.js
```diff
@@ -326,8 +329,10 @@ module.exports = {
     } else {
       doGetHierarchy(resultObj)
       .then((resultObj) => {
-        u.logAPI(req, resultObj, resultObj.retval);
-        res.status(httpStatus.OK).json(resultObj.retval);
+        const subject = resultObj.retval;
+        u.removeFieldsFromResponse(opts.fieldsToExclude, subject);
```
iamigo: This is only going to remove the fields from the top-level subject in the hierarchy, right? What about n-depth children?

api/v1/helpers/verbs/doFind.js
```diff
@@ -71,10 +71,15 @@ function doFindResponse(reqResNext, props, opts, resultObj) {
   .then((retval) => {
     u.sortArrayObjectsByField(props, retval);
 
+    // Merge props and opt fields to exclude
+    let fieldsToExclude = opts.fieldsToExclude ? opts.fieldsToExclude : [];
+    fieldsToExclude = props.fieldsToExclude ? fieldsToExclude.fieldsToExclude
```
iamigo: I think lines 75-76 would be easier to read like this:

```
const fieldsToExclude = [].concat(opts.fieldsToExclude || [],
  props.fieldsToExclude || [])
```

api/v1/helpers/verbs/doGet.js
```diff
@@ -37,49 +37,57 @@ function doGet(req, res, next, props) {
...
       if (cacheErr || !reply) {
         // if err or no reply, get from db and set redis cache
-        u.findByKey(props, req.swagger.params, scopes)
-        .then((o) => {
+        u.findByKey(props, reqParams, scopes)
+        .then((returnedObject) => {
           resultObj.dbTime = new Date() - resultObj.reqStartTime;
-          u.logAPI(req, resultObj, o);
-          res.status(httpStatus.OK).json(u.responsify(o, props, req.method));
+          u.logAPI(req, resultObj, returnedObject);
```
iamigo: Logging has to happen AFTER removing fields from response because we're logging response size.

api/v1/helpers/verbs/doGetWriters.js
```diff
@@ -61,10 +61,12 @@ function getWriter(req, res, next, props) {
 function getWriters(req, res, next, props) {
   const resultObj = { reqStartTime: req.timestamp };
   const params = req.swagger.params;
+  const opts = u.buildFieldList(params, props.model);
   u.findAssociatedInstances(props, params, props.belongsToManyAssoc.users, {})
   .then((o) => {
-    resultObj.dbTime = new Date() - resultObj.reqStartTime;
     const retval = u.responsify(o, props, req.method);
+    u.removeFieldsFromResponse(opts.fieldsToExclude, retval);
+    resultObj.dbTime = new Date() - resultObj.reqStartTime;
```
iamigo: Calculating the db time should not include the time we spent removing fields from response, so it should come immediately after line 66

api/v1/helpers/verbs/utils.js
```diff
@@ -238,20 +240,46 @@ function handleAssociations(reqObj, inst, props, method) {
...
-function buildFieldList(params) {
+function buildFieldList(params, model) {
   const opts = {};
   if (params.fields && params.fields.value) {
     opts.attributes = params.fields.value;
     if (!opts.attributes.includes('id')) {
       opts.attributes.push('id');
     }
+
+    /*
```
iamigo: fits on one line so should be formatted as single-line comment

`/* Includes all model's FK for an eventual outer join. */`

api/v1/helpers/verbs/utils.js
```diff
@@ -238,20 +240,46 @@ function handleAssociations(reqObj, inst, props, method) {
...
+     * Includes all model's FK for an eventual outer join.
+     */
+    if (model) {
+      opts.fieldsToExclude = [];
+
```
iamigo: why empty line here? i don't feel like it adds to readability (but it's fine if you do)

api/v1/helpers/verbs/utils.js
```diff
@@ -238,20 +240,46 @@ function handleAssociations(reqObj, inst, props, method) {
...
+      const keys = Object.keys(model.attributes);
+      keys.forEach((key) => {
+        if (model.attributes && model.attributes[key].references) {
+          if (opts.attributes.indexOf(key) === NOT_FOUND) {
+            opts.attributes.push(key);
+          }
+          /*
```
iamigo: code style: empty line after `if` block

api/v1/controllers/subjects.js
cassiodias: It's just removing from the top-level. Seems that this methods is always removing the root object. Even when there is a list of objects, it iterates and removes attributes from the top-level.

DoFind.js, method doFindAndCountAll:
```
for (let i = retval.length - 1; i >= 0; i--) {
 u.removeFieldsFromResponse(fieldsToExclude, retval[i]);
}
```

Do you think we need to change that? Should we create a story?

By the way, I feel this method name is not clear. It is not removing from response, it is removing FOR response in that case (doing something to help / benefit). Also, in fact it should be more generic as the documentation describes it: "In-place removal of certain keys from the input object", so it might be worth to remove the response and it helps to remove attributes for a given Object (even now we are just using to give it back to the client).

api/v1/helpers/verbs/doFind.js
cassiodias: Wow, awesome. Thanks! Really like the 'defensive' way in JS using "|| default_value". Will try next time 🙂

api/v1/helpers/verbs/doGet.js
cassiodias: Done.

api/v1/helpers/verbs/doGetWriters.js
cassiodias: Done

api/v1/controllers/subjects.js
iamigo: feel free to rename the fn to make it more clear

the hierarchy response is a special case because the structure is hierarchical, where any subject can contain an array of subjects (the "children" attribute) -- *IF* we are ever removing attributes from a subject in the response from a call to `GET /v1/subjects/{key}/ hierarchy` then we need to traverse the hierarchy and remove those fields from every subject, not just the root-level subject

tests/api/v1/generators/multipleAssociations.js
```diff
@@ -0,0 +1,170 @@
...
+        updatedAt: Joi.string().required(),
+        GeneratorCollectors: Joi.object().required(),
+      })),
+  };
+
+  after(u.forceDelete);
+  after(gtUtil.forceDelete);
+  after(tu.forceDeleteUser);
+
+  describe('Checking multiple associations', () => {
```
jgraff2: do we need this file? Isn't this covered in common/testAssociations?

api/v1/helpers/verbs/utils.js
```diff
@@ -238,20 +240,41 @@ function handleAssociations(reqObj, inst, props, method) {
...
     }
+
+    /* Includes all model's FK for an eventual outer join. */
+    if (model) {
+      opts.fieldsToExclude = [];
+      const keys = Object.keys(model.attributes);
+      keys.forEach((key) => {
+        if (model.attributes && model.attributes[key].references) {
+          if (opts.attributes.indexOf(key) === NOT_FOUND) {
+            opts.attributes.push(key);
```
jgraff2: what happened to only adding the key when necessary? (when multiple keys are references) I don't think we should be adding the fields to the db request for every model when it's only needed for Generators.

api/v1/helpers/verbs/utils.js
```diff
@@ -238,20 +240,41 @@ function handleAssociations(reqObj, inst, props, method) {
...
-function buildFieldList(params) {
+function buildFieldList(params, model) {
   const opts = {};
   if (params.fields && params.fields.value) {
     opts.attributes = params.fields.value;
     if (!opts.attributes.includes('id')) {
       opts.attributes.push('id');
     }
+
+    /* Includes all model's FK for an eventual outer join. */
```
jgraff2: Can you make it clear that this is a workaround? (and hopefully temporary)

api/v1/controllers/subjects.js
jgraff2: the fields param only affects the top-level anyway, so we don't need to remove what wasn't added.

api/v1/helpers/verbs/doFind.js
```diff
@@ -71,10 +71,14 @@ function doFindResponse(reqResNext, props, opts, resultObj) {
   .then((retval) => {
     u.sortArrayObjectsByField(props, retval);
 
+    // Merge props and opt fields to exclude
+    const fieldsToExclude = [].concat(opts.fieldsToExclude || [],
+      props.fieldsToExclude || []);
```
jgraff2: why do we need to use both? why not just use the existing props.fieldsToExclude?

api/v1/controllers/lenses.js
```diff
@@ -287,7 +287,10 @@ module.exports = {
           extraAttributes = ['lensLibrary'];
         }
 
-        u.findByKey(helper, req.swagger.params, extraAttributes)
+        const params = req.swagger.params;
+        const opts = u.buildFieldList(params, helper.model);
+
+        u.findByKey(helper, params, extraAttributes)
```
jgraff2: seems like overkill to have to change the subject and lens controllers when the issue never happens on those models.

tests/api/v1/generators/multipleAssociations.js
cassiodias: True, will remove it.

api/v1/helpers/verbs/utils.js
cassiodias: How would you define "when necessary"? What happens if it is not a Generator and we got multiple keys? Should we check "if model.name === "generator""?

I am assuming this as a general workaround when we have sub-query using FK.

api/v1/helpers/verbs/utils.js
cassiodias: Sure.

api/v1/helpers/verbs/doFind.js
cassiodias: The method who is generating the list of attributes is utils.buildFieldList - there is supposedly the place to generate a list of attribute and now the list to remove (I have added the model as an argument. Would you change the model by reference inside that method? To me, it will do more than just build a list of Sequelize attributes (including or excluding) and might cause a side effect.

Is it ok changing object by reference as a refocus design definition?

cassiodias:
fixed by second pull request

PR #919, opened 2018-07-05
Call generator assignment methods

pallavi2209:
- Generator beforeCreate (post): generator.assignToCollector()
- Generator beforeUpdate (put): generator.assignToCollector() - if the collectors field is changed such that the currentCollector is no longer valid 
- Collector afterCreate (start): generator.assignToCollector() - for any currently unassigned generators 
- Collector afterUpdate (resume): when status is changed to running, generator.assignToCollector() - for any currently unassigned generators 
- Collector afterUpdate (stop): when status is changed to stopped, collector.reassignGenerators()
- Collector afterUpdate (pause): when status is changed to paused, collector.reassignGenerators()

### DIFF
--- a/api/v1/controllers/collectors.js
+++ b/api/v1/controllers/collectors.js
@@ -10,7 +10,6 @@
  * api/v1/controllers/collectors.js
  */
 'use strict'; // eslint-disable-line strict
-const featureToggles = require('feature-toggles');
 const Promise = require('bluebird');
 const jwtUtil = require('../../../utils/jwtUtil');
 const apiErrors = require('../apiErrors');
@@ -402,20 +401,37 @@ function startCollector(req, res, next) {
 
     return coll;
   })
-  /* Update or create */
-  .then((coll) => coll ? coll.update(body) : helper.model.create(body))
-  .then((coll) => {
-    collToReturn = coll;
-    /* TODO: change to use currentGenerators once that includes current gens only */
 
-    // return Generator.findAll({ where: { currentCollector: coll.name } });
-    /*
-     * TODO: this actually maps to Generator.possibleCollectors, not currentCollector.
-     * This is necessary for now since currentCollector is not being set yet.
-     * Change to use only current collectors once all the assignment logic is in place.
-     */
-    return coll.getPossibleGenerators();
+  /* Update or create collector. If collector is updated, find active and
+     unassigned generators. For each generator, assign to collector. */
+  .then((coll) => {
+    // Set lastHeartbeat to make collector alive for generators to be assigned
+    body.lastHeartbeat = Date.now();
+    if (coll) {
+      return coll.update(body)
+      .then((updatedColl) => {
+        collToReturn = updatedColl;
+        return Generator.findAll(
+          { where: { currentCollector: null, isActive: true } }
+        );
+      })
+      .then((unassignedGenerators) => Promise.map(unassignedGenerators,
+        (generator) => {
+          generator.assignToCollector();
+          return generator.save();
+        }));
+    }
+
+    return helper.model.create(body)
+    .then((createdColl) => {
+      collToReturn = createdColl;
+      return Promise.resolve();
+    });
   })
+  /* TODO: change to use currentGenerators once that includes current gens only */
+  .then(() => Generator.findAll(
+    { where: { currentCollector: collToReturn.name } })
+  )
   /* Add all the attributes necessary to send back to collector. */
   .then((gens) => Promise.all(gens.map((g) => g.updateForHeartbeat())))
   .then((gens) => {
@@ -448,7 +464,7 @@ function startCollector(req, res, next) {
  * POST /collectors/{key}/stop
  *
  * Change collector status to Stopped. Invalid if the collector's status is
- * Stopped.
+ * Stopped. Reassign corresponding generators.
  *
  * @param {IncomingMessage} req - The request object
  * @param {ServerResponse} res - The response object
@@ -458,14 +474,33 @@ function stopCollector(req, res, next) {
   req.swagger.params.queryBody = {
     value: { status: 'Stopped' },
   };
-  doPatch(req, res, next, helper);
+
+  const resultObj = { reqStartTime: req.timestamp };
+  const requestBody = req.swagger.params.queryBody.value;
+
+  let returnValue;
+
+  u.findByKey(helper, req.swagger.params)
+  .then((o) => u.isWritable(req, o))
+  .then((o) => {
+    u.patchJsonArrayFields(o, requestBody, helper);
+    u.patchArrayFields(o, requestBody, helper);
+
+    return o.update(requestBody);
+  })
+  .then((retVal) => {
+    returnValue = retVal;
+    return retVal.reassignGenerators();
+  })
+  .then(() => u.handleUpdatePromise(resultObj, req, returnValue, helper, res))
+  .catch((err) => u.handleError(next, err, helper.modelName));
 } // stopCollector
 
 /**
  * POST /collectors/{key}/pause
  *
  * Change collector status to Paused. Invalid if the collector's status is not
- * Running.
+ * Running. Reassign corresponding generators.
  *
  * @param {IncomingMessage} req - The request object
  * @param {ServerResponse} res - The response object
@@ -475,7 +510,25 @@ function pauseCollector(req, res, next) {
   req.swagger.params.queryBody = {
     value: { status: 'Paused' },
   };
-  doPatch(req, res, next, helper);
+
+  const resultObj = { reqStartTime: req.timestamp };
+  const requestBody = req.swagger.params.queryBody.value;
+  let returnValue;
+
+  u.findByKey(helper, req.swagger.params)
+  .then((o) => u.isWritable(req, o))
+  .then((o) => {
+    u.patchJsonArrayFields(o, requestBody, helper);
+    u.patchArrayFields(o, requestBody, helper);
+
+    return o.update(requestBody);
+  })
+  .then((retVal) => {
+    returnValue = retVal;
+    return retVal.reassignGenerators();
+  })
+  .then(() => u.handleUpdatePromise(resultObj, req, returnValue, helper, res))
+  .catch((err) => u.handleError(next, err, helper.modelName));
 } // pauseCollector
 
 /**
--- a/api/v1/controllers/collectors.js
+++ b/api/v1/controllers/collectors.js
@@ -10,6 +10,7 @@
  * api/v1/controllers/collectors.js
  */
 'use strict'; // eslint-disable-line strict
+const featureToggles = require('feature-toggles');
 const Promise = require('bluebird');
 const jwtUtil = require('../../../utils/jwtUtil');
 const apiErrors = require('../apiErrors');
@@ -375,6 +376,9 @@ function startCollector(req, res, next) {
   const body = req.swagger.params.queryBody.value;
   body.status = 'Running';
   body.createdBy = req.user.id;
+
+  // Set lastHeartbeat to make collector alive for generators to be assigned
+  body.lastHeartbeat = Date.now();
   let collToReturn;
   return helper.model.findOne({ where: { name: body.name } })
   /* Already exists? Verify that this user has write permission. */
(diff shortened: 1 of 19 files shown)


api/v1/swagger.yaml
```diff
@@ -3636,6 +3636,11 @@ paths:
                 readOnly: true
                 description: >-
                   Filter by the name of current collector assigned to the generator.
+              isActive:
+                type: boolean
+                description: >-
+                  Default false. When set to true, the generator will be 
+                  assigned to an available collector and regularly send in samples.
```
iamigo: `available collector` --> `running collector` ?

api/v1/swagger.yaml
```diff
@@ -3757,7 +3762,12 @@ paths:
                 type: string
                 readOnly: true
                 description: >-
-                  Filter by the name of current collector assigned to the generator.
+                  Name of current collector assigned to the generator.
+              isActive:
+                type: boolean
+                description: >-
+                  Default false. When set to true, the generator will be 
+                  assigned to an available collector and regularly send in samples.
```
iamigo: `running` instead of `available`?

api/v1/controllers/collectors.js
```diff
@@ -402,20 +402,51 @@ function startCollector(req, res, next) {
...
+     collectors list contains this collector. */
   .then((coll) => {
-    collToReturn = coll;
-    /* TODO: change to use currentGenerators once that includes current gens only */
+    if (coll) {
+      return coll.update(body)
+      .then((updatedColl) => {
+        collToReturn = updatedColl;
+        return Generator.findAll(
+          { where: { currentCollector: null, isActive: true } }
```
iamigo: after Jon's changes, can isActive be true if currentCollector is null?

api/v1/controllers/collectors.js
```diff
@@ -458,14 +489,28 @@ function stopCollector(req, res, next) {
...
+  u.findByKey(helper, req.swagger.params)
+  .then((o) => u.isWritable(req, o))
+  .then((o) => {
+    u.patchJsonArrayFields(o, requestBody, helper);
+    u.patchArrayFields(o, requestBody, helper);
+
+    return o.update(requestBody);
+  })
+  .then((retVal) => retVal.reassignGenerators()
+    .then(() => u.handleUpdatePromise(resultObj, req, retVal, helper, res)))
```
iamigo: does this promise need to be nested or can we unwind the nesting?

api/v1/controllers/collectors.js
```diff
@@ -475,7 +520,21 @@ function pauseCollector(req, res, next) {
...
```
iamigo: does this promise need to be nested or can we unwind the nesting?

api/v1/controllers/collectors.js
```diff
@@ -458,14 +489,28 @@ function stopCollector(req, res, next) {
   req.swagger.params.queryBody = {
     value: { status: 'Stopped' },
   };
-  doPatch(req, res, next, helper);
+
```
jgraff2: I didn't realize we would have to duplicate the code from doPatch here in order to make this work.
I'm wondering if it might be better to call reassignGenerators in db hooks instead?

api/v1/controllers/generators.js
```diff
@@ -228,6 +233,17 @@ module.exports = {
...
+    .then((_retInstance) => {
+      /* No need to assign collector if collectors list provided in request
+      already has current collector */
+      if (Array.isArray(collectors) &&
+       collectors.includes(_retInstance.currentCollector)) {
+        return _retInstance;
+      }
+
+      _retInstance.assignToCollector();
+      return _retInstance.save();
```
jgraff2: So you weren't able to make it work without calling save right after?
In that case we should probably change back assignToCollector to also save the changes since everywhere it's being called from is just calling save() right after. The one tricky place is in the beforeUpdate hook where it calls assignToCollector if isActive has changed. We would have to move that to the api layer.

Or we could do all the assignment logic in the db layer instead. I'm starting to think that would be better since calling it from the api isn't as straightforward as I expected it to be. Thoughts?

api/v1/controllers/collectors.js
pallavi2209: We can unwind it at the cost of storing the results in a variable. Will update it if thats preferable.

api/v1/controllers/collectors.js
pallavi2209: So, in collector afterUpdate hook, we check if the status is changed to 'Paused' or 'Stopped', then we call reassignGenerators? Sounds good to me. I don't see any disadvantage.

api/v1/controllers/generators.js
pallavi2209: "So you weren't able to make it work without calling save right after?" - No, it would have involved going through updateInstance method (which is called by almost all update api's), which is very clean right now and calls save() at last - my changes would have added various checks in the method and would have made it more complex. So, I ended up not touching that.
I can still try and take that route if we are okay with that. What do you think?

"Or we could do all the assignment logic in the db layer instead" - You mean only this logic when we PUT? We can do that, but I feel its more cleaner in api's for this case because we need this only in PUT method (when currentCollector can be deleted from list of collectors). If we add this logic in db hooks, we will be executing the check for all updates. 

"The one tricky place is in the beforeUpdate hook where it calls assignToCollector if isActive has changed. We would have to move that to the api layer." >
How about adding another parameter 'isSavedAfterAssignment' to assignToCollector?

api/v1/controllers/collectors.js
```diff
@@ -402,20 +402,51 @@ function startCollector(req, res, next) {
...
+        return Generator.findAll(
+          { where: { currentCollector: null, isActive: true } }
+        );
+      })
+      .then((unassignedGenerators) => {
+        unassignedGenerators.forEach((generator) => {
+          if (generator.collectors) {
+            for (let i = 0; i < generator.collectors.length; i++) {
+              const genCollector = generator.collectors[i];
+              if (genCollector.name === collToReturn.name) {
```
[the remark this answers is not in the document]
pallavi2209: The test already exists - I will update it.

pallavi2209:
@iamigo Updated PR as per comments. @jgraff2 Updated PR as per some comments. Added some comments for further discussion.

api/v1/controllers/collectors.js
jgraff2: yeah.

api/v1/controllers/generators.js
jgraff2: Makes sense, I can see why it's a lot more complicated than I was thinking it would be when I changed assignToCollector to not save. I think it's fine as-is.

jgraff2: We would be executing the check for all updates, but we can do the check first-thing so it wouldn't really be any additional work:
if (inst.changed('collectors') && !inst.collectors.includes(inst.currentCollector))

jgraff2: Adding a parameter would work, but I think it would overly complicate things. I think it would be easier to just move it.

The reason there's a conflict there is because it needs to be one way if it's done in the controller and another way if done in db hooks, so I think always doing it in the same place is the best solution.

api/v1/controllers/collectors.js
jgraff2: And we could also check if it was changed to 'Started', and move the assignment logic from above.

db/helpers/collectorUtils.js
```diff
@@ -82,8 +82,26 @@ function validateVersion(version) {
...
+ * @param {Object} seq - Sequelize object
+ * @returns {Promise} - Resolves to array of assigned generator db objects
+ */
+function findAndAssignGenerators(seq) {
+  return seq.models.Generator.findAll(
+    { where: { currentCollector: null, isActive: true } }
+  )
+  .then((unassignedGenerators) =>
+    Promise.all(unassignedGenerators.map((g) => {
+      g.assignToCollector();
```
iamigo: I feel like this assignToCollector (which calls seq "find") should return a promise which should be resolved before calling g.save ... otherwise do we risk saving g *before* a collector has actually been assigned?

db/model/collector.js
```diff
@@ -124,6 +131,23 @@ module.exports = function collector(seq, dataTypes) {
           throw new ValidationError(msg);
         }
       }, // hooks.beforeUpdate
+
+      afterUpdate(inst /* , opts */) {
+        /* if status is changed to Running, then find and assign unassigned
+         generators */
+        if (inst.changed('status')) {
+          if (inst.status === collectorStatus.Running) {
+            return u.findAndAssignGenerators(seq);
```
iamigo: promise?

db/model/collector.js
```diff
@@ -124,6 +131,23 @@ module.exports = function collector(seq, dataTypes) {
...
+          } else if (inst.status === collectorStatus.Stopped ||
```
iamigo: the preceding "if" block returns so shouldn't this just be a separate if rather than an "else"?

db/helpers/collectorUtils.js
```diff
@@ -82,8 +82,26 @@ function validateVersion(version) {
...
+ * Find unassigned generators and assign them.
+ * @param {Object} seq - Sequelize object
+ * @returns {Promise} - Resolves to array of assigned generator db objects
+ */
+function findAndAssignGenerators(seq) {
+  return seq.models.Generator.findAll(
+    { where: { currentCollector: null, isActive: true } }
+  )
+  .then((unassignedGenerators) =>
+    Promise.all(unassignedGenerators.map((g) => {
```
jgraff2: can be simplified to .map((g) => {

db/helpers/collectorUtils.js
```diff
@@ -82,8 +82,26 @@ function validateVersion(version) {
   }
 }
 
+/**
+ * Find unassigned generators and assign them.
+ * @param {Object} seq - Sequelize object
+ * @returns {Promise} - Resolves to array of assigned generator db objects
+ */
+function findAndAssignGenerators(seq) {
```
jgraff2: It's not clear from the function name that this only applies to unassigned generators.

db/model/collector.js
```diff
@@ -103,17 +104,23 @@ module.exports = function collector(seq, dataTypes) {
...
-        }
+        return Promise.all([
+          Promise.resolve()
+          .then(() => {
+            // Add createdBy user to Collector writers.
+            if (inst.createdBy) {
+              return new seq.Promise((resolve, reject) =>
+                inst.addWriter(inst.createdBy)
+                .then(() => resolve(inst))
+                .catch((err) => reject(err))
```
jgraff2: can be simplified to inst.addWriter(inst.createdBy)

db/model/generator.js
```diff
@@ -161,16 +161,31 @@ module.exports = function generator(seq, dataTypes) {
...
 
       beforeUpdate(inst /* , opts */) {
         const gtName = inst.generatorTemplate.name;
         const gtVersion = inst.generatorTemplate.version;
+
         if (inst.changed('isActive')) {
           inst.assignToCollector();
         }
 
+        /* If possibleCollectors are changed and this generator current
```
jgraff2: generator's

tests/api/v1/generators/putWithCollector.js
```diff
@@ -189,4 +196,164 @@ describe('tests/api/v1/generators/putWithCollector.js >', () => {
...
+        collectorAlive2 = collectors[1];
+        return done();
+      })
+      .catch(done);
+    });
+
+    afterEach(() => clock.restore());
+
+    it('unassigned generator, put generator with possible collectors should ' +
+      'set currentCollector if alive collector is available', (done) => {
```
jgraff2: Will this same case work on patch, since we're not mocking it there? Can you add a test for patch as well, just for this one case where a possible collector is added to an unassigned generator?

tests/db/model/collector/methods.js
```diff
@@ -195,19 +228,28 @@ describe('tests/db/model/collector/methods.js >', () => {
         Collector.build({ lastHeartbeat })
         .isAlive().should.be.false;
       });
-
     });
 
     describe('reassignGenerators >', () => {
-      beforeEach(() =>
-        GeneratorTemplate.create(generatorTemplate)
+      // check checkMissedHeartbeat test for comments
```
jgraff2: ?

tests/db/model/generator/createWithCollectors.js
```diff
@@ -71,34 +72,38 @@ describe('tests/db/model/generator/createWithCollectors.js >', () => {
...
-      expect(o.generatorTemplate.version).to.equal('1.0.0');
-      expect(typeof o.getWriters).to.equal('function');
-      expect(typeof o.getPossibleCollectors).to.equal('function');
-      expect(o.currentCollector).to.equal(collector1.name);
-      done();
-    })
+    // make collector1 alive
+    collector1.update({ status: 'Running', lastHeartbeat: Date.now() })
+    .then(() => Generator.createWithCollectors(localGenerator)
+      .then((o) => {
```
jgraff2: unnecessary nested "then".

db/model/collector.js
jgraff2: My two cents: I generally prefer using if/else in cases like this (as Pallavi had it originally), even though it's not strictly necessary, because it's more compact and the control flow is clear at a glance.

pallavi2209:
@iamigo I did not make any changes to package-lock.json or package.json intentionally - not sure how it ends up in changed files.

tests/db/model/collector/methods.js
pallavi2209: yeah this lame :P - will change.

pallavi2209:
@jgraff2 Updated the PR with latest changes for PATCH generator - our changes to updateWithCollectors made more tests to fail - I have skipped one of them which needs more discussion. fyi @sameesiddiqui

db/helpers/collectorUtils.js
```diff
@@ -82,8 +82,24 @@ function validateVersion(version) {
   }
 }
 
+/**
+ * Find unassigned generators and assign them.
+ * @param {Object} seq - Sequelize object
+ * @returns {Promise} - Resolves to array of assigned generator db objects
+ */
+function assignUnassignedGenerators(seq) {
```
jgraff2: Is it necessary to pass in the seq object here? Can't we just require it in this file?

db/model/collector.js
```diff
@@ -124,6 +127,25 @@ module.exports = function collector(seq, dataTypes) {
           throw new ValidationError(msg);
         }
       }, // hooks.beforeUpdate
+
+      afterUpdate(inst /* , opts */) {
+        /* if status is changed to Running, then find and assign unassigned
+         generators */
+        if (inst.changed('status')) {
+          if (inst.status === collectorStatus.Running) {
```
jgraff2: It would be better for this comment to be inside the "if" to match the comment below.

db/helpers/collectorUtils.js
pallavi2209: seq is here coming in as a parameter when we initialize model. We can probably get the seq instance from index file, but I noticed that we are usually passing in models for utils, probably because utils are used by corresponding models.
I am fine either ways, let me make the change.

PR #920, opened 2018-07-06
Fix Upsert to use Bot Name

tausifmuzaffar:
When making a bot sometimes you dont know your bot Id

We want to be able to upsert data using the bot name alone


api/v1/controllers/botData.js
```diff
@@ -105,17 +106,46 @@ module.exports = {
...
+
+        BotData.bdExists(queryBody)
+        .then((bd) => {
+          if (bd) {
+            req.swagger.params.key = { value: bd.id };
+            req.swagger.params.queryBody.value.value =
+              bdUtils.combineValue(bd.value, queryBody.value);
+            doPatch(req, res, next, helper);
+          } else {
+            doPost(req, res, next, helper);
```
iamigo: there's no test which gets in here

PR #922, opened 2018-07-06
Change collectors possible collectors

sameesiddiqui:
renamed these fields (so that the names are more reflective of what they represent):
generator.collectors -> generator.possibleCollectors and
collector.currentGenerators -> collector.possibleGenerators


api/v1/swagger.yaml
```diff
@@ -3466,7 +3466,7 @@ paths:
                 minLength: 7
                 maxLength: 4096
                 description: The query to append to GET subjects. Only "absolutePath", "name", "tags" are allowed in subjectQuery.
-              collectors:
+              possibleCollectors:
```
iamigo: Need to make sure there are no references to the old attribute names anywhere in the other refocus-collector-related repos!

db/model/collector.js
```diff
@@ -251,7 +259,7 @@ module.exports = function collector(seq, dataTypes) {
    * @returns {Promise<Array<Generator>>}
    */
   Collector.prototype.reassignGenerators = function () {
-    /* TODO: change to use currentGenerators once that includes current gens only */
+    /* TODO: change to use possibleGenerators once that includes current gens only */
```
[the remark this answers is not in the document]
sameesiddiqui: true. changed these two back.

api/v1/swagger.yaml
sameesiddiqui: is there any way to tie that to this pr? or should I just make a pr in refocus-collector and we merge them at the same time?

PR #928, opened 2018-07-07
Token verify scope

iamigo:
Speed up the verify user token query

### DIFF
--- a/tests/api/v1/userTokens/revokeRestore.js
+++ b/tests/api/v1/userTokens/revokeRestore.js
@@ -20,7 +20,7 @@ const expect = require('chai').expect;
 const registerPath = '/v1/register';
 const tokenPath = '/v1/tokens';
 
-describe('tests/api/v1/userTokens/revokeRestore.js, ' +
+describe.only('tests/api/v1/userTokens/revokeRestore.js, ' +
 `POST ${path}/U/tokens/T/[revoke|restore] >`, () => {
   const predefinedAdminUserToken = tu.createAdminToken();
   const uname = `${tu.namePrefix}test@refocus.com`;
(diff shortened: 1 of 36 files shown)


utils/jwtUtil.js
```diff
@@ -254,13 +218,17 @@ function verifyUserToken(payload, req, cb) {
     // No need to check the token record if this is the default UI token.
     if (p.username === p.tokenname) return cb();
 
-    return Token.findOne({
-      where: {
-        name: p.tokenname,
-        createdBy: req.user.id,
-      },
+    return Token.scope({ method: ['notRevoked', p.tokenname, req.user.id] })
```
pallavi2209: Update or add test with updated error info?

iamigo:
@pallavi2209 added test

tests/api/v1/userTokens/revokeRestore.js
```diff
@@ -112,6 +116,30 @@ describe('tests/api/v1/userTokens/revokeRestore.js, ' +
...
+
+      expect(res.body).to.have.property('name', tname);
+      expect(res.body.isRevoked > '0').to.be.true;
+      api.get(path)
+      .set('Authorization', voldemortToken)
+      .send({})
+      .expect(constants.httpStatus.FORBIDDEN)
+      .end((err2, res2) => {
+        expect(res2.body.errors[0])
+          .to.have.property('description', 'Authentication Failed');
```
pallavi2209: Looks like you are not updating the description when throwing the error - but we have the description here. Is this error coming from somewhere else?

iamigo: Yeah I realized that verifyToken was swallowing that explanation and replacing it with generic "Authentication Failed" anyway, so I just got rid of the custom description there.

PR #930, opened 2018-07-07
When upserting a sample, pre-attach the subject to the sample

iamigo:
(like we already do for aspect) instead of looking it up in db or in separate redis call.


jgraff2:
what about the corresponding change to stop attaching before publish?

jgraff2:
This is a great idea! Seems like it could have a big impact on performance.

PR #934, opened 2018-07-09
publish new real-time event when upserted sample has not changed

iamigo:
Waiting for https://github.com/salesforce/refocus/pull/947 to be merged first


harshkothari410:
Looks good to me.

realtime/redisPublisher.js
```diff
@@ -134,6 +141,39 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
...
+    status: sample.status, // for socket.io perspective filtering
+    absolutePath: sample.absolutePath, // used for persp filtering
+    updatedAt: sample.updatedAt,
+    subject: {
+      absolutePath: sample.absolutePath,
+      tags: sample.subjectTags, // for socket.io perspective filtering
+    },
+    aspect: {
+      name: sample.aspectName, // for socket.io perspective filtering
+      tags: sample.aspectTags, // for socket.io perspective filtering
```
jgraff2: Should we maybe strip out these fields in the emitter or perspective so it's clear to the lens that it shouldn't rely on this object for any data other than updatedAt?

iamigo: I like that idea--let me take a stab at that!!!

The first benefit is as you wrote, but another benefit is smaller event data means less work serializing, and as we saw today, that takes up a lot of our time!

iamigo:
updated the pr per Jon's comment, will test in a heroku environment on monday

iamigo:
@jgraff2 @harshkothari410 @pallavi2209 -- updated after jon's comment, tested in devbox, ready for one more round of review before merging

PR #938, opened 2018-07-10
Fixed default values for room creation

kriscfoster:
- Setting the room type shouldn't try to change active / externalId values
- Default value for externalId should be null instead of ''
- Issue with default externalId being '' is that it doesn't match the validation regex defined in db. Better just default to null as allowNull is true
- Only require params of room name and roomType to create room

### DIFF
--- a/view/rooms/new/FormController.js
+++ b/view/rooms/new/FormController.js
@@ -45,6 +45,7 @@ class FormController extends React.Component {
     this.getTypes();
   }
 
+  // Gets the room types that are available for use
   getTypes(){
     const req = request.get('/v1/roomTypes');
     req
@@ -76,6 +77,7 @@ class FormController extends React.Component {
       });
   }
 
+  // Tries to create a room based on current form.
   createRoom(){
     const req = request.post('/v1/rooms');
     const obj = {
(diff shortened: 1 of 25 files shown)


view/rooms/new/FormController.js
```diff
@@ -45,6 +45,7 @@ class FormController extends React.Component {
     this.getTypes();
   }
 
+  // Gets the room types that are available for use
   getTypes(){
```
cassiodias: Would be possible to change the method name to getAvailableRoomType and then remove the comment? Comments are really easy to be outdated.

view/rooms/new/FormController.js
```diff
@@ -76,6 +77,7 @@ class FormController extends React.Component {
       });
   }
 
+  // Tries to create a room based on current form.
```
cassiodias: Isnt this method self-explanatory? Do we need this comment?

view/rooms/new/app.js
```diff
@@ -47,7 +47,7 @@ function getPathVariables(addr){
 
   const paramType = qdata.roomType || '';
   const paramActive = qdata.active !== 'false';
-  const paramExternalId = qdata.externalId || '';
+  const paramExternalId = qdata.externalId || null;
```
cassiodias: Is there any expectation or new effect (test) from '' to null?

kriscfoster: Thanks @cassiodias for bringing my attention to this. The reason we need to change the default from '' to null is that '' does not pass the regex validation defined for externalId

PR #942, opened 2018-07-11
Passing pubOpts as arg

kriscfoster:
- Passing pubOpts as a separate argument so we don't need to call rtUtils.getNewObjAsString(key, obj) twice, which is expensive


kriscfoster:
@iamigo Thanks Ian, I have tested in devbox and everything looks good for rooms anyway. I was not able to successfully POST any samples to test realtime events in perspectives. How should this be tested?

PR #945, opened 2018-07-11
Adding log activity to jobCleanUp

cassiodias:
* adding activityLog toggle
* adding new activityLog template
* adding activityLog in jobCleanup
* adding test checking that activityLogUtil has been called in jobCleanUp
* some lint warnings removal


tests/clock/jobCleanup.js
```diff
@@ -27,6 +27,7 @@ jobQueue.completeAsync = Promise.promisify(jobQueue.complete);
 describe('tests/clock/jobCleanup.js >', () => {
   before((done) => {
     tu.toggleOverride('enableWorkerProcess', true);
+    tu.toggleOverride('enableWorkerActivityLogs', true);
```
iamigo: should be adding new toggle for this -- enableJobCleanupActivityLogs

clock/scheduledJobs/jobCleanup.js
```diff
@@ -43,6 +46,22 @@ function execute(batchSize, delay) {
     if (featureToggles.isFeatureEnabled('instrumentKue')) {
       console.log(`[KJI] Removed ${removedJobCount} jobs`);
     }
+
+    if (featureToggles.isFeatureEnabled('enableWorkerActivityLogs')) {
```
iamigo: enableJobCleanupActivityLogs ?

clock/scheduledJobs/jobCleanup.js
```diff
@@ -43,6 +46,22 @@ function execute(batchSize, delay) {
...
+      const jobEndTime = Date.now();
+      const jobResultData = {};
+      const iterations = skippedJobCount + removedJobCount - errorJobCount;
+
+      jobResultData.activity = 'jobCleanup';
```
iamigo: add the "template" for the new activity log type in /config/activityLog.js

clock/scheduledJobs/jobCleanup.js
```diff
@@ -53,46 +69,45 @@ function execute(batchSize, delay) {
...
+        console.log('Error getting completed jobs from queue', err);
+        return Promise.reject(err);
+      })
+
+      // delete n jobs
+      .then((jobs) =>
+        Promise.all(jobs.map((job) => {
+          if (delay > 0 && now - job.updated_at < delay) {
+            return 'skipped';
+          }
```
cassiodias: Just removed lint warnings when
``
else {
   return ...
``

config/toggles.js
```diff
@@ -96,6 +96,8 @@ const longTermToggles = {
     'unauthorized'),
   enableWorkerActivityLogs: envVarIncludes(pe, 'ENABLE_ACTIVITY_LOGS',
     'worker'),
+  enableJobCleanupActivityLogs: envVarIncludes(pe, 'ENABLE_ACTIVITY_LOGS',
+    'jobCleanUp'),
```
iamigo: should be 'jobCleanup' (since "cleanup" is a real unhyphenated word, the "u" does not get camel-cased)

worker/jobs/jobCleanupJob.js
```diff
@@ -21,6 +21,11 @@ module.exports = (job, done) => {
   const reqStartTime = job.data.reqStartTime;
   const dbStartTime = Date.now();
 
+  if (featureToggles.isFeatureEnabled('instrumentKue')) {
```
iamigo: why adding this here?

clock/scheduledJobs/jobCleanup.js
```diff
@@ -53,46 +69,45 @@ function execute(batchSize, delay) {
 
     // get n jobs
     return kue.Job.rangeByStateAsync('complete', from, to, 'asc')
-    .catch((err) => {
-      console.log('Error getting completed jobs from queue', err);
-      return Promise.reject(err);
-    })
+      .catch((err) => {
+        console.log('Error getting completed jobs from queue', err);
+        return Promise.reject(err);
```
iamigo: Where is this rejection being caught and handled?
(Hint: if we're not handling it somewhere, it probably shouldn't be a rejection!)

clock/scheduledJobs/jobCleanup.js
```diff
@@ -53,46 +69,45 @@ function execute(batchSize, delay) {
...
-          .catch((err) => {
-            console.log(`Error removing job ${job.id}`, err);
-            return Promise.reject(err);
-          });
-        }
-      }))
-    )
+            .catch((err) => {
+              errorJobCount++;
+              return Promise.reject(err);
```
iamigo: Where is this rejection being caught and handled?
(Hint: if we're not handling it somewhere, it probably shouldn't be a rejection!)

clock/scheduledJobs/jobCleanup.js
```diff
@@ -53,46 +69,45 @@ function execute(batchSize, delay) {
...
+            });
+        }))
+      )
+
+      // count skipped jobs
+      .then((results) => {
+        const skippedCount = results.reduce((count, result) =>
+            result === 'skipped' ? count + 1 : count,
```
iamigo: `count + 1` should be `count++` ?

clock/scheduledJobs/jobCleanup.js
```diff
@@ -53,46 +69,45 @@ function execute(batchSize, delay) {
...
-        result === 'skipped' ? count + 1 : count,
-      0);
-      skippedJobCount += skippedCount;
-      removedJobCount += results.length - skippedCount;
-      return Promise.resolve(results);
-    })
+      // get and delete next n jobs
+      .then((results) => {
+        if (results.length < n) {
+          return Promise.resolve();
```
iamigo: this whole "if" can be on one line
```if (results.length < n) return Promise.resolve();```

File: worker/jobs/jobCleanupJob.js (hunk 6 above)
Comment (cassiodias) [reply worker/jobs/jobCleanupJob.js]: This was inside the jobCleanup, as the job subject delete has the same, I thought that would be the place to be moving from jobCleanup to JobCleanupJob. 

Should I get rid off?

File: clock/scheduledJobs/jobCleanup.js (hunk 7 above)
Comment (cassiodias) [reply clock/scheduledJobs/jobCleanup.js]: Ok, didn't touch that part - the change is just formatting.

This is happening if something goes wrong getting the JOBs from kue.Job.rangeByStateAsync

Should I change reject to resolve?

File: clock/scheduledJobs/jobCleanup.js (hunk 8 above)
Comment (cassiodias) [reply clock/scheduledJobs/jobCleanup.js]: Same here... just added the ++... Should I change to resolve?

File: clock/scheduledJobs/jobCleanup.js (hunk 9 above)
Comment (cassiodias) [reply clock/scheduledJobs/jobCleanup.js]: I agree, didn't change because it was already here and the tests are showing the right value. Will change that anyway.

Comment (cassiodias) [reply clock/scheduledJobs/jobCleanup.js]: Actually should be ++count or count+1, otherwise reduce will generate: 1.

``
var numbers = [1,1,1,1,1];

const quantityInArray = numbers.reduce((count, result) => ++count);
console.log(quantityInArray); // 4

const quantityInArray = numbers.reduce((count, result) => count++);
console.log(quantityInArray); // 1

const quantityInArray = numbers.reduce((count, result) => count+1);
console.log(quantityInArray); // 4
``

File: clock/scheduledJobs/jobCleanup.js (hunk 7 above)
Comment (iamigo) [reply clock/scheduledJobs/jobCleanup.js]: nah, don't change anything now since you didn't just write this now

we already have a user story rid of all our UnhandledPromiseRejectionWarning from tests and in prod so we'll just deal with it as part of that

File: clock/scheduledJobs/jobCleanup.js (hunk 8 above)
Comment (iamigo) [reply clock/scheduledJobs/jobCleanup.js]: nah, don't change anything now since you didn't just write this now

we already have a user story rid of all our UnhandledPromiseRejectionWarning from tests and in prod so we'll just deal with it as part of that

File: config/toggles.js (hunk 5 above)
Comment (cassiodias) [reply config/toggles.js]: Done

File: tests/clock/jobCleanup.js [hunk 11]
```diff
@@ -99,7 +110,12 @@ describe('tests/clock/jobCleanup.js >', () => {
       expectNJobs(jobCount)
       .then(() => jobCleanup.execute(batchSize, delay))
       .then(() => expectNJobs(expectedCount))
-      .then(done).catch(done);
+      .then(() => {
+        expect(activityLogUtil.printActivityLogString.calledOnce)
```
sameesiddiqui: Nice! Clever way to check for the logging. I was just using winston before but this seems cleaner

sameesiddiqui: Downside is that it doesn't check content of the logs. Can you add a test to make sure the correct fields are being logged?

iamigo:
I like Samee's last comment - @cassiodias?

tests/clock/jobCleanup.js
cassiodias: Sure! Good point.

Will added that. 👊🏻

cassiodias:
@iamigo liked that too - will check the content.

cassiodias:
I am not sure how we handle pre-prod tasks but we have to append **jobCleanup** to **ENABLE_ACTIVITY_LOGS**

PR #948, opened 2018-07-12
Job cleanup activity log

cassiodias:
* adding activityLog toggle
* adding new activityLog template
* adding activityLog in jobCleanup
* adding test checking that activityLogUtil has been called in jobCleanUp
* some lint warnings removal


cassiodias:
just checking flaky tests

PR #951, opened 2018-07-17
If a room name changes, the subtitle of all open rooms should change too

kriscfoster:


db/model/room.js
```diff
@@ -123,7 +123,7 @@ module.exports = function room(seq, dataTypes) {
       },
 
       afterUpdate(instance /* , opts */) {
-        if (instance.changed('settings')) {
+        if (instance.changed('settings') || instance.changed('name')) {
```
iamigo: if the name changed, then the namespace for the old room is obsolete, and the responsible thing to do is to delete the socket.io namespace from that old room

granted it's not a big deal for us in prod since all the dynos are restarted at least once a day, but it is nevertheless good to clean up after ourselves

iamigo: @kriscfoster any comment on my comment? i'll still approve this as is if you say you want to punt on that for now, but at the very least there should be a user story to resolve this

kriscfoster:
@iamigo yes I agree, I'll make a story for it

PR #953, opened 2018-07-17
minor perspective/room namespace initialization optimization

iamigo:
Create sequelize scopes for loading perspectives and rooms for namespace initialization.
Execute room & persp namespace initialization in parallel.

### DIFF
--- a/db/model/room.js
+++ b/db/model/room.js
@@ -184,6 +184,14 @@ module.exports = function room(seq, dataTypes) {
         'id',
         'name',
       ],
+      exclude: [
+        'externalId',
+        'settings',
+        'active',
+        'bots',
+        'createdAt',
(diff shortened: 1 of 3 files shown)


db/model/room.js
```diff
@@ -176,6 +176,13 @@ module.exports = function room(seq, dataTypes) {
       through: 'RoomWriters',
       foreignKey: 'roomId',
     });
+    Room.addScope('namespace', {
+      attributes: ['id', 'name'],
+      where: {
+        active: true,
+        isDeleted: 0,
```
iamigo: Oops--none of us caught that the Room model does not have an "isDeleted" attribute!!! Fixing in https://github.com/salesforce/refocus/pull/960

PR #955, opened 2018-07-18
Job Activity Log Fix - changing the activity counter

cassiodias:


clock/scheduledJobs/jobCleanup.js
```diff
@@ -21,23 +21,29 @@ kue.Job.rangeByStateAsync = Promise.promisify(kue.Job.rangeByState);
 kue.Job.prototype.removeAsync = Promise.promisify(kue.Job.prototype.remove);
 const activityLogUtil = require('../../utils/activityLog');
 
-function logActivity(skippedJobCount, removedJobCount, errorJobCount,
-                     jobStartTime) {
-  const iterations = skippedJobCount + removedJobCount - errorJobCount;
-  if (iterations) {
-    const jobEndTime = Date.now();
+function hasValidCount(log) {
```
iamigo: Why wouldn't it be legit to have a log line with 0 jobs removed, 0 jobs skipped, 0 jobs errors... ? e.g. If there were no jobs in the previous period, those would all be zeros. The only value which should never be zero is "iterations" because it should always go into "deleteNextNJobs" at least once

PR #958, opened 2018-07-18
Apply filter optimization

iamigo:


realtime/utils.js
```diff
@@ -29,6 +29,7 @@ const filters = [
   'statusFilter',
 ];
 const botAbsolutePath = '/Bots';
+const SEVEN = 7; // the length of the strings "INCLUDE" and "EXCLUDE"
```
cassiodias: Never used

realtime/utils.js
```diff
@@ -119,84 +120,63 @@ function isPresent(filterValueSet, objValueArr) {
...
-    }
+  /*
+   * Short-circuit return true if the filterString is not a name-value pair
+   * with an "=" separator.
+   */
+  if (nvp.length < 2) return true;
 
-    // filter type is either INCLUDE or EXCLUDE
-    const filterType = filterComponents[0];
+  const filterType = nvp[0]; // INCLUDE or EXCLUDE
```
cassiodias: I'd remove the comment since the line 153 explains the return when there is no match with INCLUDE/EXCLUDE.

iamigo:
update per @cassiodias comments

PR #961, opened 2018-07-18
update tag change logic so that patching the same tags does not trigg…

jgraff2:
…er sample delete events, even if out of order or different case.


db/helpers/common.js
```diff
@@ -172,6 +172,27 @@ function validateContextDef(contextDef, requiredProps) {
...
 
+/**
+ * Determine whether the tags have changed by doing a deep comparison
+ * @param  {Object} inst - db instance
+ * @returns {Boolean}
+ */
+function tagsChanged(inst) {
+  let prevTags = inst.previous('tags');
+  let currTags = inst.get('tags');
+  console.log('prevTags', prevTags);
```
iamigo: remove before merging

jgraff2: oops.
removed.

jgraff2:
fixed

PR #962, opened 2018-07-19
add tests for the realtime utils applyFilter fn

iamigo:


iamigo:
Thx for the great comments, @cassiodias! I have added a test, but the other changes are actually in a separate PR with some minor optimizations to the applyFilter fn. However, I wanted to add these tests first, in their own PR, so that the changes to applyFilter are proven to *not* cause any applyFilter regressions.

PR #973, opened 2018-08-06
Collector integration fixes pt. 1

jgraff2:
(1) fix heartbeatLatencyTolerance behavior
(2) allow heartbeats from missed collectors. update status to running.
(3) on collector start, reset tracked changes
(4) cleanAndStripNulls from heartbeat response

### DIFF
--- a/api/v1/controllers/collectors.js
+++ b/api/v1/controllers/collectors.js
@@ -25,6 +25,7 @@ const doPatch = require('../helpers/verbs/doPatch');
 const u = require('../helpers/verbs/utils');
 const heartbeatUtils = require('../helpers/verbs/heartbeatUtils');
 const httpStatus = require('../constants').httpStatus;
+const status = require('../../../db/constants').collectorStatuses;
 const decryptSGContextValues = require('../../../utils/cryptUtils')
   .decryptSGContextValues;
 const encrypt = require('../../../utils/cryptUtils').encrypt;
@@ -270,10 +271,14 @@ function heartbeat(req, res, next) {
       throw new apiErrors.ForbiddenError({
         explanation: 'Authentication Failed',
       });
-    } else if (o.status !== 'Running' && o.status !== 'Paused' && o.status !== 'MissedHeartbeat') {
+    } else if (o.status === status.Stopped) {
       throw new apiErrors.ForbiddenError({
-        explanation: `Collector must be running or paused or missed. Status: ${o.status}`,
+        explanation: `Collector cannot send a heartbeat if stopped. Status: ${o.status}`,
       });
+    }
+
+    if (o.status !== status.Running) {
+      o.set('status', status.Running);
     }
 
     retval.collectorConfig.status = o.status;
@@ -355,7 +360,7 @@ function heartbeat(req, res, next) {
   .then((updated) => retval.generatorsUpdated = updated)
   .then(() => {
     u.logAPI(req, resultObj, retval);
-    res.status(httpStatus.OK).json(retval);
+    res.status(httpStatus.OK).json(u.cleanAndStripNulls(retval));
   })
   .catch((err) => u.handleError(next, err, helper.modelName));
 } // heartbeat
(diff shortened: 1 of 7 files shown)


db/model/collector.js
```diff
@@ -257,10 +256,11 @@ module.exports = function collector(seq, dataTypes) {
   Collector.prototype.isAlive = function () {
     if (!this.lastHeartbeat) return false;
     const tolerance = collectorConfig.heartbeatLatencyToleranceMillis;
+    const interval = collectorConfig.heartbeatIntervalMillis;
     const now = Date.now();
     const lastHeartbeat = this.lastHeartbeat.getTime();
     const elapsed = now - lastHeartbeat;
-    return elapsed < tolerance;
+    return elapsed < interval + tolerance;
```
iamigo: I always find it easier/faster to read a line like this with totally unnecessary parentheses added, e.g.

```
return elapsed < (interval + tolerance);
```

api/v1/controllers/collectors.js
```diff
@@ -270,12 +271,16 @@ function heartbeat(req, res, next) {
       throw new apiErrors.ForbiddenError({
         explanation: 'Authentication Failed',
       });
-    } else if (o.status !== 'Running' && o.status !== 'Paused' && o.status !== 'MissedHeartbeat') {
+    } else if (o.status === status.Stopped) {
```
jgraff2: Do we really want to accept heartbeats from paused collectors? What's the use case here?

db/model/collector.js
```diff
@@ -130,17 +131,16 @@ module.exports = function collector(seq, dataTypes) {
...
+             generators */
             return u.assignUnassignedGenerators();
-          }
-
-          if (inst.status === collectorStatus.Stopped ||
-            inst.status === collectorStatus.Paused) {
-            /* if status is changed to Stopped or Paused, then reassign the
-             generators which were assigned to this collector */
-            return inst.reassignGenerators();
+          } else if (inst.previous('status') === collectorStatus.Running) {
```
jgraff2: Are we sure we want to unassign generators when a collector is paused? What's the purpose of "Paused" if it's functionally identical to "Stopped"?

api/v1/controllers/collectors.js
```diff
@@ -270,12 +271,16 @@ function heartbeat(req, res, next) {
...
-        explanation: `Collector must be running or paused or missed. Status: ${o.status}`,
+        explanation: `Collector cannot send a heartbeat if stopped. Status: ${o.status}`,
       });
     }
 
+    if (o.status !== status.Running) {
+      o.set('status', status.Running);
+    }
+
     retval.collectorConfig.status = o.status;
```
jgraff2: What's the purpose of sending back the status in the heartbeat? As it is right now it will always be "Running". Was it supposed to be that a collector can be marked as "Stopped" via the api and will be instructed to stop next time it sends a heartbeat?

jgraff2:
@iamigo since you already approved, please be sure to take a look at the comments I just left.

api/v1/controllers/collectors.js
iamigo: Correct. Gives a collector admin the ability to stop or resume a running or paused collector without needing access to the host where it's running.

db/model/collector.js
iamigo: Paused == still sends heartbeats but doesn't generate samples (i.e. it can be resumed via api without needing access to the host running the collector)

api/v1/controllers/collectors.js
jgraff2: But if you stop a running collector via the api, it's status in the db will be set to "Stopped", and the next heartbeat it sends will be rejected, so the collector will never find out it was supposed to be stop.

@iamigo

iamigo: Oops -- loophole! Need to figure out how to allow a user to use API to stop a collector remotely. Pls. create story.

jgraff2: will do.

db/model/collector.js
pallavi2209: Is there a test for this use case = i.e. reassigning generator when previous status='Running'?

jgraff2: This is logically equivalent to what it was before, I just simplified it. The tests you wrote in collectors/pause and collectors/stop cover this.

PR #975, opened 2018-08-07
UI Test Setup

kriscfoster:
This stuff will probably need some tweaking around so this PR is just a conversation starter

- These tests are full end to end integration tests, from DB to UI

- The tests run headless (without a browser) in travis but can be run with a browser locally to view the tests happening realtime

- Added a utils file that includes a method which logs in, most ui tests will probably start off by using this function, as they can then access any other Refocus page

- Added tests for login page to render and redirect correctly

- Added test for rooms list page to render

- Added test for room page to render with 1 bot

(Before any test we can insert whatever we want into the db, and can then use this data to test)


tests/ui/pages/login.js
```diff
@@ -0,0 +1,81 @@
...
+    await page.waitFor('h1');
+    expect(await page.$eval('h1',
+      heading => heading.innerText)).to.eql('Refocus');
+    await page.waitForSelector('input[name=username]');
+    expect(await page.$eval('[name=username]',
+      input => input.placeholder)).to.eql('Enter username');
+    await page.click('input[name=username]');
+    await page.type('input[name=username]', 'potatoes');
+    expect(await page.$eval('[name=username]',
+      input => input.value)).to.eql('potatoes');
```
cassiodias: What is the result of this login? Failed? Would be nice to check whether we are showing failure once user/pass are wrong.

tests/ui/setup.js
```diff
@@ -0,0 +1,32 @@
+/**
```
cassiodias: Is it running properly on Travis? Like headless, phanton process

kriscfoster: @cassiodias yeah it runs properly on travis

kriscfoster:
@kamilsmuga gave me some useful info on testing design patterns so I'm going to add that stuff in too (maintaining a page object file for each page), this object will store the selectors for elements so they can be used in tests

PR #978, opened 2018-08-13
BugFix: Attach aspect subject when subject is invalid

cassiodias:
* added check if a sub is invalid and throw an Error message
* added test sending and invalid abs path which will result in an invalid subject (not found)

### DIFF
--- a/realtime/redisPublisher.js
+++ b/realtime/redisPublisher.js
@@ -126,7 +126,7 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
 
   // No need to attachAspectSubject if subject and aspect are already attached
   if (sampleInst.hasOwnProperty('subject') &&
-  sampleInst.hasOwnProperty('aspect')) {
+    sampleInst.hasOwnProperty('aspect')) {
     prom = Promise.resolve(sampleInst);
   } else {
     prom = rtUtils.attachAspectSubject(sampleInst, subjectModel, aspectModel);
@@ -137,7 +137,7 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
       sample.absolutePath = sample.subject.absolutePath; // reqd for filtering
       publishObject(sample, eventType);
       return sample;
-    }
+    } // What happend here if is invalid??
   });
 } // publishSample
 
--- a/realtime/redisPublisher.js
+++ b/realtime/redisPublisher.js
@@ -137,7 +137,7 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
       sample.absolutePath = sample.subject.absolutePath; // reqd for filtering
       publishObject(sample, eventType);
       return sample;
-    } // What happend here if is invalid??
+    }
   });
 } // publishSample
 
--- a/realtime/utils.js
+++ b/realtime/utils.js
@@ -375,7 +375,7 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
     return Promise.resolve(null);
   }
 
-  let nameParts = sample.name.split('|');
+  const nameParts = sample.name.split('|');
   const subAbsPath = nameParts[0];
   const aspName = nameParts[1];
   const subOpts = {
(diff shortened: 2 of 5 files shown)


cassiodias:
@iamigo "This is an error and should be handled as an error, i.e. if bulk, add to array of errors, or if non-bulk, throw" Do we use this in any bulk operation? Could not find it.

Added a check and a throw Error.

realtime/utils.js
```diff
@@ -408,16 +408,30 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
...
+    const SUBJECT_INDEX = 1;
+    let asp = response[ASPECT_INDEX];
+    let sub = response[SUBJECT_INDEX];
+
+    if (!sub) {
+      /*
+       throw error when trying to attach the subject to a sample and
+       no subject is found for the absolute path in the sample name.
+       */
+      const message = `- attachAspectSubject - Failed to find subject by
```
iamigo: don't want the carriage return and extra spaces in the message string

iamigo:
@cassiodias  /worker/jobs/bulkUpsertSamplesJob.js calls publishSample from realtime/redisPublisher, which *can* call attachAspectSubject

realtime/redisPublisher.js
```diff
@@ -137,7 +137,7 @@ function publishSample(sampleInst, subjectModel, event, aspectModel) {
       sample.absolutePath = sample.subject.absolutePath; // reqd for filtering
       publishObject(sample, eventType);
       return sample;
-    }
+    } // What happend here if is invalid??
```
cassiodias: @iamigo just pushed this code for double check... redisPublisher call attachAspect that now can throw an exception which is ok here... the .catch might be handled from BulkUpsert... anyway, IF for some reason this sample is null/undefined?

worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -54,22 +54,31 @@ module.exports = (job, done) => {
...
 
-          /*
-           * Resolve the promise right away, *before* we actually publish
-           * the sample. Under heavy load, publish a sample can get backed up
-           * because we are looking up the subject info and including that in
-           * the payload of the real-time event.
-           */
-          publisher.publishSample(result, subHelper.model);
-          return Promise.resolve();
+        if (featureToggles.isFeatureEnabled('publishSampleInPromiseChain')) {
```
cassiodias: here is the change as you asked yesterday. sucessCount ++ if data is saved or error count otherwise.

worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -54,22 +54,31 @@ module.exports = (job, done) => {
...
+            .then((sample) => {
+              successCount++;
+              return Promise.resolve(sample);
+            }).catch((err) => {
+              errors.push({ isFailed: true,
+                explanation: err.message, });
+              return Promise.resolve();
+            });
         }
+
```
cassiodias: @iamigo Problem happens here, where we don't expect the promise. We can check and count success. Should we change the rule for this job, remove the toggle and always wait for the publish sample promise? Seems that we have a different use case here if we always have to verify success/failure for publishSample.
Make sense?

iamigo: We do not need to consider the success or failure of the publish operation at all when counting success/failure of the sample upsert. The publish is irrelevant.

Do not remove the toggle.

worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -54,22 +54,28 @@ module.exports = (job, done) => {
...
-           * because we are looking up the subject info and including that in
-           * the payload of the real-time event.
-           */
-          publisher.publishSample(result, subHelper.model);
-          return Promise.resolve();
+        successCount++;
+        if (featureToggles.isFeatureEnabled('publishSampleInPromiseChain')) {
+          // Wait for publish to complete before resolving the promise.
+          return publisher.publishSample(result, subHelper.model)
+            .catch((err) => {
```
iamigo: Since both code paths calling publisher.publishSample are just catching and logging, maybe move the catch and resolve to the promise which publishSample is returning instead?

cassiodias:
Created new https://github.com/salesforce/refocus/pull/982

PR #981, opened 2018-08-14
Filter events by event type

kriscfoster:
Bots should be able to filter what type of events they want to get instead of getting all events

- Events have a field called "context"

    - this context object has a field called type (this type may be something like 'Comment', 'User' (users joining leaving rooms))

- added type param in swagger (GET v1/events?type={type})
- in controller, extracted this 'type' param and placed it inside context.type
- added test for getting events by type


tausifmuzaffar:
@kriscfoster  is possible to make a simple filter that will filter based on any JSON for context?

kriscfoster:
@tausifmuzaffar Possibly but I'm not sure how because we want to have the params defined in swagger. Is there any other field of context I should add too while we're doing this?

tausifmuzaffar:
@kriscfoster Im not sure what field it is, but it would be good to use this to find TCE

kriscfoster:
@tausifmuzaffar TCE is just stored in the logline field, which is already at the top level and not inside context

tausifmuzaffar:
@kriscfoster that sounds fine, well make another user story for that

tests/api/v1/events/get.js
```diff
@@ -33,8 +33,10 @@ describe('tests/api/v1/events/get.js >', () => {
   let testEvent = u.getStandard();
   let testEventOutput;
   let testEvent2 = u.getStandard();
+  testEvent2.context.type = 'Comment';
```
iamigo: Is the type value case-sensitive? (I think it should be case-INsensitive, unless you have a compelling reason not to do so. Add tests for case-insensitivity.)

kriscfoster:
@iamigo figured out how to make this work with case insensitivity but think this solution could be useful in other cases too

PR #982, opened 2018-08-16
BugFix: Attach invalid subject

cassiodias:
Handling _UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'get' of null_

- Added check if a subject is invalid and throws an Error message;
- Added test to check the exception in *realTimeUtils.attachAspectSubject* sending an invalid abs path which will result in an invalid subject (not found);
- Added catch in publisher to avoid error to be exposed to the upper level (BulkUpsert);
- Added more test in test-publisher checking the flow.


worker/jobs/bulkUpsertSamplesJob.js
```diff
@@ -54,22 +54,22 @@ module.exports = (job, done) => {
         if (result.isFailed) {
           errors.push(result.explanation);
           return Promise.resolve();
-        } else {
```
cassiodias: Removed the unnecessary else as lint's warning.

realtime/utils.js
```diff
@@ -408,16 +408,29 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
     sample.subject ? sample.subject : subjectModel.findOne(subOpts),
   ])
   .then((response) => {
-    let asp = response[0];
-    let sub = response[1];
-    asp = asp.get ? asp.get() : asp;
+    const ASPECT_INDEX = 0;
```
iamigo: it's a tiny thing, but you can move the definition of these two constants up to the top of the module

realtime/utils.js
```diff
@@ -408,16 +408,29 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
...
+    const SUBJECT_INDEX = 1;
+    let asp = response[ASPECT_INDEX];
+    let sub = response[SUBJECT_INDEX];
+
+    if (!sub) {
+      /*
+       throw error when trying to attach the subject to a sample and
```
iamigo: this comment seems redundant since the error message is saying the same thing

realtime/utils.js
```diff
@@ -408,16 +408,29 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
...
+       no subject is found for the absolute path in the sample name.
+       */
+      const message = `Subject not found by Sample abs path ${subAbsPath}`;
+      throw new Error(message);
```
iamigo: So what happens now when processing a batch of samples and we can't find the subject for one of the samples in the batch?

realtime/utils.js
cassiodias: Awesome! DOne

realtime/utils.js
cassiodias: Well, it will just log the error and move on. The bulk upsert will not be affected because the publisher has failed.

In the story description we have the count success or failure, but as we chat yesterday, the publishing process shouldn't affect the bulk.

realtime/utils.js
cassiodias: done

tests/realtime/redisPublisher.js
```diff
@@ -206,23 +208,39 @@ describe('tests/realtime/redisPublisher.js >', () => {
...
-
+          expect(pubObj.subject.absolutePath).to.be.equal(pubObj.absolutePath);
           done();
         })
         .catch(done);
       });
     });
 
+    it('Must be able to handle when Error from real time attachAspectSubject',
+    (done) => {
```
jgraff2: Is this supposed to be testing the catch on line 144? If so, you would expect the promise to be resolved here, right?

cassiodias: That's true @jgraff2. We were catching the exception at the upper level (bulk script) then I forgot to change the test once we move the catch to redisPublisher.

Thanks!

PR #985, opened 2018-08-17
Add iFrame security to prevent bots from affecting parent

tausifmuzaffar:


kriscfoster:
@tausifmuzaffar this seems to affect the bot heights, not sure if we can do anything about this
screen shot 2018-08-17 at 10 56 50

tausifmuzaffar:
@kriscfoster im getting this bug using master, please confirm

view/rooms/app.js
```diff
@@ -536,6 +538,8 @@ function displayBot(bot, botIndex) {
   // We don't want to scroll the iframe, just the bot inside it
   iframe.scrolling = 'no';
   iframe.frameBorder = 0;
+  iframe.sandbox = `allow-scripts allow-same-origin
```
iamigo: Cool! If there's no test for bot access parent, please either add one or create user story to add it.

kriscfoster:
@tausifmuzaffar I don't think this was happening me with master but I'm not 100% sure. Not sure how this would have been introduced?

view/rooms/app.js
kriscfoster: @tausifmuzaffar I think strings with backticks should be something like this:

iframe.sandbox = `allow-scripts allow-same-origin\
    allow-pointer-lock allow-forms`;

to prevent a newline being inserted into the string:
ref: https://stackoverflow.com/questions/37321047/wrap-long-template-literal-line-to-multiline-without-creating-a-new-line-in-the

tausifmuzaffar:
@kriscfoster try it now

PR #986, opened 2018-08-17
For Room Autocreate, GET instead of POSTing and expecting error

kriscfoster:
PR includes:
- Created a newRoom utils file that has functions to check if a room exists with the name and a function to create a new room with given parameters

- Utilise these functions from rooms/new page

- Added tests to test these functions

The old workflow for this was:
- POST v1/rooms with new room
- IF the POST response was an error because of a room with the name already existing then it would reroute to that room ELSE create the room and reroute to it

The new workflow is:
- GET v1/rooms with the new room name
- IF the room exists then reroute to it ELSE POST the new room and reroute


view/rooms/utils/newRoom.js
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Creates a room from parameters
+   *
+   * @param {Object} paramaters - Parameters to use when creating the room
+   * @returns {Promise} - Resolves when room is created.
+   */
+  createRoomFromParameters(paramaters) {
+    const postUrl = '/v1/rooms';
+    const room = {
```
cassiodias: Is there any validation when the parameter has no attributes as expected?

kriscfoster: @cassiodias there is no validation on the frontend yet but the backend will not create the room if it does not have a name or a type. Also, this function will only be called if it has these fields (ref: https://github.com/salesforce/refocus/pull/986/files#diff-8ea0877b79ba6bad7603e84e1bf270a3R91)

PR #990, opened 2018-08-23
log out values of env vars on startup

iamigo:
Define new feature toggle "enableEnvActivityLogs".
Define new log activity type "env".
When enabled, on server start, one "activity=env" log line is printed for each environment variable. Some env variables are skipped altogether, others are not skipped but are "hidden", where we print out the log line for the env var but we mask the value (i.e. to indicate that the variable has been set, but the value is secret).
When running multiple node processes in a single dyno, each of the node processes has the same env, so we only print out the env vars once.

### DIFF
--- a/utils/logEnvVars.js
+++ b/utils/logEnvVars.js
@@ -11,31 +11,68 @@
  */
 'use strict'; // eslint-disable-line strict
 const featureToggles = require('feature-toggles');
+const config = require('../config');
 const activityLog = require('./activityLog');
-const HIDE_ME_LIST = [
-  'DATABASE_BASTION_KEY',
-  'DATABASE_URL',
-  'REDIS_URL',
-  'SECRET_TOKEN',
-  'SESSION_SECRET',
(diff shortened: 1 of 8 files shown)


utils/logEnvVars.js
```diff
@@ -0,0 +1,51 @@
...
+  .map((name) => ({
+    name,
+    value: HIDE_ME_LIST.includes(name) ? '"hidden"' : `"${env[name]}"`,
+  }));
+} // prepareObjectsToLog
+
+function log(env) {
+  if (featureToggles.isFeatureEnabled('enableEnvActivityLogs')) {
+    prepareObjectsToLog(env)
+    .forEach((obj) => activityLog.printActivityLogString(obj, 'env'));
```
cassiodias: Would it be valid to add a test to check if printActivityLog has been called and the data? 
We can validate if we are communicating properly to a specific dependency.

iamigo:
not yet, @pallavi2209  - i have some changes coming

PR #992, opened 2018-08-27
Only display socket disconnect message when I/O disconnects

tausifmuzaffar:


kriscfoster:
@tausifmuzaffar I actually meant if reason !=== 'io server disconnect' should we log out what the reason actually is? Not sure if this is important or not

tausifmuzaffar:
@kriscfoster we already out put the log here

PR #1000, opened 2018-09-18
Cannot read property 'get' of null from Aspect

cassiodias:
Adding check and error if the aspect is null to help on prodops investigations


realtime/utils.js
```diff
@@ -419,6 +419,11 @@ function attachAspectSubject(sample, subjectModel, aspectModel) {
       throw new Error(message);
     }
 
+    if (!asp) {
+      const message = `Aspect not found by Aspect name ${aspName}`;
```
iamigo:
```
const message = `Aspect ${aspName} not found`;
```

cassiodias: done

PR #1004, opened 2018-09-26
Clock/worker changes

jgraff2:
Organized by commit:

core changes: (16 files)
- maintain clock job interval independent of dyno restarts
- consolidate worker and clock config
- test clock
- test worker

other changes: (40 files)
- clock job format changes
- change job names
- existing test changes
- misc

Note: This includes changes to the config variables for the clock and worker jobs.
We need to make sure they are updated before this is released, and the old ones removed after.


clock/scheduledJobs/persistSampleStore.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Copyright (c) 2017, salesforce.com, inc.
```
cassiodias: 2018

worker/index.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright (c) 2016, salesforce.com, inc.
```
cassiodias: 18

PR #1007, opened 2018-10-03
Revert "Keep params bug"

kriscfoster:
- Caused regression bug with "Open IMC" button from GUS

Reverts salesforce/refocus#1006


kriscfoster:
@iamigo I have tested this and can verify it fixes the issue we had last night

PR #1010, opened 2018-10-09
Multiple associations returning extra FK to the client

cassiodias:
* Adding attributes when multiple associations and returning extra attributes (FK) to the client.


api/v1/helpers/verbs/utils.js
```diff
@@ -237,20 +237,53 @@ function handleAssociations(reqObj, inst, props, method) {
   });
 } // handleAssociations
 
+function isMultipleAssociations(opts) {
+  const MULTIPLE_ASSOCIATIONS = 3;
+  return opts.attributes.length >= MULTIPLE_ASSOCIATIONS;
```
jgraff2: This will also return true when multiple non-association fields are specified.
I don't think this will cause problems, it just means we're doing extra work when not necessary. Correct?

package-lock.json
```diff
@@ -1025,12 +1025,12 @@
         "babylon": {
           "version": "6.18.0",
           "resolved": "https://registry.npmjs.org/babylon/-/babylon-6.18.0.tgz",
-          "integrity": "sha1-ry87iPpvXB5MY00aD46sT1WzleM="
+          "integrity": "sha512-q/UEjfGJ2Cm3oKV71DJz9d25TPnq5rhBVL2Q4fA5wcC3jcrdn7+SssEybFIxwAvvP+YCsCYNKughoF33GxgycQ=="
         },
```
jgraff2: package-lock not needed.

tests/api/v1/common/testAssociations.js
```diff
@@ -105,7 +105,7 @@ function testAssociations(path, associations, joiSchema, conf) {
       .expect(constants.httpStatus.OK)
       .expect((res) => {
         expect(res.body).to.be.an('object');
-        expect(res.body).to.have.keys('id', 'name', assoc, 'apiLinks');
+        expect(res.body).to.contains.all.keys('id', 'name', assoc, 'apiLinks');
```
jgraff2: Reason for using "all" here but not above?

tests/enableCache/perspectives.js
```diff
@@ -126,7 +126,6 @@ describe(`tests/enableCache/perspectives.js, api: GET ${path} >`, () => {
       }
 
       expect(res.body).to.have.length(ONE);
-      expect(res.body[ZERO].lensId).to.not.exist;
```
jgraff2: This is going to be added back after the TODO is done, right? Would it be better to comment this out until then?

tests/api/v1/common/testAssociations.js
```diff
@@ -90,7 +90,7 @@ function testAssociations(path, associations, joiSchema, conf) {
       .expect((res) => {
         expect(res.body).to.be.an('array');
         res.body.forEach((record) => {
-          expect(record).to.have.keys('id', 'name', assoc, 'apiLinks');
+          expect(record).to.contains.keys('id', 'name', assoc, 'apiLinks');
```
jgraff2: This will be changed back after the TODO is done, right?

api/v1/helpers/verbs/utils.js
cassiodias: Good question. I thought that was the way to define 'multiple associations'.

Is there a correct way to check that? Then it will be better when flagging the fields to exclude before sending back to the client.

package-lock.json
cassiodias: Sure, pushed by mistake.

tests/api/v1/common/testAssociations.js
cassiodias: code complete mistake.

tests/enableCache/perspectives.js
cassiodias: yeap

tests/api/v1/common/testAssociations.js
cassiodias: Yeap

api/v1/helpers/verbs/utils.js
jgraff2: This checks for multiple fields, but not all fields are associations. You would need to also check if the field has a foreign key, like you are doing elsewhere.

cassiodias: Can you approve that and merge to master? then I can address that within the part 3. @jgraff2

PR #1011, opened 2018-10-10
Collector integration tests

jgraff2:


tests/collector/integration.js
```diff
@@ -0,0 +1,957 @@
...
+ */
+
+/**
+ * tests/collector/integration.js
+ *
+ * This is a simulation of a full collector setup: one refocus instance
+ * and multiple separate collectors. The collector runs in a separate process,
+ * time is mocked, and requests are intercepted.
+ *
+ * To run this, you will need to have the collector source code in the same directory
```
iamigo: If we publish refocus-collector as npm module, would we then be able to make that a *dev* dependency and execute these integration tests?

jgraff2: Yeah, that should work, and it would allow us to set it up to run in travis.
It would mean we would have to pulish to npm any time something changes that could affect the tests.

jgraff2: published and updated.

jgraff2: Confirmed that it runs in travis now, but I'm disabling it for now because there are still flappers that need to be cleaned up in a future story.

tests/collector/utils/requestUtils.js
```diff
@@ -0,0 +1,272 @@
...
+    const { controller, method } = conf;
+    if (controller[method].restore) controller[method].restore();
+    const originalMethod = controller[method];
+
+    sinon.stub(controller, method)
+    .callsFake(callThenResolve.bind(null, originalMethod, conf));
+  }
+
+  function awaitRequest(conf, collectorName='') {
+    const timeout = conf.expectedInterval && conf.expectedInterval() * 1.5;
```
cassiodias: what is the default value if conf.expectedInterval is undefined?

jgraff2: If it's undefined no timeout will be set.

PR #1017, opened 2018-10-17
change resetJobCounter default

jgraff2:
The problem was that it was just too small of an interval. The reset has to be several times the cleanup or it will delete active jobs, which causes the test to timeout as it waits for completion.

### DIFF
--- a/config.js
+++ b/config.js
@@ -126,7 +126,7 @@ const clockJobConfig = {
     kueStatsActivityLogs: '1m',
     pubStatsLogs: '1m',
     queueStatsActivityLogs: '1m',
-    resetJobCounter: '24h',
+    resetJobCounter: '2h',
     sampleTimeout: '30s',
   },
   useWorker: {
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -153,6 +153,10 @@ Object.keys(clockJobConfig.intervals).forEach((jobName) => {
   const defaultValue = clockJobConfig.intervals[jobName];
   const envValue = pe[`CLOCK_JOB_INTERVAL:${jobName}`];
   const value = envValue || defaultValue;
+  console.log('Configuring clock job intervals: ' +
```
cassiodias: Any pros/cons of using: 

console.log(`This is a very long single line string which might be used \
to display assertion messages or some text. It has much more than \
80 symbols so it would take more than one screen in your text \
editor to view it. ${someVar}`);

iamigo: This is just a temporary message to help debug in a heroku environment :)

PR #1019, opened 2018-10-18
Adding method to add FK for multiple association just when needed

cassiodias:
The idea of this solution is adding the FK just when is really needed: user requires associations **and** the number of associations is more than 1.

* id, name doesn't need the extra FK;
* id, name, user doesn't need the extra FK;
*  id, name, currentCollector doesn't need the extra FK;
* id, name, user, currentCollector needs extra FK (createdBy & collectorId) otherwise the FKs won't be exposed in the inner query.

When limiting the case of manual FK the side effect is avoided for the other use cases since the multiple associations is a corner case.

### DIFF
--- a/api/v1/helpers/verbs/utils.js
+++ b/api/v1/helpers/verbs/utils.js
@@ -255,8 +255,9 @@ function multipleAssociations(helperModule, requestedAttributes) {
      (createdBy, collectorId, etc)
      */
     const extraForeignKeys = associations
-      .map((assocName) => helperModule.model.associations[assocName].foreignKey)
-      .map((foreignKey) => {
+      .map((assocName) => {
+        const foreignKey = helperModule.model.associations[assocName]
+          .foreignKey;
         const attribute = helperModule.model.attributes[foreignKey];
         return attribute ? attribute.fieldName : '';
       })
--- a/api/v1/helpers/verbs/utils.js
+++ b/api/v1/helpers/verbs/utils.js
@@ -237,9 +237,35 @@ function handleAssociations(reqObj, inst, props, method) {
   });
 } // handleAssociations
 
-function isMultipleAssociations(opts) {
-  const MULTIPLE_ASSOCIATIONS = 3;
-  return opts.attributes.length >= MULTIPLE_ASSOCIATIONS;
+/**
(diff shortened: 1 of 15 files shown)


api/v1/helpers/verbs/utils.js
```diff
@@ -237,9 +237,36 @@ function handleAssociations(reqObj, inst, props, method) {
...
+     (createdBy, collectorId, etc)
+     */
+    const extraForeignKeys = associations
+      .map((assocName) => {
+        const foreignKey = helperModule.model.associations[assocName]
+          .foreignKey;
+        const attribute = helperModule.model.attributes[foreignKey];
+        return attribute ? attribute.fieldName : '';
+      })
+      .filter((it) => it !== '');
```
jgraff2: I think it would be better to combine this check with line 262.

api/v1/helpers/verbs/utils.js
```diff
@@ -237,9 +237,36 @@ function handleAssociations(reqObj, inst, props, method) {
...
-function isMultipleAssociations(opts) {
-  const MULTIPLE_ASSOCIATIONS = 3;
-  return opts.attributes.length >= MULTIPLE_ASSOCIATIONS;
+/**
+ * When more than one association is required, Sequelize creates inner query
+ * but doesn't expose FK causing SQL failure.
+ *
+ * This method, in order to avoid Sequelize issue, will add all
+ * the foreign keys to the query when more than one association is required.
+ */
```
jgraff2: Good solution.
The foreign key will still be included in the response in the corner case of multiple associations, correct? I think that's fine, but can you add a test to specify that that's the expected behavior?

api/v1/helpers/verbs/utils.js
cassiodias: I am not sure if I got your point. 

Problem here is that a association doesnt have the FK in the model and there is no way to skip a .map iteration when that happens.

I have changed a bit doing a filter first getting just associations with FK in the model and then the .map will not be affected, will be 1:1 (association fk to model attribute fk).

Same result and number of iterations... maybe easier to read?!

jgraff2: You're right, makes sense.

PR #1020, opened 2018-10-18
change env var names

jgraff2:


jgraff2:
I already did, in the other PR. (1017)

PR #1021, opened 2018-10-23
Filter List By Type/Active

kriscfoster:
screen shot 2018-10-23 at 16 07 50
screen shot 2018-10-23 at 16 07 40


tests/view/rooms/list/app.js
```diff
@@ -0,0 +1,57 @@
...
+const app = require('../../../../view/rooms/list/app.js');
+
+describe('tests/view/rooms/list/app.js', () => {
+  const testArr = ['1', '2', '3'];
+
+  it('ok, select element created with 3 options', () => {
+    const selectEl = app.createSelectEl(testArr)
+    expect(selectEl.tagName).to.equal('SELECT');
+    expect(selectEl.childElementCount).to.equal(testArr.length);
+    expect(selectEl.value).to.equal('1');
```
cassiodias: is the first position the default value?

tests/view/rooms/list/app.js
```diff
@@ -0,0 +1,57 @@
...
+    const selectEl = app.createSelectEl(testArr, '2')
+    expect(selectEl.value).to.equal('2');
+  });
+
+  it('ok, constructing url after page filter change', () => {
+    const url = app.constructListFilterUrl('page', '3');
+    expect(url).to.equal('/rooms?page=3');
+  });
+
+  it('ok, constructing url after type filter change (page is reset to 1)', () => {
```
cassiodias: must reset page to 1 when changing the type filter?

cassiodias: if skiping the previous test, would this one pass? Tests should be independent.

api/v1/controllers/rooms.js
```diff
@@ -48,7 +51,23 @@ module.exports = {
...
+    const type = req.swagger.params.type;
+    if (type && !u.looksLikeId(type.value)) {
+      RoomType.findOne({
+        where: {
+          name: { [Op.iLike]: req.swagger.params.type.value },
+        },
+      })
+      .then((o) => {
+        if (o) {
+          req.swagger.params.type.value = o.dataValues.id;
```
cassiodias: if it's by type, does the SQL return just one RoomType? 

"o" is just one?

kriscfoster: @cassiodias we want it to just return 1 roomType. We're filtering rooms by their roomType and There should not be more than 1 roomType with the same name.

tests/view/rooms/list/app.js
kriscfoster: @cassiodias yeah we need to reset page to 1 when the type/active filter is changed as there may be 0 results. There is no state being stored in the function so all tests are independent

tests/view/rooms/list/app.js
kriscfoster: Yeah if no default value is passed to createSelectEl function, the first element in the array is used as the default

PR #1022, opened 2018-10-24
deleteSampleKeys - delete each one individually via separate del cmd

iamigo:
... instead of sending one del command with an array of sample hash keys to delete.

(I already tested this in sandbox, and it successfully deleted all the samples associated with an aspect when I deleted the aspect itself. The original code only deleted one of the associated samples and left the other sample hash keys as they were.)


cache/redisOps.js
```diff
@@ -182,17 +182,17 @@ function deleteSampleKeys(associationType, name) {
...
-    cmds.push(['srem', indexName, Array.from(keyArr)]);
-
-    // delete the hashes too
-    cmds.push(['del', Array.from(keyArr)]);
-    return redisClient.batch(cmds).execAsync();
+    /*
+     * Delete all the sample hashes and remove the entries from the master
+     * index of samples.
+     */
+    const sampleRemovalCmds = keyArr.map((key) => ['del', key]);
```
jgraff2: Any idea why this wasn't working? Redis docs say del can accept multiple keys... Maybe a bug in the npm redis client? Maybe something to do with running the commands as a batch?

Could there be other places in the code where we are doing the same thing that need to be changed?

iamigo: > Redis docs say del can accept multiple keys
Yes, and I've seen it work with a limited number of keys. I wonder if the size of the array matters? I haven't tried testing how many keys tips it over to failure, and I couldn't find any documented limits on the size of the array of keys in the redis or node_redis docs.

> ...other places in the code...
Yeah, I'll look into that.

iamigo: I updated one more place where we do this. The other place we do this in sampleStoreInit, which is deprecated and just waiting to be removed from codebase.

PR #1029, opened 2018-10-31
add travis_retry for npm scripts

iamigo:
This may be the "band-aid" approach for dealing with flappers in travis - [travis_retry](https://docs.travis-ci.com/user/common-build-problems/#travis_retry) will retry the command up to three times if the return code is non-zero.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,15 +41,15 @@ services:
   - redis-server
 
 script:
-  - npm run jscs-enforce
-  - npm run checkdb
-  - npm run test-travis
+  - travis_retry npm run jscs-enforce
+  - travis_retry npm run checkdb
+  - travis_retry npm run test-travis
 
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -43,13 +43,13 @@ services:
...
-  - npm run test-travis
+  - travis_retry npm run test-travis
 
 before_install:
-  - npm install -g npm@^5.7.0
+  - travis_retry npm install -g npm@^5.7.0
 
 install:
-  - npm ci
+  - travis_retry npm ci
```
jgraff2: Do we need the retry for before_install and install?

iamigo: The "typical" use case for this feature is to retry if you get network timeouts when trying to download dependencies, so I figured once I was retrying our tests, I may as well also protect against problems with npm component installation

PR #1030, opened 2018-10-31
Prev status

iamigo:


cache/models/samples.js
```diff
@@ -192,58 +192,88 @@ function cleanAddSubjectToSample(sampleObj, subjectObj) {
...
- * Update the status, previousStatus and changedAt fields if:
- *   a) sampObj is not present, or
- *   b) sampObj is present and previousStatus != current status.
+ * Update the sample from the request body. Updates in place. Compares to the
+ * previous state of the same sample, recalcuates status as needed.
  *
- * @param  {Object} qbObj - Query body object
- * @param  {Object} sampObj - Sample object
- * @param  {Object} aspectObj - Aspect object
+ * @param  {Object} qbObj - current samnpe (from request body)
```
cassiodias: typo

cache/models/samples.js
```diff
@@ -192,58 +192,88 @@ function cleanAddSubjectToSample(sampleObj, subjectObj) {
...
-      qbObj[sampFields.STATUS] = status;
+  if (!prev) {
+    /*
+     * This is a brand new sample so calculate current status based on value,
+     * set previous status to invalid, and set status changed at to now.
+     */
+    curr[sampFields.STATUS] =
+      sampleUtils.computeStatus(aspect, curr[sampFields.VALUE]);
+    curr[sampFields.PRVS_STATUS] = dbConstants.statuses.Invalid;
+    curr[sampFields.STS_CHNGED_AT] = now;
```
cassiodias: shouldnt be STS_CHANGED_AT?

tests/cache/models/samples/updateSampleAttributes.js
```diff
@@ -0,0 +1,346 @@
...
+    };
+    updateSampleAttributes(qb, samp, asp);
+    expect(qb).to.deep.equal({
+      name: '___Subject|___ThreeHours',
+      value: 2,
+      subjectId: '323ef102-3b64-458d-b134-29d2cd840bb2',
+      aspectId: '1979b734-bece-446e-9cb1-03917f60788e',
+      previousStatus: 'Invalid',
+      statusChangedAt: '2018-10-30T20:24:37.053Z',
+      status: 'Invalid',
```
jgraff2: This is invalid because the value is not a string, right? That's not very clear, is there a reason for writing the test this way?

tests/cache/models/samples/updateSampleAttributes.js
```diff
@@ -0,0 +1,346 @@
...
+      createdAt: '2018-10-29T22:42:30.918Z',
+      writers: [],
+    };
+    updateSampleAttributes(qb, samp, asp);
+    expect(qb).to.deep.equal({
+      name: '___TEST_SUBJECT|___TEST_ASPECT',
+      value: '',
+      subjectId: 'd2dfc000-5498-4bf0-a8c4-9c42c4569f05',
+      aspectId: '50513365-f24f-455c-8d47-c507d1c62a96',
+      previousStatus: 'Critical',
```
jgraff2: Why isn't this invalid?

iamigo: In this case, the user passed in a "value" attribute, but the value itself (undefined) does not map to a status, so that means the new status is "Invalid" and the previousStatus becomes the "status" from the previous version of the this sample (samp), which was "Critical".

tests/cache/models/samples/updateSampleAttributes.js
iamigo: In this test, it doesn't actually matter what the value is -- this test is really about exercising the block of code we enter when this is a brand new sample, i.e. we don't have a "prev" sample to compare to.

cache/models/samples.js
iamigo: fixed

cache/models/samples.js
iamigo: fixed

PR #1031, opened 2018-11-05
Generator template API

cassiodias:
**Task: API returning Generator Templates.**

* Added API to return a single Generator Template by /$name/$version
* Fixed generatorTemplates?name=$name returning a list of Generator
  * Added the attribute **nameFinderWithoutLimit** to the noun which will skip the limit verification when unique field is used (overwritten or default).

I have added a new **findOne** module returning a single object because the query is not by the key (via doGet) and it must returns a single object (original doFind returns an Array).

_Item to remove API querying by name will be handled in a separate story._

### DIFF
--- a/api/v1/controllers/generatorTemplates.js
+++ b/api/v1/controllers/generatorTemplates.js
@@ -50,6 +50,20 @@ module.exports = {
   },
 
   /**
+   * GET /generatorTemplates/{name}/{version}
+   *
+   * Retrieves the generatorTemplate by name and version and sends it back
+   * in the response.
+   *
+   * @param {IncomingMessage} req - The request object
+   * @param {ServerResponse} res - The response object
+   * @param {Function} next - The next middleware function in the stack
+   */
+  getGeneratorTemplateByNameAndVersion(req, res, next) {
+    doFind(req, res, next, helper);
+  },
+
+  /**
    * PATCH /generatorTemplates/{key}
    *
    * Modifies the generatorTemplate and sends it back in the response.
--- a/api/v1/controllers/generatorTemplates.js
+++ b/api/v1/controllers/generatorTemplates.js
@@ -14,6 +14,7 @@
 const helper = require('../helpers/nouns/generatorTemplates');
 const doDeleteOneAssoc = require('../helpers/verbs/doDeleteOneBToMAssoc');
 const doFind = require('../helpers/verbs/doFind');
+const doFindOne = require('../helpers/verbs/doFindOne');
 const doGet = require('../helpers/verbs/doGet');
 const doPatch = require('../helpers/verbs/doPatch');
 const doPost = require('../helpers/verbs/doPost');
@@ -60,7 +61,7 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   getGeneratorTemplateByNameAndVersion(req, res, next) {
-    doFind(req, res, next, helper);
+    doFindOne(req, res, next, helper);
   },
 
   /**
--- a/api/v1/helpers/nouns/generatorTemplates.js
+++ b/api/v1/helpers/nouns/generatorTemplates.js
@@ -36,5 +36,7 @@ module.exports = {
   fieldScopeMap: {
     user: 'user',
   },
+  uniqueFieldWithoutLimit: 'name',
+
 }; // exports
 
--- a/api/v1/helpers/nouns/generatorTemplates.js
+++ b/api/v1/helpers/nouns/generatorTemplates.js
@@ -36,7 +36,8 @@ module.exports = {
   fieldScopeMap: {
     user: 'user',
   },
-  uniqueFieldWithoutLimit: 'name',
+
+  nameFinderWithoutLimit: true,
 
 }; // exports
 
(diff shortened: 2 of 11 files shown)


api/v1/helpers/verbs/doFind.js
```diff
@@ -33,19 +32,20 @@ const config = require('../../../../config');
...
 function doFindAndCountAll(reqResNext, props, opts, resultObj) {
-  return u.getScopedModel(props, opts.attributes).findAndCountAll(opts)
-  .then((o) => {
-    resultObj.dbTime = new Date() - resultObj.reqStartTime;
-    reqResNext.res.set(COUNT_HEADER_NAME, o.count);
-    return o.rows.map((row) => {
-      if (props.modelName === 'Lens') {
-        delete row.dataValues.library;
-      }
+  return u.getScopedModel(props, opts.attributes)
```
cassiodias: Just formatting

pallavi2209:
@cassiodias We have been asked not to put story numbers in PRs because of security reasons.

cassiodias:
@pallavi2209 thx, removed

api/v1/helpers/nouns/generatorTemplates.js
```diff
@@ -36,5 +36,7 @@ module.exports = {
   fieldScopeMap: {
     user: 'user',
   },
+  uniqueFieldWithoutLimit: 'name',
```
jgraff2: I don't think this field name makes sense - the "name" fileld isn't actually unique

jgraff2: We don't really need a new field for this, all we need to do is specify that this model shouldn't treat name as a unique field by default, whereas all the others do.
It might make more sense to just set the existing "nameFinder" to an empty string, or false/null.
Or change the nameFinder behavior so it always has to be specified.

api/v1/controllers/generatorTemplates.js
```diff
@@ -49,6 +50,20 @@ module.exports = {
...
+   *
+   * Retrieves the generatorTemplate by name and version and sends it back
+   * in the response.
+   *
+   * @param {IncomingMessage} req - The request object
+   * @param {ServerResponse} res - The response object
+   * @param {Function} next - The next middleware function in the stack
+   */
+  getGeneratorTemplateByNameAndVersion(req, res, next) {
+    doFindOne(req, res, next, helper);
```
jgraff2: Is it necessary to set up a whole new helper for this? We're already enforcing that a name/version pair is unique, so we can be sure we would only get one result if we use doFind.

cassiodias: doFind returns an array and the expectation is the object.

api/v1/helpers/nouns/generatorTemplates.js
cassiodias: The first approach was as you said... but...

Setting **nameFinder** and using **doFind** (/generatorTemplates/?name=template1): it works fine because it will not go into the limit and when we have same name with more than one version, they will be retrieved.

The problem is when we execute doGet which will look for key (key can be **ID**, **name** or  **nameFinder**). 

**Getting by id (doGet)**

**1)**
/generatorTemplates/**template1** and **nameFinder** = true (for instance as you said).

findBykey, util.js
`opts.where[props.nameFinder || 'name'] = keyClause`

Query will be `where true = template1` and the will not retrieve by name.

**2)**
/generatorTemplates/123-456-789 and **nameFinder** = true.
We will look by ID first and the method **findByIdThenName** removes the where clause retrieving by ID.

Actually for GeneratorTemplate using this approach solves even the story to remove GET by name, however, we are using **nameFinder** to generate a wrong SQL.

I have added this new flag to avoid any side effect related to this concept for nameFinder/uniqueField and 'name' as default option.

And _"the "name" fileld isn't actually unique"_ - I know, but it's default if there is no nameFinder.

Sorry this long comment, but I think now that just satting nameFinder as true will solve as we just look by KEY using ID or name. There are no side effect... hope we can remember why where are executing true = 'templateName' in the future.

api/v1/controllers/generatorTemplates.js
jgraff2: Hmm. Makes sense, that would be weird.
Not ideal, but this is probably the best we can do short of completely refactoring the helpers.

We could at least get rid of any code in the new helper that not's necessary for this case, like the caching.

cassiodias: yeap! We have to refactor those helpers and revisit some steps.

Thx for pointed out the cache, if you see something else let me know. I don't have to much context to deliberately remove them.

api/v1/helpers/nouns/generatorTemplates.js
cassiodias: @iamigo @jgraff2 do you agree setting nameFinder and generating a wrong SQL when searching for generatorTemplate/${name}?

api/v1/helpers/verbs/doFind.js
```diff
@@ -15,7 +15,6 @@ const fu = require('./findUtils');
 const COUNT_HEADER_NAME = require('../../constants').COUNT_HEADER_NAME;
 const httpStatus = require('../../constants').httpStatus;
 const redisCache = require('../../../../cache/redisCache').client.cache;
-const config = require('../../../../config');
```
cassiodias: removed bcz not used.

api/v1/helpers/verbs/findUtils.js
```diff
@@ -362,15 +372,13 @@ function options(params, props) {
...
   }
 
-  if (filter) {
-    opts.where = toSequelizeWhere(filter, props);
-    if (props.modifyWhereClause) {
-      props.modifyWhereClause(params, opts);
-    }
-
-    applyLimitIfUniqueField(opts, props);
+  opts.where = toSequelizeWhere(filter, props);
```
cassiodias: if (filter) is always **true**

PR #1032, opened 2018-11-05
track active namespaces

jgraff2:


realtime/utils.js
```diff
@@ -310,6 +311,21 @@ function initializePerspectiveNamespace(inst, io) {
   return io;
 }
 
+function addEventsForNamespaces(io) {
+  Object.keys(io.nsps).forEach((n) => {
+    const namespace = io.of(n);
+    namespace.on('connect', (socket) => {
+      connectedNamespaces.addNamespace(n);
+      socket.on('disconnect', function () {
+        const namespace = io.of(n);
```
cassiodias: it's necessary again io.of(n)?

jgraff2: Technically no, since io.of() returns the namespace object directly so by the time we're in the disconnect event namespace.connected should have the updated value. But I had to check the code to confirm that; it could just as easily be the case that io.of() returns a cloned namespace object. I did it this way to make it clear that we are getting the up-to-date namespace object and we can rely on namespace.connected being accurate.

cassiodias: yeah, make sense. Cool

PR #1033, opened 2018-11-06
clock job to auto-delete tokens

iamigo:
Add class-level function to the Token model to delete tokens not used since some specified threshold (e.g. "-30d"). Add clock job to call that fn. Add tests.


clock/scheduledJobs/deleteUnusedTokens.js
```diff
@@ -0,0 +1,41 @@
...
+   * Use environment variable name "CLOCK_JOB_INTERVAL_deleteUnusedTokens" to
+   * set the frequency of the clock job, i.e. how often should we check for
+   * unused tokens. Set the value to a time offset like "12h" if you want to
+   * run the clock job every twelve hours or "1d" if you want to run the job
+   * once a day.
+   *
+   * Use environment variable name "DELETE_UNUSED_TOKENS_SINCE" to set the
+   * threshold of what we should consider "unused". Set the value to a negative
+   * time offset like "-30d" if you want to consider a token unused if it has
+   * not been used in the last 30 days, or "-3m" for three months, etc.
```
jgraff2: "3m" would be 3 minutes. ms doesn't have a month operator, but you can do "3w" or "1y".

db/model/token.js
```diff
@@ -109,6 +109,36 @@ module.exports = function token(seq, dataTypes) {
...
+   */
+  Token.deleteUnused = function (since) {
+    try {
+      // eslint-disable-next-line no-magic-numbers
+      const millis = ms(since) || 0;
+      if (millis < 0) { // eslint-disable-line no-magic-numbers
+        return Token.destroy({
+          where: {
+            lastUsed: {
+              [Op.lt]: new Date(new Date().getTime() + ms(since)),
```
jgraff2: I think this would be more readable with Date.now() so we're not doing new Date() twice.

README.md
```diff
@@ -44,11 +44,24 @@ See the [Quickstart](https://salesforce.github.io/refocus/docs/01-quickstart.htm
...
+    * Use environment variable name
+        `CLOCK_JOB_INTERVAL_deleteUnusedTokens` to set the frequency of
+        the clock job, i.e. how often should we check for unused tokens.
+        Set the value to a time offset like `12h` if you want to run the
+        clock job every twelve hours or `1d` if you want to run the job
+        once a day.
+    * Use environment variable name `DELETE_UNUSED_TOKENS_SINCE` to set
+        the threshold of what we should consider "unused". Set the value
+        to a negative time offset like `-30d` if you want to consider a token unused if it has
+        not been used in the last 30 days, or `-3m` for three months, etc.
```
jgraff2: "3m" would be 3 minutes. ms doesn't have a month operator, but you can do "3w" or "1y".

PR #1036, opened 2018-11-10
log collector assignment

jgraff2:


db/model/generator.js
```diff
@@ -481,14 +482,44 @@ module.exports = function generator(seq, dataTypes) {
...
-  };
+  }; // assignToCollector
+
+  function logAssignment(gen, prevColl, newColl) {
+    prevColl = prevColl && prevColl.name || null;
+    newColl = newColl && newColl.name || null;
+
+    let action;
+    let type;
+    if (!prevColl && newColl) {
```
iamigo: I think it might be easier to follow in the logs with just a "from" and a "to" attribute and inferring "type", rather than actual deriving "type" because we would end up having to parse the action and wrap the value in quotes since it would have spaces... so what I would propose would be:

when reassigning: `activity=collectorAssignment from=frodo to=bilbo`
when assigning new: `activity=collectorAssignment from=None to=frodo`
when unassigning without reassigning: `activity=collectorAssignment from=bilbo to=None`

I wonder if we might also be able to say something about why it's happening, e.g.did someone modify the SG? or is there a newer match for the SGT version? or did the coll miss a heartbet? or was the coll stopped? paused? other... ?

jgraff2: This has "from" and "to" as the "previousCollector" and "newCollector" fields. I was thinking those would be the main fields used for sumo queries, I just added the type and action as extras that might be useful for filtering or for reading individual log lines.

I can change the field names to "from" and "to" if that's more clear.
Is having the extra fields a problem? Worst case we just never use them, and I do think they could be useful.

I did look into tracing the source of the change, but it would be difficult because re-assignment is called from several different places, some of which are db hooks which themselves can be triggered from several different places. I figured the main question is just whether the assignment was triggered by an API update or a missed heartbeat, so I added separate logging for missed heartbeat to answer that.

PR #1041, opened 2018-11-19
Adding collector SSL only

cassiodias:
* Defining environment variable enforcing "SSL only for collectors".
    * added config attribute with default value;
    * changed swagger definition;
    * added test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ See the [Quickstart](https://salesforce.github.io/refocus/docs/01-quickstart.htm
         the threshold of what we should consider "unused". Set the value
         to a negative time offset like `-30d` if you want to consider a
         token unused if it has not been used in the last 30 days.
-1. If you want Refocus Collectors collecting only from sources using HTTPS, set environment variable `COLLECTOR_REQUIRE_SSL_TO_REMOTE_DATA_SOURCE` to `true`.
+1. If you want to require SSL for connections from Refocus Collectors to the remote data sources, set environment variable ``COLLECTOR_REQUIRE_SSL_TO_REMOTE_DATA_SOURCE`` to ``true.``
 
 ### IP Restrictions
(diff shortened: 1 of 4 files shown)


config/collectorConfig.js
```diff
@@ -31,6 +31,8 @@ if (heartbeatLatencyToleranceMillis < FLOOR) {
   heartbeatLatencyToleranceMillis = 1000 * LATENCY_DEFAULT;
 }
 
+const sslOnly = +pe.COLLECTOR_SSL_ONLY || false;
```
iamigo: We use the "+" symbol when we want to try to coerce the env var to numeric value.
In this case here, we would want to set sslOnly to true if the env var **string** was "true" (and probably also accept "True"|"TRUE", but not "False" or "FALSE" or "false" ...

So I would suggest something like this:

```
const requireSslToRemoteDataSource = false;
if (pe.COLLECTOR_REQUIRE_SSL_TO_REMOTE_DATA_SOURCE && pe.COLLECTOR_REQUIRE_SSL_TO_REMOTE_DATA_SOURCE.toLowerCase() === 'true') {
  requireSslToRemoteDataSource = true;
}
```

iamigo: Oh, we should also write this up in the README section "Securing Refocus" so that a Refocus admin knows how to do this.

cassiodias: good point, thx

cassiodias: Done

README.md
```diff
@@ -62,6 +62,7 @@ See the [Quickstart](https://salesforce.github.io/refocus/docs/01-quickstart.htm
         the threshold of what we should consider "unused". Set the value
         to a negative time offset like `-30d` if you want to consider a
         token unused if it has not been used in the last 30 days.
+1. If you want Refocus Collectors collecting only from sources using HTTPS, set environment variable `COLLECTOR_REQUIRE_SSL_TO_REMOTE_DATA_SOURCE` to `true`.
```
iamigo: If you want to require SSL for connections from Refocus Collectors to the remote data sources, set environment variable `COLLECTOR_REQUIRE_SSL_TO_REMOTE_DATA_SOURCE` to `true`.

PR #1042, opened 2018-11-19
disable GET /name for generatorTemplates

pallavi2209:
The user must NOT be able to GET Generators Template by name /v1/generatorTemplates/$name, however, must be able to GET by /generatorTemplates/$ID.

- Return return 400 (invalid) error when the query is by name.
- Updated/Added tests.

### DIFF
--- a/api/v1/apiErrors.js
+++ b/api/v1/apiErrors.js
@@ -153,6 +153,17 @@ errors.create({
   'in your Sample Generator\'s context',
 });
 
+errors.create({
+  scope: exports,
+  code: 11112,
+  status: 400,
+  name: 'InvalidKey',
+  parent: this.ValidationError,
+  fields: [],
+  defaultMessage: 'You cannot GET this resource by name, try using id as ' +
+    'key or name as query parameter, e.g. ?name=<name>',
+});
+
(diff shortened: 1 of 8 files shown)


api/v1/apiErrors.js
```diff
@@ -153,6 +153,17 @@ errors.create({
...
 });
 
+errors.create({
+  scope: exports,
+  code: 11112,
+  status: 400,
+  name: 'InvalidKey',
+  parent: this.ValidationError,
+  fields: [],
+  defaultMessage: 'You cannot GET this resource by name, try using id as ' +
```
iamigo: "This resource has a multipart key. Try using name as a query parameter to return a list of all records with this name, e.g. ?name=<name>"

api/v1/helpers/nouns/generatorTemplates.js
```diff
@@ -37,7 +37,7 @@ module.exports = {
     user: 'user',
   },
 
-  nameFinderWithoutLimit: true,
+  nonUniqueName: true,
```
iamigo: how about "hasMultipartKey" as the var name?

PR #1046, opened 2018-11-29
Collector model: add collectorGroupId column for assoc

pallavi2209:


iamigo:
I pulled this branch and ran npm install and then npm start. I got this:

Loaded configuration file "config/migrationConfig.js".
Using environment "development".
== 20181129172340-collector-add-collectorGroup-assoc: migrating =======

...

ERROR: column "collectorGroupId" of relation "Collectors" already exists

[./db/migrate] Error: Command failed: node_modules/.bin/sequelize db:migrate
...

ERROR: column "collectorGroupId" of relation "Collectors" already exists

pallavi2209:
@iamigo I think you should have done npm cleanup on your local (which recreated the full db just from models) and hence you already have the column in your localhost db. The right test would be - restore staging db on your localhost and then "npm run checkdb" (which should run migration) and "npm run postinstall".

PR #1047, opened 2018-12-03
Redirect to different instance of refocus if feature toggle is enabled

kriscfoster:
**_config/toggles.js_**
- Long term feature toggle for redirecting to a different instance of refocus

**_index.js_**
- Allow default landing page to be configured with an env var, otherwise default to /perspectives

**_tests/api/v1/index/v1index.js_**
- Test for setting default landing page with env var

**_view/loadView.js_**
- Added function getRedirectURI which gets a redirect uri based on environment variables. If it should not redirect anywhere then just returns an empty string.
- Use this getRedirectURI function in loadView if feature toggle is enabled
- Export the function for testing

**_tests/view/loadView.js_**
- Tests the getRedirectURI function with various scenarios

**_package.json_**
- Needed to add extra test script to test loadView. Running the loadView test inside the old script test-view script breaks because the loadView file has dependences which break during compilation. Went over with @cassiodias and we think this is the best workaround


kriscfoster:
separating this into 2 PRs from meeting with Ian

PR #1049, opened 2018-12-05
Reassign generators among running collectors to distribute load

pallavi2209:
Implement this by changing the assignment logic in Generator.assignToCollector:
Out of all the possible collectors that are running and alive, find the one with the fewest number of currently assigned generators. Break ties based on the number of possible generators each collector has.
@jgraff2 I did not go through raw sql query route because its not always that we are going to db to get info, we go to db only when we do not have currentGenerators or possibleGenerators on collector object.

### DIFF
--- a/db/model/generator.js
+++ b/db/model/generator.js
@@ -479,7 +479,6 @@ module.exports = function generator(seq, dataTypes) {
     .then(() => g);
   }; // updateForHeartbeat
 
-
   /**
    * Out of all the possible collectors that are running and alive, find the
    * one with the fewest number of currently assigned generators. Break ties
--- a/db/model/generator.js
+++ b/db/model/generator.js
@@ -160,65 +160,110 @@ module.exports = function generator(seq, dataTypes) {
       }, // beforeCreate
 
       beforeUpdate(inst /* , opts */) {
-        return Promise.all([
-          Promise.resolve().then(() => {
-            let isCurrentCollectorIncluded = true;
+        if (featureToggles.isFeatureEnabled('distributeGenerators')) {
+          const promises = [];
 
-            // if possibleCollectors have changed, check if the currentCollector
-            // is still included
-            if (inst.possibleCollectors && inst.changed('possibleCollectors')) {
-              isCurrentCollectorIncluded = inst.possibleCollectors.some(
-                (coll) => coll.name === inst.currentCollector.name
-              );
-            }
+          let isCurrentCollectorIncluded = true;
 
-            /*
-             Assign to collector in following cases:
-             1) isActive is changed.
-             2) If possibleCollectors are changed and this generator current
-              collector is not included in the changed possibleCollectors.
-             3) If possibleCollectors are changed and collector is not assigned
-              to generator
-             */
-            if (
-                inst.changed('isActive') ||
-                !isCurrentCollectorIncluded ||
-                (inst.changed('possibleCollectors') && !inst.currentCollector)
-               ) {
-              return inst.assignToCollector();
-            }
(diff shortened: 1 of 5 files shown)


db/model/generator.js
```diff
@@ -475,22 +570,29 @@ module.exports = function generator(seq, dataTypes) {
...
-    // a database reload.
-    this.collectorId = newColl ? newColl.id : null;
-    this.currentCollector = newColl || null;
+    return Promise.resolve()
+    .then(() => {
+      const possibleCollectors = this.possibleCollectors;
+      if (this.isActive && possibleCollectors && possibleCollectors.length) {
+        return getMostAvailableCollector(possibleCollectors)
+        .then((mostAvailCollector) => {
+          logAssignment(this.name, this.currentCollector, mostAvailCollector);
```
iamigo: Shouldn't we check if we got a collector here since the fn *can* return "empty" Promise.resolve() on line 495?

db/model/generator.js
```diff
@@ -475,22 +570,29 @@ module.exports = function generator(seq, dataTypes) {
...
+          // needing a database reload.
+          this.collectorId = mostAvailCollector ? mostAvailCollector.id : null;
+          this.currentCollector = mostAvailCollector || null;
+          return Promise.resolve();
+        });
+      }
+
+      this.collectorId = null;
+      this.currentCollector = null;
+      return Promise.resolve();
```
iamigo: does it make testing any easier or more transparent if this fn returns the collector it assigned to?

pallavi2209: Hmm - thats good point - what do you think @jgraff2

pallavi2209:
@iamigo

> Do you feel like there is sufficient logging to troubleshoot this?

We have logging from "logAssignment" on what was previousCollector and what is new collector selected, but we do not have logging on what happens inside this _getMostAvailableCollector_ function. We can always add more log lines to help debugging but the question would be how much detail we need in logs?
One log line format (for collectorAssignment activity) could be: 
```
{
generator: ..,
possibleCollectors: [c1,c2,...],
aliveAndRunningCollectors: [
     c1: { numberOfCurrentGenerators: 2,
             numberOfCurrentGenerators:3
            },
     c3: { numberOfCurrentGenerators: 2,
             numberOfCurrentGenerators:4
            }
],
breakTiesUsingPossibleGenerators: [c1,c3]
mostAvailableCollector: c1,
}
```
This could be helpful if it does not overwhelm our logs.

> What if we need to roll it back? Can we wrap in a short-term toggle?
Yes, we should do this for sure - esp when I will not be here when this is deployed. Let me update the PR.

iamigo:
Let's hold off on expanding the logging for now but let's definitely wrap it in a feature toggle

db/model/generator.js
```diff
@@ -467,6 +479,89 @@ module.exports = function generator(seq, dataTypes) {
...
+    }
+
+    const promises = [];
+
+    /* For all collectors, create a map of collector and its corresponding
+     current and possible generators */
+    for (let i = 0; i < upCollectors.length; i++) {
+      const ithCollector = upCollectors[i];
+      const promise = new Promise((resolve, reject) => {
```
jgraff2: We could get rid of the wrapping promise by setting the "promise" var directly:

```
let promise;
if (ithCollector.currentGenerators...) {
  promise = Promise.resolve({ ... });
} else {
  promise = ithCollector.getCurrentGenerators.then(...)
}
```

pallavi2209:
@iamigo @jgraff2 Updated the PR. Fyi, I have duplicated much code so that its easier with toggle.

PR #1055, opened 2018-12-11
allow refocus admin override of collector retry behavior

iamigo:


jgraff2:
package-lock not needed

PR #1065, opened 2018-12-19
Bulk POST Events

kriscfoster:
api/v1/controllers/events.js
- bulkPostEvent controller method which will respond with { "Status": "OK" } before the create has happened

api/v1/swagger.yaml
- /events/bulk route
- BulkCreateResponse

db/model/event.js
- bulkCreate class method

tests/api/v1/events/postBulk.js
- Tests for bulk posting


kriscfoster:
@iamigo changed this so it is now asynchronous

api/v1/controllers/events.js
```diff
@@ -74,4 +76,39 @@ module.exports = {
...
+   * indicating merely that the bulk create request has been received.
+   */
+  bulkPostEvent(req, res, next) {
+    const resultObj = { reqStartTime: req.timestamp };
+    const value = req.swagger.params.queryBody.value;
+    const body = { status: 'OK' };
+
+    /**
+     * Performs bulk upsert through worker, cache, or db model.
+     * Works regardless of whether user if provided or not.
```
iamigo: if --> is

api/v1/controllers/events.js
```diff
@@ -74,4 +76,39 @@ module.exports = {
...
+     *
+     * @param {Object} user Sequelize result. Optional
+     * @returns {Promise} a promise that resolves to the response object
+     * with status and body
+     */
+    function bulkPost(user) {
+      helper.model.bulkCreate(value, user);
```
iamigo: no way to know if this succeeds or fails?

kriscfoster: @iamigo I was following the same pattern as here: https://github.com/salesforce/refocus/blob/master/api/v1/controllers/samples.js#L448, except for in this PR we won't be using worker dynos. What would the action be if this succeeds/fails, as the response has already been sent to the client at this stage.

iamigo: The response says "ok we got your request" but it can't say "yes, we saved these events to the db" -- what if there are errors saving the records? That's why for bulk operations we give users the ability to get job status with array of errors if any

kriscfoster:
@iamigo with all of that workerJob stuff & tests etc, this has turned into a massive PR. It works really nice but if you have any suggestions on how to split it up I'm up for that because I understand this size of a PR is very difficult to review.

If enableWorkerProcess is toggled on then the POST events/bulk response will contain a jobId and we can use this to find the status.

kriscfoster:
@iamigo Sounds good, thanks for letting me know. Enjoy the holiday season!

kriscfoster:
@iamigo I followed the same pattern for testing bulkPostEvents as was done for bulk upserting subjects and as a result the tests in tests/jobQueue/v1/bulkPostEvents.js all rely on the API to actually work.

If I were to split this up, then the first PR with the job/worker stuff first then this PR would be very light on tests as the tests wouldn't be able to utilize the API but following PRs would include all the tests. Instead, if you'd like I could just take all of the get job status stuff out of this PR and put it in a different PR

iamigo:
nah, just leave it all in here

PR #1066, opened 2018-12-22
collector integration tests - test upsert intervals

jgraff2:


iamigo:
@jgraff2 
[31m  1 failing[0m

[0m  1) tests/collector/integration.js > assignment > collector status changed > missed heartbeat:

      [31mAssertionError: expected 1000 to equal 60000[0m
      [32m+ expected[0m [31m- actual[0m

      [31m-1000[0m
      [32m+60000[0m
      [0m[90m
      at awaitMatch.then.then.then (tests/collector/utils/requestUtils.js:288:33)

jgraff2:
Tests won't pass until the collector-side fix is published.
Notice that it's failing with a value of 1000 - that's because duplicate repeaters are being created and are running 1s apart.

PR #1068, opened 2018-12-31
Socket.io change - use rooms as well as namespaces

jgraff2:
These are the required changes in the main app for the new namespace format. (implemented in the realtime app)
- use the new format, based on a toggle, in the perspective and room pages
- changes to publishing so the event emitter has the information it needs for the new format
- comments marking code for the old format that can be removed once we've made the transition

Rollout process: This can be deployed at any time, as long as the toggle isn't enabled.
The toggle should be enabled right after it is enabled in the realtime app.

Main PR: salesforce/refocus-real-time#3


view/rooms/index.pug
```diff
@@ -77,5 +77,6 @@ html
     var trackingId = '#{trackingId}';
     var user = '#{user}';
     var userSession = '#{userSession}';
+    var useNewNamespaceFormat = #{useNewNamespaceFormat};
```
BrandonMurnane: should this be surrounded in ' ' like the others?

jgraff2: This is a boolean, whereas the others are strings.

iamigo:
@jgraff2 is this ready to merge?

jgraff2:
@iamigo I think we should wait for approval from everyone on all the PRs before merging any of them. There's also some open discussion on the bdk PR.

jgraff2:
I added a commit to split the toggles separately for perspectives and rooms. That way we can enable the new format just for perspectives and let rooms continue to use the old format until imc is ready to switch over.

@iamigo @pallavi2209 please review.

pallavi2209:
LGTM

jgraff2:
FYI @cassiodias @BrandonMurnane @ConalCosgrove @kriscfoster 
We will be moving forward with these changes, enabled for perspectives only. IMC will continue to use the old format until you're ready to switch over.

PR #1083, opened 2019-01-17
remove all the extra sequelize stuff from real-time event payload

iamigo:


jgraff2:
@iamigo @pallavi2209 
Any ideas why coveralls has been reporting 68%? I don't see how it could be anything we changed.

PR #1084, opened 2019-01-17
Set room origin

tausifmuzaffar:
Anytime a room is created from web we want to track where the room was created by setting origin


tausifmuzaffar:
@ConalCosgrove waiting for the origin route to be added

ConalCosgrove:
@iamigo yes, this was waiting for the previous story which added the 'origin' field to the room table in the db. Ready to be merged now that one is in

PR #1088, opened 2019-01-23
Bot token workflow fix

BrandonMurnane:
Fixed where token would have bot name twice instead of username
Followed same as other workflows for api token for bots
Added new function for checking socket.io.handshake


harshkothari410:
What is difference between verifyBotSocketToken and verifyBotToken ? @BrandonMurnane

BrandonMurnane:
One is dealing with requests. The socket one happens after a socket
handshake

On 29 Jan 2019, at 18:20, Harsh Kothari <notifications@github.com> wrote:

What is difference between verifyBotSocketToken and verifyBotToken ?
@BrandonMurnane <https://github.com/BrandonMurnane>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/salesforce/refocus/pull/1088#issuecomment-458649730>,
or mute the thread

utils/jwtUtil.js
```diff
@@ -167,9 +167,41 @@ function verifyGeneratorToken(payload, req, cb) {
...
+function verifyBotSocketToken(token) {
   return jwtVerifyAsync(token, secret, {})
-  .then((payload) => Bot.findOne({ where: { name: payload.username } }));
+  .then((payload) => Bot.findOne({ where: { name: payload.tokenname } }));
+} // verifyBotSocketToken
+
+/**
+ * Function to verify if a generator token is valid or not, i.e. does the token
+ * name match an actual generator name.
+ *
```
kriscfoster: comment above refers to generator

utils/jwtUtil.js
```diff
@@ -167,9 +167,41 @@ function verifyGeneratorToken(payload, req, cb) {
...
+ * @param {Object} payload - the decoded payload
+ * @param  {object} req - request object
+ * @param  {Function} cb - callback function - Optional
+ * @returns {Function|undefined} - Callback function or undefined
+ * @throws {ForbiddenError} If a generator record matching the token name is
```
kriscfoster: reference to generator again

kriscfoster:
@BrandonMurnane tests failing

realtime/setupSocketIO.js
```diff
@@ -185,7 +185,7 @@ function init(io, redisStore) {
 
     // check if connection using an authorization header
     } else if (socket.handshake.headers.authorization) {
-      jwtUtils.verifyBotToken(socket.handshake.headers.authorization)
+      jwtUtils.verifyBotSocketToken(socket.handshake.headers.authorization)
```
harshkothari410: Here is there any way you can use verifyToken method ? So I can authorize any token over there. i.e user token

BrandonMurnane: not really the verifyToken method is for verifying tokens that come in from requests not from socket connections it requires a request object

harshkothari410: but you can do something like this verifyToken(socket.handshake) ? right ?

BrandonMurnane: doing this refocus will crash due to their not being any session field inside the socket.handshake

kriscfoster:
@harshkothari410 @iamigo @jgraff2 is there anything blocking this PR from going in?

realtime/setupSocketIO.js
harshkothari410: Can you make one more function to verifySocketToken ? And use that method over there so we can use user socket token as well in future for verifying user as well. This seems hard coded for bot only. @BrandonMurnane  @iamigo

BrandonMurnane: i think thats outside the scope of the stories this code was for if we can get this in and another story made for making a generic verifysockettoken we can prioritize that

BrandonMurnane:
@harshkothari410 if i understand correctly you guys would like the socket token to go through the same or a similar function to verify token to be more generic. Is that right ?

harshkothari410:
Correct @BrandonMurnane so that we can verify user as well.

PR #1089, opened 2019-01-23
add collector(s) to collector group

iamigo:
POST /v1/CollectorGroups/[GroupName]/collectors [ ... ] array of collector names

If any collector named in the array is already in the list or is already assigned to a different group, reject the whole request

Send an empty array to clear the whole list

### DIFF
--- a/db/model/collectorgroup.js
+++ b/db/model/collectorgroup.js
@@ -144,7 +144,7 @@ module.exports = function collectorgroup(seq, dataTypes) {
   /**
    * Add the named collectors to this collector group. Reject if any of the
    * named collectors is already assigned to this group, or to a different
-   * group. If array is empty, remove all collectors from this collector group.
+   * group.
    *
    * @param {Array<String>} arr - array of collector names
    * @returns {Promise<any | never>}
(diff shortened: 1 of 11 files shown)


db/model/collectorgroup.js
```diff
@@ -128,35 +141,20 @@ module.exports = function collectorgroup(seq, dataTypes) {
...
+   * named collectors is already assigned to this group, or to a different
+   * group. If array is empty, remove all collectors from this collector group.
+   *
+   * @param {Array<String>} arr - array of collector names
+   * @returns {Promise<any | never>}
+   */
+  CollectorGroup.prototype.addCollectorsToGroup = function (arr) {
+    return collectorUtils.validate(seq, arr)
+      .then(collectorUtils.alreadyAssigned)
+      .then((collectors) => this.setCollectors(collectors))
```
pallavi2209: Is this supposed to add to existing collectors? Should this be addCollectors instead? Test for this use case?

iamigo: Oh good point--thx!!!

PR #1095, opened 2019-01-28
Add origin column to Room table

ConalCosgrove:


db/model/room.js
```diff
@@ -61,6 +61,14 @@ module.exports = function room(seq, dataTypes) {
       allowNull: true,
       comment: 'Bot names to be used in rooms',
     },
+    origin: {
+      type: dataTypes.STRING,
```
iamigo: free text here is recipe for chaos! what are the values you actually expect to see here? would it be possible to make it an enum with "Other" as one of the options? then add options as needed down the line? it's not going to be infinite! i'm sure you can come up with a half dozen values or fewer

api/v1/swagger.yaml
```diff
@@ -7402,6 +7402,17 @@ paths:
           in: query
           required: false
           type: string
+        -
+          name: origin
+          description: Filter rooms by origin
+          in: query
+          required: false
+          type: string
+          enum:
```
kriscfoster: @ConalCosgrove do we want 'slack' here aswell?

migrations/20190128134129-add-room-origin-column.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+const TBL = 'Rooms';
+module.exports = {
+  up: (queryInterface, Sequelize) =>
+    queryInterface.addColumn(
+      TBL,
+      'origin', {
+        type: Sequelize.ENUM('auto_create', 'GUS', 'other', 'web'),
```
kriscfoster: @ConalCosgrove slack missing from the migration

ConalCosgrove: nice catch 👍🏻

db/model/room.js
ConalCosgrove: @iamigo good point! changed to enum and added required options

api/v1/swagger.yaml
```diff
@@ -7402,6 +7402,18 @@ paths:
...
+            - auto_create
+            - GUS
```
iamigo: is there any way to substitute a more generic (NON-SALESFORCE-Y) label for this?

iamigo: since you can create a room from a refocus sample, shouldn't that be an option too?

ConalCosgrove: If we're looking for more generic options then that might be an argument for free text rather than an enum, what do you think? Adding a refocus entry to the enum now

kriscfoster: We need to be careful here because the code is already in GUS to send the param origin=GUS. If we remove GUS out of the enum, then Refocus Rooms will not be able to be created from GUS anymore (it will fail on room create) unless GUS is released with a different origin=. I think there are 2 options:

1. Change enum to "TicketingSystem" and not release until GUS change is released
2. Change back to free text

BrandonMurnane: I was thinking that if in future we would like to have other specific origins like org62 etc that we dont want to have to keep coming up with other names for them

ConalCosgrove: Would free text result in useless data? We've set all of the different creation links for IMC rooms to include an origin="" param, so we can ensure they're coherent and should still be able to query by origin regardless of if it's an enum or not, right? 

Having the field as free text allows us to add new origins without a refocus code change and also stops the issue of the enum containing any "salesforce-y" terms or having broad values like "TicketingSystem" which going forward could make queries less granular.

Obviously free text leaves it more open for someone to input junk data, but in general it seems to fit the requirements of this field best. 
(As much as it pains me to recommend throwing away all of my enum code changes!)

iamigo: It pains me, too, but I agree.
So the values would be designated by each bot, right? So they're not "wide open" to end-users to modify?

ConalCosgrove: Exactly, they're hardcoded into places like GUS and refocus where users create rooms from, we don't have any free text fields where end users specify their own origins.

iamigo:
What's the status of this?

ConalCosgrove:
@iamigo was the change back to 'free-text' what was decided? If so will update PR.

iamigo:
yeah, go for it

db/model/room.js
```diff
@@ -51,6 +51,14 @@ module.exports = function room(seq, dataTypes) {
       allowNull: true,
       comment: 'Key/Value pairs for user specific settings',
     },
+    origin: {
+      type: dataTypes.STRING,
+      defaultValue: 'other',
+      validate: {
+        is: constants.nameRegex,
+      },
+      comment: 'Tracks where IMC room was created from',
```
kriscfoster: @ConalCosgrove should be Refocus Room I think

migrations/20190128134129-add-room-origin-column.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
kriscfoster: @iamigo I'm just curious but will this migration run if this is deployed now even though the date time string is back from January?

kriscfoster:
@ConalCosgrove is this ready for @iamigo to look at again?

ConalCosgrove:
@kriscfoster yep, good to go @iamigo

migrations/20190128134129-add-room-origin-column.js
iamigo: I *think* so, because sequelize maintains a list of migrations which it has executed in the SequelizeMeta table, no timestamps in that table, just the migration file names.

BUT it would be great to vet this in staging or devbox

db/model/room.js
```diff
@@ -51,6 +51,14 @@ module.exports = function room(seq, dataTypes) {
       allowNull: true,
       comment: 'Key/Value pairs for user specific settings',
     },
+    origin: {
+      type: dataTypes.STRING,
+      defaultValue: 'other',
+      validate: {
+        is: constants.nameRegex,
+      },
+      comment: 'Tracks where refocus room was created from',
```
iamigo: why is this comment different from swagger description of same field? I like this one

package.json
```diff
@@ -44,7 +44,7 @@
     "start-worker": "node worker/index.js",
     "test": "npm run reset-coverage && npm run test-all && npm run report",
     "test-all": "npm run test-main && npm run test-extra && npm run test-collector",
-    "test-api": "npm run test-with-coverage -- --recursive tests/api",
+    "test-api": "npm run test-with-coverage -- --recursive tests/api/",
```
iamigo: why is this line changing?
did you intend to update package-lock.json in this PR as well?

iamigo:
@ConalCosgrove resolve conflicts?

package.json
ConalCosgrove: reverted this line as well as package-lock.json

db/model/room.js
ConalCosgrove: Updated swagger to be consistent with db/model/room.js

cassiodias:
@iamigo can we merge that one - do you feel that is all done?

PR #1096, opened 2019-01-28
Aspect status ranges migration

pallavi2209:
If valueType is BOOLEAN or PERCENT and the current ranges wouldn't validate with new rules, change the valueType to NUMERIC.

How I tested this?
1. Backup prod data
2. Copied prod data to localhost
3. Ran migration
Results:
_48 aspects -> critical range was greater than MAX NUMBER, so changed critical range to MAX NUMBER
152 aspects -> changed valueType to NUMERIC_
4. undo migration -> nothings happens here
5. migrate again -> no change

Note: For a smooth release, when this goes to production, lets try and deploy in sandbox first. If everything looks good, then deploy in production.

### DIFF
--- a/migrations/20190126123621-update-aspect-rows-valueType.js
+++ b/migrations/20190126123621-update-aspect-rows-valueType.js
@@ -28,16 +28,27 @@ module.exports = {
               statusRangeMap.set('okRange', asp.okRange);
               const updateObj = {};
               for (const [key, value] of statusRangeMap.entries()) {
-                if (value && value[0] < Number.MIN_SAFE_INTEGER) {
-                  updateObj[key] = [Number.MIN_SAFE_INTEGER, value[1]];
-                }
+                let minVal = value[0];
+                let maxVal = value[1];
 
-                if (value && value[1] > Number.MAX_SAFE_INTEGER) {
-                  updateObj[key] = [value[0], Number.MAX_SAFE_INTEGER];
+                if (value) {
+                  if (value[0] < Number.MIN_SAFE_INTEGER ||
+                    value[0] > Number.MAX_SAFE_INTEGER) {
+                    minVal = Number.MIN_SAFE_INTEGER;
+                  }
+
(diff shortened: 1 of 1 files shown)


migrations/20190126123621-update-aspect-rows-valueType.js
```diff
@@ -0,0 +1,63 @@
...
+              promisesArr.push(Aspect.update(
+                { valueType: 'NUMERIC' }, { where: { id: asp.id } }
+              ));
+            }
+          }
+        });
+
+        // waits for all promises to resolve or reject
+        return Promise.all(
+          promisesArr.map((p) => p.catch(() => undefined))
```
iamigo: what is the purpose of "undefined" here?

migrations/20190126123621-update-aspect-rows-valueType.js
```diff
@@ -0,0 +1,63 @@
...
+            // if numeric, the only possible exception will be invalid range
+            if (asp.valueType === 'NUMERIC') {
+              const statusRangeMap = new Map();
+              statusRangeMap.set('criticalRange', asp.criticalRange);
+              statusRangeMap.set('warningRange', asp.warningRange);
+              statusRangeMap.set('infoRange', asp.infoRange);
+              statusRangeMap.set('okRange', asp.okRange);
+              const updateObj = {};
+              for (const [key, value] of statusRangeMap.entries()) {
+                if (value && value[0] < Number.MIN_SAFE_INTEGER) {
```
iamigo: isn't it theoretically possible to have both the start and end values be less than the min safe int? if that was the case, we'd still have an invalid range, because we're just correcting the start value here
same if both start and end were greater than max safe int value

migrations/20190126123621-update-aspect-rows-valueType.js
pallavi2209: Nothing - we just want to skip the promise which may reject and move on to next promise (instead of returning from Promise.all - which is default behaviour)

PR #1100, opened 2019-01-29
remove sequelize wrapper when calling publishObject

iamigo:
from subject model and perspective model


pallavi2209:
Travis failing?

PR #1101, opened 2019-01-30
Migration null check

pallavi2209:


pallavi2209:
ok @iamigo - that solves it - I deployed and checked in devbox. Lets deploy on sandbox again after merging, and then to prod.

PR #1102, opened 2019-01-31
PUT /v1/collectorGroups

iamigo:


tests/api/v1/collectorGroups/put.js
```diff
@@ -0,0 +1,181 @@
...
+      .end((err, res) => {
+        if (err) {
+          return done(err);
+        }
+
+        expect(res.body).to.have.property('description', 'new desc');
+        expect(res.body.collectors).to.be.empty;
+        return done();
+      });
+  });
```
jgraff2: Should also test the case where there are already collectors set.

iamigo: oh yeah - i think put and patch should both be *replacing* whatever is there, so should not fail if it is sending a collector which is already assigned to it (but *should* fail if a collector it is assigning is already assigned to a *different* collector group)

iamigo:
changed/added test per jon's comments @jgraff2 @pallavi2209

PR #1105, opened 2019-02-04
Added actionType field to Event

kriscfoster:
**_api/v1/swagger.yaml_**
- Updated EventResponse
- Updated prams for /events routes

**_db/model/event.js_**
- Added actionType field

**_migrations/20190204134143-event-add-type.js_**
- Migration for actionType field

Tests


api/v1/swagger.yaml
```diff
@@ -3829,6 +3829,9 @@ paths:
               context:
                 type: object
                 description: Any object, typically JSON, used to describe to some infomration that needed to be logged.
+              type:
```
iamigo: what about adding type to the list of attributes you can filter from query params, e.g. ...&type=foo

db/model/event.js
```diff
@@ -33,6 +34,11 @@ module.exports = function event(seq, dataTypes) {
       allowNull: false,
       comment: 'This is a readable event logline',
     },
+    type: {
+      type: dataTypes.STRING(constants.fieldlen.normalName),
```
iamigo: free text here is recipe for chaos! what are the values you actually expect to see here? would it be possible to make it an enum with "Other" as one of the options? then add options as needed down the line? it's not going to be infinite! i'm sure you can come up with a half dozen values or fewer

kriscfoster: In this case, I think free text is actually the best option for us. The "type" is defined by the bot developer. e.g. We will have a type values of: ["sendEmailTCE", "uploadFileToCase", "pageTeam", "createSlackChannel", ...] .. If a new bot developer wants to create events of their own unique type, they should be able to do that without needing to make changes to the platform

api/v1/swagger.yaml
kriscfoster: Thanks Ian, good catch, done

db/model/event.js
iamigo: I agree with you, but I also still think it's a recipe for chaos... oh well

api/v1/swagger.yaml
```diff
@@ -3889,6 +3898,9 @@ paths:
                 context:
                   type: object
                   description: Any object, typically JSON, used to describe to some infomration that needed to be logged.
```
BrandonMurnane: also here information

PR #1106, opened 2019-02-05
GET api for collectorGroups

pallavi2209:


tests/api/v1/collectorGroups/get.js
```diff
@@ -0,0 +1,189 @@
...
+      api.get(path + '/' + collectorGrp.id)
+        .set('Authorization', token)
+        .expect(constants.httpStatus.OK)
+        .expect((res) => {
+          expect(res.body.id).to.equal(collectorGrp.id);
+          expect(res.body.collectors.length).to.equal(2);
+
+          /*
+            Response has collector id, name, status and lastHeartbeat
+            Returned collector list is not in particular order
```
pallavi2209: So, I expected the returned collector list to be ordered by name (we already have order: ['name'] in the included collector scope). But for some reason, its not always ordered. I tried to figure this out for some time, but could not maintain the order. Any ideas? Maybe another sequelize problem :(

PR #1107, opened 2019-02-06
Add owner attribute to models

jgraff2:
This change adds an owner attribute to the following models:
  - aspects
  - botActions
  - botData
  - bots
  - collectorGroups
  - collectors
  - events
  - generators
  - generatorTemplates
  - lenses
  - perspectives
  - rooms
  - roomTypes
  - subjects

Things this change doesn't cover:
  - functional effects. Anything that's currently based on the user/createdBy will still be that way. This should be addressed in the spike for restricting permissions.
  - Reconciling with the existing User associations (with foreign key createdBy/installedBy/userId). Having both a user and an owner might be redundant in some cases, and it might make sense to change the name of the existing "user" field to "createdBy" and standardize the foreign key names; this change doesn't address that, it just creates a new owner attribute and leaves the existing ones as they are. We can clean this up in another story if necessary.
- The following models do not have an owner attribute. (Let me know if you think any of these should, or any of the above shouldn't)
  - auditEvent
  - globalConfig
  - profile
  - sample
  - ssoconfig
  - token
  - user

Changes are organized by commit.

### DIFF
--- a/api/v1/controllers/bots.js
+++ b/api/v1/controllers/bots.js
@@ -144,7 +144,9 @@ module.exports = {
 
       const user = req.user;
       seqObj.installedBy = user.id;
-      helper.model.create(seqObj)
+      u.setOwner(seqObj, req)
+        .then(() => helper.model.create(seqObj))
+        .then((o) => o.reload())
         .then((o) => {
           o.dataValues.ui = uiObj;
           if (featureToggles.isFeatureEnabled('addIsBotToToken')) {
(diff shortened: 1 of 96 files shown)


iamigo:
> change the name of the existing "user" field to "createdBy" and standardize the foreign key names

Yes, we should do that--please create a story for that--but we should discuss whether this will finally require us to rev the API to v2.

iamigo:
FYI @cassiodias model and API changes to all models

pallavi2209:
@jgraff2 quick confirmation - were you able to test the migration and related changes with production data?
I usually restore prod data on my localhost and run migratedb and undo it (and repeat) to test.

jgraff2:
@pallavi2209 I tested it locally for all models with sample data I created. I didn't use production data.

PR #1108, opened 2019-02-08
Handle null subject tags and relatedLinks

pallavi2209:
**What was the problem?**
When doing PUT /subjects/{name} without tags, we were explicitly settings tags = null, and since our db is fine with tags being null (allowNull: true) - it will bypass setting it to default value.

**Resolution:** 
1. Change PUT logic to set tags to default value (provided in swagger) if tags are not present in request.
2. Since we do not have a use-case for tags to be null, we should enforce tags to be notNull

This problem applies to almost all the array columns in our models including samples (related links) When it came to migration, I realized it will be too much for one PR. So, I am taking **staggered** **approach**. This PR is only for subjects model. Next would be - **aspects, perspectives, generators, generatorTemplates**. And then final PR for **samples** (db is redis - so migration will not work to update data).

Note: There is a sortBy column on Subjects model as well - allowNull: true, default: ''. This needs to be further looked into, I have a feeling we might have use cases for this one to be as is. I will create another user story for this.

### DIFF
--- a/tests/api/v1/subjects/put.js
+++ b/tests/api/v1/subjects/put.js
@@ -656,12 +656,12 @@ describe('tests/api/v1/subjects/put.js >', () => {
               done(err);
             }
 
-            Subject.findOne({ where: { id: subjectId } })
-              .then((subj) => {
-                expect(subj.tags).to.eql([]);
-                expect(subj.relatedLinks).to.eql([]);
-                done();
-              });
+            Subject.findById(subjectId)
+            .then((subj) => {
+              expect(subj.tags).to.eql([]);
+              expect(subj.relatedLinks).to.eql([]);
+              done();
+            });
           });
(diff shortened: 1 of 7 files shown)


tests/api/v1/subjects/put.js
```diff
@@ -637,6 +637,33 @@ describe('tests/api/v1/subjects/put.js >', () => {
         done();
       });
     });
+
+    it('tags and related links not provided, should set to empty array',
+      (done) => {
```
jgraff2: Are we already testing this for all other endpoints?

pallavi2209: not explicitly checking tags and related links in other possible tests. Added now.

iamigo:
@pallavi2209 test failures?

PR #1110, opened 2019-02-11
fix tests/db/model/generator/methods.js flapper

iamigo:
tests/db/model/collector/methods.js > instanceMethods > reassignGenerators > reassignGenerators:
      AssertionError: expected '___CollectorthirdCollector' to equal '___CollectorsecondCollector'
      + expected - actual
      -___CollectorthirdCollector
      +___CollectorsecondCollector
  at Promise.resolve.then.then.then.then.spread (tests/db/model/collector/methods.js:296:49)

tests/db/model/generator/methods.js > assignToCollector > collectors specified, first choice available:
     TypeError: Cannot read property 'name' of null
      at generator1.updateWithCollectors.then (tests/db/model/generator/methods.js:89:44)

### DIFF
--- a/tests/db/model/collector/methods.js
+++ b/tests/db/model/collector/methods.js
@@ -214,8 +214,7 @@ describe('tests/db/model/collector/methods.js >', () => {
         collectorConfig.heartbeatIntervalMillis = interval;
         collectorConfig.heartbeatLatencyToleranceMillis = tolerance;
 
-        Collector.build({ lastHeartbeat })
-        .isAlive().should.be.true;
+        Collector.build({ lastHeartbeat }).isAlive().should.be.true;
       });
 
       it('dead', () => {
@@ -227,8 +226,7 @@ describe('tests/db/model/collector/methods.js >', () => {
         collectorConfig.heartbeatIntervalMillis = interval;
         collectorConfig.heartbeatLatencyToleranceMillis = tolerance;
 
-        Collector.build({ lastHeartbeat })
-        .isAlive().should.be.false;
+        Collector.build({ lastHeartbeat }).isAlive().should.be.false;
       });
 
       it('undefined', () => {
@@ -240,8 +238,7 @@ describe('tests/db/model/collector/methods.js >', () => {
         collectorConfig.heartbeatIntervalMillis = interval;
         collectorConfig.heartbeatLatencyToleranceMillis = tolerance;
 
-        Collector.build({ lastHeartbeat })
-        .isAlive().should.be.false;
+        Collector.build({ lastHeartbeat }).isAlive().should.be.false;
       });
     });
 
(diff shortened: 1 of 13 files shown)


tests/db/model/generator/methods.js
```diff
@@ -85,9 +84,9 @@ describe('tests/db/model/generator/methods.js >', () => {
         isActive: true,
         possibleCollectors: [coll2.name, coll3.name],
       })
-      .then(() => {
-        expect(generator1.currentCollector.name).to.equal(collector2.name);
-        expect(generator1.currentCollector.id).to.equal(collector2.id);
```
jgraff2: Is this all just for readability?

tests/db/model/generator/methods.js
```diff
@@ -85,9 +84,9 @@ describe('tests/db/model/generator/methods.js >', () => {
...
+      .then((updated) => {
+        expect(updated.currentCollector.name).to.equal(collector2.name);
+        expect(updated.currentCollector.id).to.equal(collector2.id);
```
jgraff2: Is this all just for readability?

iamigo: The call to updateWithCollectors returns the updated and reloaded object, so we want to be verifying the values on that reloaded object. That was the only thing I found here that would have been responsible for the flapper, so I changed it on the flapping test and other obvious places we were doing the same thing.

tests/db/model/collector/methods.js
```diff
@@ -288,12 +285,12 @@ describe('tests/db/model/collector/methods.js >', () => {
...
-          Generator.find({ where: { name: generator1.name } }),
-          Generator.find({ where: { name: generator2.name } }),
-          Generator.find({ where: { name: generator3.name } }),
+          Generator.findOne({ where: { name: generator1.name } }),
+          Generator.findOne({ where: { name: generator2.name } }),
+          Generator.findOne({ where: { name: generator3.name } }),
         ))
         .spread((gen1, gen2, gen3) => {
-          expect(gen1.currentCollector.name).to.equal(collector2.name);
+          expect(gen1.currentCollector.name).to.not.equal(collector1.name);
```
jgraff2: So the problem was that gens are being assigned inconsistently?
This could have been triggered by the change to balancing the assignment (these tests were written before that).  But I think that should still be deterministic. 
It could be that the test setup is creating things in parallel, and the order those operations execute in can determine where things get assigned down the line.

tests/db/model/generator/methods.js
jgraff2: Oh ok. Hmm... we already have a reference to the object, so I wouldn't think it would make a difference. Maybe a sequelize bug? In any case it's fine either way so it makes sense to try this.

tests/db/model/collector/methods.js
iamigo: thx i'll look into that some more

iamigo:
@jgraff2 updated the pr based on your comments - thx!

PR #1112, opened 2019-02-13
tags|relatedLinks|filters not null for Aspects|Generators|GeneratorTemplates|Perspectives|Samples

pallavi2209:
**Swagger changes**: Add default:[] to corresponding atributes
**Model changes**: Update columns, allowNull: false
**Tests**: Put|Patch|Post tests to check these attributes return empty array when not set.
**Migration**: To update rows with empty array if column is null, and update columns to disallow null.

### DIFF
--- a/migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
+++ b/migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
@@ -9,7 +9,6 @@ const modelsAndCols = {
   GeneratorTemplate: ['tags'],
   Perspective:
     ['aspectFilter', 'aspectTagFilter', 'subjectTagFilter', 'statusFilter'],
-  Sample: ['relatedLinks'],
 };
 
 function updateRowsAndColumns(qi, seq, modelName, colNamesArr) {
--- a/migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
+++ b/migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
@@ -21,7 +21,7 @@ function updateRowsAndColumns(qi, seq, modelName, colNamesArr) {
 
   const model = db[modelName];
   return model.findAll({ // find records with any of given columns = null
-    where: { $or: whereArr },
+    where: { $or: whereArr }, paranoid: false,
   })
     .then((records) => Promise.mapSeries(records, (record) => {
       const toUpdate = {};
--- a/migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
+++ b/migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
@@ -1,6 +1,7 @@
 'use strict';
 const db = require('../db/index');
 const constants = require('../db/constants');
+const Promise = require('bluebird');
 
 const modelsAndCols = {
   Aspect: ['tags', 'relatedLinks'],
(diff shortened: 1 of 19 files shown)


migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
```diff
@@ -0,0 +1,77 @@
...
+
+        return record.update(toUpdate);
+      })
+    ))
+    .then(() => Promise.all(colNamesArr.map((colName) => {
+      if (colName === 'relatedLinks') {
+        return qi.changeColumn(`${modelName}s`, colName, {
+          type: seq.ARRAY(seq.JSON),
+          allowNull: false, // change relatedLinks column to allowNull: false
+          defaultValue: constants.defaultJsonArrayValue,
```
jgraff2: Is it necessary to include the type and defaultValue even though they aren"t changing?

migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
```diff
@@ -0,0 +1,77 @@
...
+    const o = {};
+    o[cname] = { $eq: null };
+    whereArr.push(o);
+  });
+
+  const model = db[modelName];
+  return model.findAll({ // find records with any of given columns = null
+    where: { $or: whereArr },
+  })
+    .then((records) => Promise.all(
```
jgraff2: This might run into the same problem I did in the owner migration. You should use Promise.mapSeries to make sure it runs the queries one-at-a-time.

migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
pallavi2209: For changeColumn, the sequelize docs says:

> This method changes the meta data of an attribute. It is possible to change the default value, allowance of null or the data type. Please make sure, that you are completely describing the new data type.

migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
```diff
@@ -0,0 +1,76 @@
...
+const constants = require('../db/constants');
+const Promise = require('bluebird');
+
+const modelsAndCols = {
+  Aspect: ['tags', 'relatedLinks'],
+  Generator: ['tags'],
+  GeneratorTemplate: ['tags'],
+  Perspective:
+    ['aspectFilter', 'aspectTagFilter', 'subjectTagFilter', 'statusFilter'],
+  Sample: ['relatedLinks'],
```
iamigo: shouldn't have to update Sample table - we don't use it

migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
```diff
@@ -0,0 +1,75 @@
...
+      const toUpdate = {};
+      colNamesArr.forEach((colName) => {
+        if (record[colName] === null) {
+          toUpdate[colName] = []; // set the record column to empty array
+        }
+      });
+
+      return record.update(toUpdate);
+    }))
+    .then(() => Promise.all(colNamesArr.map((colName) => {
```
jgraff2: Should also use mapSeries here.

migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
```diff
@@ -0,0 +1,75 @@
...
+      return qi.changeColumn(`${modelName}s`, colName, {
+        type: seq.ARRAY(seq.STRING(constants.fieldlen.normalName)),
+        allowNull: false, // change tags column to allowNull: false
+        defaultValue: constants.defaultArrayValue,
+      });
+    })));
+}
+
+module.exports = {
+  up: (qi, seq) => Promise.all(Object.keys(modelsAndCols).map((model) =>
```
jgraff2: and here

migrations/20190212004845-tags-rlinks-not-null-multiple-models.js
```diff
@@ -0,0 +1,75 @@
...
+    updateRowsAndColumns(qi, seq, model, modelsAndCols[model]))),
+
+  down: (qi, seq) => Promise.all(Object.keys(modelsAndCols).map((modelName) => {
+    const colNamesArr = modelsAndCols[modelName];
+    return Promise.all(colNamesArr.map((colName) => {
```
jgraff2: and here

pallavi2209: Should we prefer mapSeries over all if we know that there will be limited number of promises, say 24 in this case?

jgraff2: You're right, it shouldn't be necessary for this one.

PR #1114, opened 2019-02-17
Multiple pubsub publishers

iamigo:


tests/realtime/pubSub.js
```diff
@@ -0,0 +1,121 @@
...
+const Subject = tu.db.Subject;
+const Aspect = tu.db.Aspect;
+const Sample = tu.Sample;
+const redisPublisher = require('../../realtime/redisPublisher');
+const event = require('../../realtime/constants').events.sample;
+const rtu = require('../cache/models/redisTestUtil');
+const subClients = require('../../cache/redisCache').client.subPerspectives;
+
+describe('tests/realtime/pubsub.js >', () => {
+  describe('publish and subscribe>', () => {
```
jgraff2: " >"

tests/realtime/pubSub.js
```diff
@@ -0,0 +1,121 @@
...
+      function waitForMessages(time) {
+        // check for messages when we get all of them and then return
+        if (receivedMsgs.length === publishCount) {
+          if (subClients.length === 1) {
+            expect(subMsgCount[0]).to.equal(publishCount);
+          } else if (subClients.length > 1) {
+            let countSum = 0;
+            subMsgCount.forEach((count) => {
+              // messages should be distributed among subscribers
+              expect(count).to.be.below(publishCount);
```
jgraff2: I guess this is the best we can do here since it's random.
Did you verify manually that it's distributed fairly evenly across all the subscribers?
Maybe we could also check that every subscriber got at least one event?

pallavi2209: "Maybe we could also check that every subscriber got at least one event?" - I agree - let me update the PR
I did test manually with different runs - It was distributed mostly evenly like, 20-30, 27-23 etc

pallavi2209: So, I looked, Math.random generators number with **approximately** uniform distribution - I don't like the word **approximately**, so let me update to remove "at least one event" check.

iamigo:
LGTM

PR #1115, opened 2019-02-19
added migration and test

BrandonMurnane:
Added field to botaction for logging purposes to allow extra granularity to be set


kriscfoster:
@BrandonMurnane tests, think it's just some styling stuff

PR #1121, opened 2019-02-22
For pubsub stats logging, aggregate the data in memory

iamigo:
in a global variable instead of in redis.

Because it's in memory (inside each process), we can't use clock dyno
to emit the log lines at regular intervals, so this adds a timer in
each web dyno's node process and in each worker dyno to emit the log
lines.


tests/realtime/pubSubStats.js
```diff
@@ -0,0 +1,212 @@
...
+        expect(() => pss.track('pub', 'hello.world')).to.throw();
+        ;
+        expect(() => pss.track('pub', 'hello.world', null)).to.throw();
+        ;
+        expect(() => pss.track('pub', 'hello.world', undefined)).to.throw();
+        ;
+        expect(() => pss.track('pub', 'hello.world', [])).to.throw();
+        ;
+        expect(() => pss.track('pub', 'hello.world', '')).to.throw();
+        ;
```
jgraff2: ?

PR #1122, opened 2019-03-01
Collector group assignment

jgraff2:
Use collector groups for generator assignment.
Includes model changes required to make it work correctly.
Includes removing the existing possibleCollectors association, with migration.
Also includes a fix for the issue with foreign keys being included in the response when associations are present.
Organized by commit.


api/v1/helpers/nouns/perspectives.js
```diff
@@ -64,6 +64,7 @@ module.exports = {
     owner: 'owner',
     lens: 'lens',
   },
+  fieldsToExclude: ['createdBy', 'ownerId', 'lensId'],
```
kriscfoster: Not really related to this PR but I'm curious.. Are we depreciating the createdBy field and using owner instead, or is there a difference between these?

tests/api/v1/collectorGroups/create.js
```diff
@@ -191,4 +191,42 @@ describe('tests/api/v1/collectorGroups/create.js >', () => {
...
+      .expect(httpStatus.OK)
+      .end((err, res) => {
+        if (err) {
+          return done(err);
+        }
+
+        api.post('/v1/collectorGroups')
+        .set('Authorization', token)
+        .send({ // Creating a new Collector Group with one Coll assigned
+          name: 'coll-group-must-fail',
```
pallavi2209: misleading name and description?

tests/api/v1/collectorGroups/put.js
```diff
@@ -203,4 +236,254 @@ describe('tests/api/v1/collectorGroups/put.js >', () => {
...
+              expect(res.body.currentCollector.name).to.equal(collectorAlive1.name);
+              return done();
+            });
+          });
+        }).catch(done);
+      });
+
+    it('already assigned generator, put collectorGroup should set ' +
+      'currentCollector to null if currentCollector does not exist in ' +
+      'updated collector list', (done) => {
```
pallavi2209: How about another similar test but not including collectors attribute in collectorGroups put request?
```
        api.put(`/v1/collectorGroups/${cg.name}`)
        .set('Authorization', token)
        .send({
          name: cg.name,
        })
```

api/v1/helpers/nouns/perspectives.js
iamigo: We are not deprecating `createdBy`.
`createdBy` is the user who created the record, but we are introducing a new concept of "owner" because the user who creates a record can come and go, change roles, etc., and we need the ability to reassign "ownership" of the record.

jgraff2:
Makes sense. I moved the migration to a separate PR: https://github.com/salesforce/refocus/pull/1125.

tests/api/v1/collectorGroups/put.js
jgraff2: added.

tests/api/v1/collectorGroups/create.js
jgraff2: changed.

PR #1123, opened 2019-03-04
Added event log

BrandonMurnane:


kriscfoster:
@BrandonMurnane is this ready for reviewers?

api/v1/controllers/events.js
```diff
@@ -78,7 +81,40 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   postEvents(req, res, next) {
-    doPost(req, res, next, helper);
+    // doPost(req, res, next, helper);
+    const resultObj = { reqStartTime: req.timestamp };
+    const reqObj = req.swagger.params.queryBody.value;
+    const user = req.user;
+    reqObj.installedBy = user.id;
```
kriscfoster: Do we need to add installedBy as a field to the request object here?

config/activityLog.js
```diff
@@ -54,6 +54,23 @@ module.exports = {
...
+  event: {
+    activity: 'event',
+    action_id: 'None',
+    actionSuccess: 'None',
+    botName: 'None',
+    ipAddress: 'None',
+    dbTime: '0ms',
+    method: 'None', // one of HTTP verbs
+    process: 'None',
+    request_id: 'None',
```
iamigo: "request_id" is lowercase and has underscore because it is "inherited" from heroku.

If there is no good reason for "action_id" to be "action_id" then I would suggest it should be camel-case "actionId"

api/v1/controllers/events.js
```diff
@@ -78,7 +81,39 @@ module.exports = {
    * @param {Function} next - The next middleware function in the stack
    */
   postEvents(req, res, next) {
-    doPost(req, res, next, helper);
+    // doPost(req, res, next, helper);
```
kriscfoster: @BrandonMurnane Do we need this comment?

PR #1129, opened 2019-03-07
attachSamplesPrefilterAspects

iamigo:
If there is an aspect name filter, do it earlier in attachSamples instead of doing extra hgetall calls for aspects and samples we don't need!
Add short-term toggle "attachSamplesPrefilterAspects" (env var "ATTACH_SAMPLES_PREFILTER_ASPECTS")

### DIFF
--- a/cache/models/subject.js
+++ b/cache/models/subject.js
@@ -10,7 +10,7 @@
  * cache/model/subject.js
  */
 'use strict'; // eslint-disable-line strict
-
+const featureToggles = require('feature-toggles');
 const helper = require('../../api/v1/helpers/nouns/subjects');
 const fu = require('../../api/v1/helpers/verbs/findUtils.js');
 const utils = require('../../api/v1/helpers/verbs/utils');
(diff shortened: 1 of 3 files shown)


cache/models/subject.js
```diff
@@ -99,11 +99,33 @@ function attachSamples(res) {
...
-      cmds.push(['hgetall', sampleKey]);
-      cmds.push(['hgetall', aspectKey]);
+
+      if (featureToggles.isFeatureEnabled('attachSamplesPrefilterAspects')) {
+        /*
+         * If there is an aspect name filter, do it now instead of doing extra
+         * hgetall calls for aspects and samples we don't need!
+         */
+        let aspectPassesFilter = true; // default true if no aspect name filter
+        if (filters.aspect.includes && filters.aspect.includes.size) {
```
cassiodias: wouldn't underscore lib help with undefined chain (_.isUndefined)? or something like that... I remember somewhere we were using a lib for that.

iamigo: There's just one usage of `lodash` outside of views (in bots db model, using "extend"), but we *do* use `just-safe-get` in some places instead of having to check for existence of many nested layers of attributes... here it wasn't so convoluted that I felt it needed a semantically cleaner way of expressing it

PR #1130, opened 2019-03-08
GET api endpoints - createdAt/updatedAt filter

pallavi2209:
Organized by commits.

e.g. ?createdAt=-7d or ?updatedAt=-12h
All times relative to "now"
Supported these units: second (s), minute (m), hour (h), day (d).
Invalid request if no minus sign or if no number or if no unit.

Tests cover following endpoints:
GET /v1/aspects
GET /v1/auditEvents
GET /v1/collectorGroups
GET /v1/collectors
GET /v1/events
GET /v1/generators
GET /v1/generatorTemplates
GET /v1/perspectives
GET /v1/rooms
GET /v1/roomTypes
GET /v1/botActions?botId=<>
GET /v1/botData?botId=<>
GET /v1/bots
GET /v1/globalconfig
GET /v1/profiles
GET /v1/users

### DIFF
--- a/api/v1/helpers/verbs/findUtils.js
+++ b/api/v1/helpers/verbs/findUtils.js
@@ -19,6 +19,7 @@ const ONE = 1;
 const MINUS_ONE = -1;
 const RADIX = 10;
 const Op = require('sequelize').Op;
+const ms = require('ms');
 
 /**
  * Escapes all percent literals so they're not treated as wildcards.
@@ -156,6 +157,7 @@ function toSequelizeWhere(filter, props) {
   for (let i = ZERO; i < keys.length; i++) {
     const key = keys[i];
     if (filter[key] !== undefined) {
+      const filterKeyValue = filter[key];
       if (!Array.isArray(filter[key])) {
         filter[key] = [filter[key]];
       }
@@ -196,6 +198,12 @@ function toSequelizeWhere(filter, props) {
       else if (props.tagFilterName && key === props.tagFilterName) {
         const tagArr = filter[key];
         Object.assign(where, toWhereClause(tagArr, props));
+      } else if (props.timePeriodFilters &&
+        props.timePeriodFilters.includes(key)) {
+        const whereClause = {};
+        whereClause[Op.lte] = new Date();
+        whereClause[Op.gte] = new Date(Date.now() + ms(filterKeyValue));
+        where[key] = whereClause;
       } else {
         for (let j = ZERO; j < filter[key].length; j++) {
           const v = filter[key][j];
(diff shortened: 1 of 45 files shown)


api/v1/helpers/verbs/findUtils.js
```diff
@@ -196,6 +198,12 @@ function toSequelizeWhere(filter, props) {
       else if (props.tagFilterName && key === props.tagFilterName) {
         const tagArr = filter[key];
         Object.assign(where, toWhereClause(tagArr, props));
+      } else if (props.timePeriodFilters &&
+        props.timePeriodFilters.includes(key)) {
+        const whereClause = {};
+        whereClause[Op.lte] = new Date();
```
iamigo: is this "earlier than now" part of the where clause necessary, since every date will be earlier than now?

pallavi2209: Yeah, this was required for testing, because I am mocking the now time to be something in past  and some things like admin user is created at real now. I added this later on as a part of my iterative workarounds to get the tests working.

tests/api/v1/common/createdAtUpdatedAtFilters.js
```diff
@@ -0,0 +1,460 @@
...
+        modelName, nameAttr, util));
+    after((done) => util.forceDelete(done, testStartTime));
+
+    describe('GET, createdAt for specific time period >', () => {
+      it('createdAt=-5 seconds', () => {
+        const filterString = 'createdAt=-5s';
+        const expectedStatus = constants.httpStatus.OK;
+        return getResources({ modelName, filterString, expectedStatus })
+          .then((res) => {
+            expect(createdResources.length).equals(4);
```
jgraff2: Isn't this the same in every test?

api/v1/helpers/verbs/findUtils.js
```diff
@@ -196,6 +198,12 @@ function toSequelizeWhere(filter, props) {
       else if (props.tagFilterName && key === props.tagFilterName) {
         const tagArr = filter[key];
         Object.assign(where, toWhereClause(tagArr, props));
+      } else if (props.timePeriodFilters &&
```
jgraff2: Does doing this in an else-if mean we can't filter on both createdAt and tags?

pallavi2209: This is inside a for loop, so we are creating a where clause for each key at a time. I should probably add a test in one of the model to specifically test this.

jgraff2: ah ok.

PR #1131, opened 2019-03-11
lensId fix

jgraff2:
As part of the recent collectorGroups change, I removed foreign keys from the api response and changed existing references to assoc.id. I missed this one.


view/perspective/utils.js
```diff
@@ -280,7 +280,7 @@ function getValuesObject(accumulatorObject) {
      * lensLoadEvent is dispatched. Since hierarchyLoadEvent is truthy,
      * it is also dispatched.
      */
-    const getLens = getPromiseWithUrl('/v1/lenses/' + perspective.lensId)
+    const getLens = getPromiseWithUrl('/v1/lenses/' + perspective.lens.id)
```
pallavi2209: Do we not have test for this?

pallavi2209: aah never mind - did not notice that this is view changes

PR #1133, opened 2019-03-13
Token based authentication to join socket server

pallavi2209:
**Previously:**
We had only cookie based authentication to join socket server (we had token-based authentication for Bots though). 

**After these changes:**
Added token based authentication for the socket client to connect to socket server for user tokens

**Impact:**
This will make it easier for us to test socket communication using supertest.

**Tests:**
Added some tests.
Also tested manually for realtime events

### DIFF
--- a/utils/jwtUtil.js
+++ b/utils/jwtUtil.js
@@ -180,7 +180,7 @@ function verifySocketToken(token) {
           method: ['notRevoked', payload.tokenname, user.id],
         }).findOne());
     });
-} // verifyBotSocketToken
+} // verifySocketToken
 
 /**
  * Function to verify if a bot token is valid or not, i.e. does the token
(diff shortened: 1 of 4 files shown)


utils/jwtUtil.js
```diff
@@ -197,7 +207,8 @@ function verifyBotToken(payload, req, cb) {
     assignHeaderValues(req, payload);
     return cb ? cb() : undefined;
   })
-  .catch(() => {
+  .catch((err) => {
+    console.log(err);
```
iamigo: did you intend to leave this in?

utils/jwtUtil.js
```diff
@@ -163,13 +163,23 @@ function verifyGeneratorToken(payload, req, cb) {
...
+        return Bot.findOne({ where: { name: payload.tokenname } });
+      }
+
+      const username = payload.username;
+      return User.findOne({ name: username })
+        .then((user) => Token.scope({
+          method: ['notRevoked', payload.tokenname, user.id],
+        }).findOne());
+    });
 } // verifyBotSocketToken
```
iamigo: verifySocketToken

utils/jwtUtil.js
pallavi2209: Nope I did not! Thanks for pointing out!

PR #1134, opened 2019-03-13
LensId is not longer exposed in perspective.

froismo:
Use `lens.id` instead.


pallavi2209:
@froismo Thanks for the PR, but I think there is already a PR for this. Check out #1131

froismo:
Closing PR because it's duplicate of https://github.com/salesforce/refocus/pull/1131

PR #1136, opened 2019-03-15
Reassign generators on missed upsert

jgraff2:
A note on the method of disabling generators for testing:
I'm doing this by mocking the heartbeat response in express middleware, telling the collector to delete and add generators in a way that's transparent to refocus.
This is not ideal, and requires the code to be overly complicated, but it's the only way I was able to get it to work.
Other things I tried:
- intercept upsert with nock
  - nock is all-or-nothing for each url - you can set filters on the request to determine the response, but that doesn't affect whether the request goes through or not. And since upserts all go to the same url, there's no way to block one generator but let the other through
- mock heartbeat assignment with nock
  - the above problem doesn't apply since each collector has a different url, but it prevents refocus from getting the heartbeat and causes it to be counted as missed.
- intercept upsert at controller level as part of existing interception, set up at the beginning of the tests
  - caused conflicts with existing interception due to semmingly incorrect promise behavior. Still not sure exactly what was wrong, and this should be possible, but I gave up after spending too much time on it.
- intercept upsert at controller level dynamically
  - mocking controller methods with sinon only works if done outside of the test hooks; it can't be enabled and disabled within a test
- intercept upsert in express middleware
  - can only add middleware at the end of the chain, so can't prevent the upsert from going to the controller

Corresponding collector changes: https://github.com/salesforce/refocus-collector/pull/181
Tests won't pass until that PR is released to npm.


tests/collector/integration.js
```diff
@@ -478,18 +481,67 @@ describe('tests/collector/integration.js >', function () {
...
+        .then(() => forkUtils.tickFor(ms('3m')))
+        .then(() => u.resumeGenerator(gen1, coll1))
+        .then(() => Promise.join(
+          u.awaitBulkUpsert(coll1)
+            .should.eventually.be.rejectedWith(Promise.TimeoutError),
+          u.expectBulkUpsertSamples(coll2, '60s', ['sub1|asp1']),
+        ))
+
+        .then(() => {
+          config.collector.generatorUpsertToleranceFactor = 3;
```
pallavi2209: maybe set generatorUpsertToleranceFactor to something which is not default?

config/activityLog.js
```diff
@@ -124,6 +124,13 @@ module.exports = {
     lastHeartbeat: 'None',
     delta: 'None',
   },
+  missedActivity: {
```
BrandonMurnane: should this log type be on by default it seemed like the more common way is to put it behind the logs toggle

jgraff2: It's also on by default for the missed heartbeat logging - the idea is that this isn't something that should happen regularly so when it does we want to know about it.

iamigo: is there maybe a more precise word we can use here to refer to what was missed? "activity" seems vague

config/collectorConfig.js
```diff
@@ -71,6 +74,17 @@ module.exports = {
    */
   heartbeatLatencyToleranceMillis,
 
+  /*
+   * The number of upserts a generator can miss before it will be assumed to be
+   * stuck and reassigned to a different collector.
+   */
+  generatorUpsertToleranceFactor,
```
iamigo: Are these new config items intended to be sent to the running refocus collectors? I think the config items in this file are config items which can be defined or overridden by Refocus but are sent to the collectors to alter the behavior of the collectors

config/activityLog.js
```diff
@@ -124,6 +124,13 @@ module.exports = {
...
+    activity: 'missedActivity',
+    generator: 'None',
+    currentCollector: 'None',
```
iamigo: I think in the activity log line we can get away with calling it just "collector" rather than "currentCollector"

config.js
```diff
@@ -109,6 +109,17 @@ const JOB_QUEUE_TTL_SECONDS_SYNC = pe.TTL_KUE_JOBS_SYNC
 // sent to collector.
 const encryptionAlgoForCollector = 'aes-256-cbc';
 
+/*
+ * The number of upserts a generator can miss before it will be assumed to be
+ * stuck and reassigned to a different collector.
+ */
+let generatorUpsertToleranceFactor = +pe.GENERATOR_UPSERT_TOLERANCE_FACTOR || 3;
```
iamigo: is this really a "let"? where would we be reassigning a value to this variable?

config/activityLog.js
BrandonMurnane: cool thanks for clearing that up

db/model/generator.js
```diff
@@ -455,6 +488,27 @@ module.exports = function generator(seq, dataTypes) {
...
+
+    const expectedInterval = this.intervalSecs * 1000;
+    const toleranceFactor = conf.generatorUpsertToleranceFactor;
+    const retries = conf.generatorMissedUpsertRetries;
+    const heartbeatInterval = collectorConfig.heartbeatIntervalMillis;
+    const timeSinceUpsert = Date.now() - this.lastUpsert.getTime();
+    const upsertTolerance = expectedInterval * toleranceFactor;
+
+    const delta = timeSinceUpsert - upsertTolerance;
+    const retryCap = heartbeatInterval * retries;
```
pallavi2209: Comment for this please.

PR #1139, opened 2019-03-20
don't run clock jobs in tests

jgraff2:
Should fix generator currentCollector flappers.


jgraff2:
Taking a different approach to this after discussing with Ian and thinking more about it.
I extracted the express setup code to a new "express.js" file, and updated the tests to require that directly. This way, the tests only run the express app, and the other stuff (clock, sample store, logging) is only run when the full app is actually being run.

express.js
```diff
@@ -0,0 +1,280 @@
...
+const rejectMultipleXForwardedFor =
+  require('./config/rejectMultipleXForwardedFor');
+const bodyParser = require('body-parser');
+const env = conf.environment[conf.nodeEnv];
+const ENCODING = 'utf8';
+const compress = require('compression');
+const cors = require('cors');
+const etag = require('etag');
+
+// set up server side socket.io and redis publisher
```
iamigo: `redis publisher`? not being set up here anymore, right?

iamigo: Also maybe just move this comment up to line 12

index.js
```diff
@@ -34,340 +32,72 @@ function start(clusterProcessId = 0) { // eslint-disable-line max-statements
...
   /*
-   * Call this *before* the static pages and the API routes so that both the
-   * static pages *and* the API responses are compressed (gzip).
+   * Clock jobs
+   * If the clock dyno is NOT enabled, schedule all the scheduled jobs right
+   * from here.
    */
-  app.use(compress());
+  if (!featureToggles.isFeatureEnabled('enableClockProcess')) {
+    require('./clock/index'); // eslint-disable-line global-require
```
iamigo: Does this even work? If I'm starting up the clock INSIDE a web dyno, won't I have multiple clocks if I have multiple dynos, and multiple processes per dyno?
I think we should no longer support running the clock inside the main web process.

config/toggles.js
```diff
@@ -111,9 +111,6 @@ const longTermToggles = {
   enableEventActivityLogs: envVarIncludes(pe, 'ENABLE_ACTIVITY_LOGS',
     'event'),
 
-  // Enable heroku clock dyno
-  enableClockProcess: environmentVariableTrue(pe, 'ENABLE_CLOCK_PROCESS'),
```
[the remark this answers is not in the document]
jgraff2: oops, I meant to undo this change. Thanks for catching it.
This is from the first approach - the new approach doesn't require changing our current env var setup.

index.js
jgraff2: This is exactly how it was before.

Yes, it would create multiple clocks, which is why when running in heroku enableClockProcess is set and this line doesn't run.

This was the discussion we had last week - we decided it was important to support a minimum-effort deployment where you can run a single command to start a single process and have everything just work.

The way the environment variable works here is kind of strange, since we're relying on a variable that isn't actually required to run the separate clock process. I did think about trying to detect whether it was running standalone or with a separate clock process so we don't need an environment variable at all, but that didn't seem very promising. I decided it was best to stick with the way it works now.

jgraff2: If we wanted to clean this up, we could require that the clock process only runs if the env var is set, so there's no way to run it in both places. And maybe prevent it from running within the web process if there are multiple processes.
That would be a separate story - this PR is specifically for allowing the tests to be run without the clock jobs, it doesn't change anything about how it works when running the full app.

express.js
jgraff2: The sample store initialization was moved, the pub/sub is still here because it's dependent on the express app.

jgraff2: Though maybe it would make sense to extract that as well, and maybe also socket.io.
My initial approach was to keep everything that had a dependency on the express app in this file, which is why I kept these here. But then I had to break that rule and use the exported app object for the processId because it was only available in that context. So I could probably do the same thing for redis and socket.io, as long as the tests don't rely on them.

index.js
iamigo: Let's get a story out there to *only* run the clock jobs in a separate process.

iamigo:
@jgraff2 can you update the subject line of this pull request? "Don't run clock jobs by default" doesn't feel right.

index.js
jgraff2: "Let's get a story out there to only run the clock jobs in a separate process."

From my comment above:
"This was the discussion we had last week - we decided it was important to support a minimum-effort deployment where you can run a single command to start a single process and have everything just work."

Did something change? Do we need to discuss again?

iamigo: I changed my mind because our current implementation is flawed because it can allow multiple clocks to be running, which we would never want in a production environment, and because in dev mode a developer can just use `heroku local` (so there is no big huge burden of having to start multiple processes).

jgraff2: We could fix that without removing the ability to run in a separate process by tieing the clock process to the environment variable, and when running in the web process doing the same thing we do for env var logging - only run if process id is 0 or 1. 

I will create a story and we can discuss further at backlog grooming.

PR #1140, opened 2019-03-22
Log roomid

BrandonMurnane:


view/rooms/app.js
```diff
@@ -622,16 +623,17 @@ function userEnterRoom() {
...
+    isActive: true,
+  };
   const events = {
-      log: message,
-      context: eventType,
-      userId: _user.id,
-      roomId: parseInt(ROOM_ID, 10),
-    };
+    log: message,
+    actionType: eventType.type + 'Login',
```
kriscfoster: Just a minor thing but should this be named Join/Leave instead of Login/Logout, just a preference thing though

PR #1142, opened 2019-03-25
increase timeout to prevent clock test flapping

jgraff2:


jgraff2:
It's forking a new process (which itself takes about 1s), mocking the time moving forward by one hour, and for all the clock jobs that would be run throughout that hour: processing through kue, executing, and awaiting the result.
Running locally takes about 3.5s, so it's understandable how it could go above 5 sometimes in Travis.
It's also possible it's getting hung up on processing the worker jobs, but I want to try this first before looking into that.

PR #1143, opened 2019-03-28
testing event-flapper

BrandonMurnane:


iamigo:
Perhaps getting closer, @BrandonMurnane -- looks like just this now...

  1) tests/api/v1/events/get.js > "before each" hook for "Pass, get array of multiple":
     TypeError: Cannot read property 'then' of undefined
      at Context.beforeEach (tests/api/v1/events/get.js:55:20)

BrandonMurnane:
@iamigo is passing without any flappers re ran it a few times to be sure

iamigo:
@BrandonMurnane looking good
I think you might need to rebase (?) this so that it only contains *your* changes (and not changes from other people's recent commits)

BrandonMurnane:
@iamigo rebased

PR #1146, opened 2019-04-02
log collector name on upsert

jgraff2:
This will allow us to tell where upserts are coming from in sumo.
It would be very useful for the demo if we could get this released today.

### DIFF
--- a/config/activityLog.js
+++ b/config/activityLog.js
@@ -26,6 +26,7 @@ module.exports = {
     totalTime: '0ms',
     uri: 'None',
     user: 'None',
+    collector: 'None',
   },
   collectorAssignment: {
     activity: 'collectorAssignment',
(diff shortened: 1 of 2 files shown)


config/activityLog.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
     totalTime: '0ms',
     uri: 'None',
     user: 'None',
+    collector: 'None',
```
iamigo: does not seem like a great fit adding this to api log line since most api endpoints will not have a "collector" value... I will approve but I think this actually needs more discussion

iamigo: also, our convention is for the fields to be in alphabetical order after the "activity" field

jgraff2: I agree, but I think the usefulness of being able to see which collector upserts are coming from outweighs the slight inconvenience of having collector=None in the other log lines. The way our logging is set up there isn't any way around this; we would have to rethink the api logging logic (Is it necessary to log the default value? Could we just skip the field if it's not provided?)

Should I create a separate story to think about this more and clean it up?

PR #1152, opened 2019-04-16
Ip whitelist utils

iamigo:
If refocus-whitelist service defined (env var `IP_WHITELIST_SERVICE`), use that service's API to verify the caller's IP address against the configured whitelist. Used for both REST API requests *and* socket connections.

Otherwise, just use existing whiltelist behavior.


pallavi2209:
@iamigo Looks like this one has changes from ipAddressUtils branch as well - how about rebasing this PR on that branch?

iamigo:
@pallavi2209 yes, I was just going to wait until the first one was merged

utils/ipWhitelistUtils.js
```diff
@@ -0,0 +1,43 @@
...
+ * Resolves to true if the `addr` arg is a valid and whitelisted IP address.
+ *
+ * @param {String} addr - the IP address to test
+ * @returns {Promise<Boolean>} true if the `addr` arg is a valid and
+ *  whitelisted IP address
+ */
+function isWhitelisted(addr) {
+  return request.get(`${conf.ipWhitelistService}/${path}/${addr}`)
+    .then((_res) => _res.status === 200 && _res.body.allow === true)
+    .catch((err) => false);
```
jgraff2: I think we should distinguish between the whitelist app returning false and it returning an error or the request failing. If we can't allow the request because the whitelist app is down, that shouldn't be considered "Unauthorized".

jgraff2: I think we should distinguish between the whitelist app returning false and it returning an error or the request failing. If we can't allow the request because the whitelist app is down, that shouldn't be considered "Unauthorized".

realtime/setupSocketIO.js
```diff
@@ -116,7 +117,17 @@ function init(io, redisStore) {
           throw new Error(DISCO_MSG);
         }
 
-        rtUtils.isIpWhitelisted(ipAddress, ipWhitelist); // throws error
+        if (conf.ipWhitelistService) {
+          ipWhitelistUtils.isWhitelisted(ipAddress)
+            .then((allowed) => {
+              if (!allowed) {
+                socket.disconnect();
+                return;
```
jgraff2: Any reason for doing this here instead of throwing an error to be handled later like we do currently?

PR #1153, opened 2019-04-18
Event flapper

BrandonMurnane:


BrandonMurnane:
its not really how id have liked to fix it but it was working

PR #1160, opened 2019-05-06
Update Perspective and Room to set up socket.io connection to separate

iamigo:
real-time application if separate real-time application is configured
via REALTIME_APPLICATION environment variable. Default to current
behavior (i.e. serve real-time events from the main Refocus application
if no real-time application is configured).

Also: only create redis subscribers in the main app if NO separate
real-time application is configured.


tests/realtime/pubSub.js
```diff
@@ -67,6 +67,12 @@ describe('tests/realtime/pubsub.js >', () => {
     });
 
     it('subscribers receive all published messages', (done) => {
+      // No subscribers if running with separate real-time application
+      if (process.env.REALTIME_APPLICATION) {
```
jgraff2: Does this actually get set when running the tests?

cache/redisCache.js
```diff
@@ -61,14 +61,22 @@ if (featureToggles.isFeatureEnabled('enableRedisConnectionLogging')) {
...
+  }
+
   pubPerspectives.push(redis.createClient(rp, opts));
 });
 
-const subBot = redis.createClient(rconf.instanceUrl.pubsubBots, opts);
-subBot.subscribe(rconf.botChannelName);
+// Only create subscribers here if we're doing real-time events from main app
+let subBot;
+if (!process.env.REALTIME_APPLICATION) {
```
jgraff2: Any reason for not using a toggle?

tests/realtime/pubSub.js
iamigo: This new script in `package.json`:

```
    "test-realtime-application": "REALTIME_APPLICATION=https://abc.def npm run test-with-coverage -- tests/realtime/pubSub.js",
```

cache/redisCache.js
iamigo: Updated to set up and use toggle @jgraff2

PR #1164, opened 2019-05-08
Shared context works for bots outside room, but not for inactive bots

kriscfoster:
We have a requirement for a global bot (our sync bot) to receive data from bots inside a room. This means we need to open up shared context to work for bots outside room. We decided that a bot should be active to receive shared data from other bots.

- Switched the sharedContext findAll to look for active bots instead of just bots in the room
- Added a test to handle each of these cases


kriscfoster:
@iamigo Thank you for pointing this out. @cassiodias and I spoke about this today and we agree with you.

We are going to leave sharedContext by definition as: A way for bots within a room to share data amongst themselves. Opening this up to bots outside the room breaks this definition.

We are going to instead define this 'Sync-Bot' as not a bot anymore (as it isn't actually in a room), so nothing needs to be shared with it. Instead, it will just pull this information directly.

PR #1171, opened 2019-05-10
update api helper doGetWriters to NOT send back response - let the controllers do that

iamigo:
this will help us evolve towards better use of modularity via middleware

(will be evolving all the helpers to do the same over time)


BrandonMurnane:
if this includes the work from the doget pr should we just close that one ?

iamigo:
@BrandonMurnane I was trying to layer the PRs to keep the code reviews smaller, but because they all touch a lot of the same files they necessarily have all that overlap, so ideally we'd be able to merge the doGet and then rebase this one... so... it's complicated :(

PR #1173, opened 2019-05-12
remove short term toggle returnUser/RETURN_CREATEDBY_ON_TOKEN_INPUT

iamigo:
(retaining the "new" behavior)
update tests


tests/api/v1/subjects/post.js
```diff
@@ -142,6 +142,8 @@ describe(`tests/api/v1/subjects/post.js, POST ${path} >`, () => {
         const result = JSON.parse(res.text);
         expect(Object.keys(result)).to.contain('parentAbsolutePath');
         expect(result.parentAbsolutePath).to.equal.null;
+        expect(result).to.have.property('user');
```
pallavi2209: How is owner different than user here? _result_ has both owner and user with exactly same value. This probably applies to other endpoints as well.

iamigo: @pallavi2209 we have a user story "#spike figure out how we should clarify/clean up user associations #rf" to address this concern.

PR #1174, opened 2019-05-12
reject http API requests if requireHttps is enabled

iamigo:
instead of trying to redirect to https


tests/disableHttp/disableHttp.js
```diff
@@ -12,15 +12,19 @@
...
+const tu = require('../testUtils');
 const path = '/v1/users';
-const expect = require('chai').expect;
 
 describe('tests/disableHttp/disableHttp.js, http is disabled >', () => {
-  it('GET is redirected', (done) => {
+  before(() => tu.toggleOverride('requireHttps', true));
+  after(() => tu.toggleOverride('requireHttps', false));
+
+  it('GET is rejected', (done) => {
```
pallavi2209: I think both the tests are failing here because of 'Authentication Failed' since we are not providing token.
Also path does not specifically has http - does it defaults to http behind the scene?
We should check the error description as well to make sure we are testing the right thing.

iamigo: Great catch, @pallavi2209! Because we were loading the express module BEFORE setting the toggle override, it actually wasn't using the override. I have updated the test so that it loads the express module AFTER setting the toggle override, and also check the error message text.

tests/disableHttp/disableHttp.js
```diff
@@ -12,15 +12,24 @@
...
 
 const supertest = require('supertest');
-const api = supertest(require('../../express').app);
+const expect = require('chai').expect;
 const constants = require('../../api/v1/constants');
+const tu = require('../testUtils');
 const path = '/v1/users';
-const expect = require('chai').expect;
 
 describe('tests/disableHttp/disableHttp.js, http is disabled >', () => {
```
jgraff2: should also test that it isn't rejected when sent with https.

pallavi2209:
@iamigo Tests failing - same authentication error.

tests/disableHttp/disableHttp.js
iamigo: yeah i wanted to do that but it ended up being way too much effort, tried to not get bogged down in setting up ssl for tests

at least it doesn't test *less* than it was testing before

jgraff2: ok - you verified manually?

iamigo:
@pallavi2209 i'm not getting those errors locally... obvs. have to spend more time on this

PR #1176, opened 2019-05-20
no more paranoid db models

iamigo:
Remove "paranoid" setting from db models (exc. Users)
Remove isDeleted attribute (exc. Users)
Add migration to clean up the soft-deleted data and update the indices
Update tests

Note: we are keeping paranoid on for Users because we can't delete some of the soft-deleted user records due to referential integrity, where another record has a soft-deleted user associated with it as creator and/or owner.

Changes to /api/v1/helpers/nouns/*:
- pull "isDeleted" out of the list of "readOnlyFields"

Changes to /db/model/*:
- remove "paranoid" setting on model
- remove "isDeleted" field defn
- no longer call setIsDeleted in beforeDestroy
- remove "isDeleted" from list of fields in any index defns
- remove "isDeleted" from list of "ignoreAttributes" when calling publishObject

Migration:

Up
1. fixReferences - clear the collectorGroupId col for collectors which reference a soft-deleted collector group
1.  destroySoftDeleted - deletes any soft-deleted records from the db
1. removeOldIndices - deletes the indices which include deletedAt/isDeleted
1. removeFields - removes the deletedAt/isDeleted columns
1. dropAndAddUniqueIndices - drops and adds new indices without deletedAt/isDeleted - we have to do this here because the db tries to create the new unique indices on db startup, but fails because the soft-deleted records are still there, so now that they're gone, we can drop and recreate these

Down
1. recreateOldFields - adds back the deletedAt/isDeleted columns
1. removeNewIndices - removes the new indices without deletedAt/isDeleted, if they were created


api/v1/controllers/ssoconfig.js
```diff
@@ -80,22 +80,23 @@ module.exports = {
...
-        const err = new apiErrors.ResourceNotFoundError();
-        err.info = 'There is no sso config to delete.';
-        u.handleError(next, err, helper.modelName);
+        found = o;
+        return o.destroy();
       }
+
+      const err = new apiErrors.ResourceNotFoundError();
+      err.info = 'There is no sso config to delete.';
+      u.handleError(next, err, helper.modelName);
```
jgraff2: Won't this still allow the next "then" block to execute?

db/model/user.js
```diff
@@ -198,12 +195,7 @@ module.exports = function user(seq, dataTypes) {
             seq.models.Profile.findById(_this.profileId)
             .then((p) => {
               if (p) {
-                if (Number(p.isDeleted) === 0) {
-                  resolve();
-                } else {
-                  reject(new Error(
-                    `Profile ${_this.profileId} has been soft-deleted`));
-                }
+                resolve();
```
jgraff2: ok to change this even though we're keeping paranoid for users?

tests/db/model/aspect/delete.js
```diff
@@ -49,13 +45,10 @@ describe('tests/db/model/aspect/delete.js', () => {
...
       .then((o) => o.destroy())
-      .then((o) => {
-        if (o.deletedAt && o.isDeleted) {
-          done();
-        } else {
-          done(new Error('expecting it to be soft-deleted'));
-        }
-      })
+      .then((o) => Aspect.findOne({ where: { name: u.name } }))
+      .then((o) => Aspect.findOne({ where: { name: u.name } }))
```
jgraff2: duplicate line

migrations/20190520221710-no-paranoid.js
```diff
@@ -0,0 +1,300 @@
...
+ */
+'use strict';
+const Promise = require('bluebird');
+const db = require('../db/index');
+const IS = 'isDeleted';
+const AT = 'deletedAt';
+
+/**
+ * If a collector references a soft-deleted collector group, clear that
+ * collector group from the collector.
```
jgraff2: Do we need to do this from the generator side as well?
Are there any other associations we need to do this for?

migrations/20190520221710-no-paranoid.js
```diff
@@ -0,0 +1,300 @@
...
+    exec('Perspective', optsIs),
+    exec('Profile', optsIs),
+    exec('SSOConfig', optsAt),
+    exec('Token', optsIs),
+    exec('Collector', optsIs),
+    exec('GeneratorTemplate', optsIs),
+    exec('Lens', optsIs),
+    exec('Subject', optsIs),
+    exec('CollectorGroup', optsIs),
+  ]))
```
jgraff2: Promise.all will run the promises in parallel, so it's possible they won't be executed in the right order.

db/model/user.js
iamigo: Not likely to be a problem for us, but you're right. I'll keep paranoid on for profiles, too, for now.

tests/db/model/aspect/delete.js
iamigo: Wow how'd that get in there?

migrations/20190520221710-no-paranoid.js
iamigo: These were the only ones where deletion was blocked for referential integrity in either sandbox or prod, I tried to think through other scenarios where similar things could happen but couldn't come up with any.

migrations/20190520221710-no-paranoid.js
iamigo: Yeah at one point I had a sequence of Promise.alls but then I wrapped the whole list in a trx so it all happens atomically.

iamigo:
Thx for the great review, @jgraff2! I will make the changes on Monday.

PR #1178, opened 2019-05-24
split out realtime app config variables for perspectives and rooms

jgraff2:
Use different config variables to specify the realtime app for perspectives and rooms. This way we can move over perspectives but let rooms continue to use the main app until imc is ready to switch.

### DIFF
--- a/config/toggles.js
+++ b/config/toggles.js
@@ -118,12 +118,18 @@ const longTermToggles = {
   hideRoutes: environmentVariableTrue(pe, 'HIDE_ROUTES'),
 
   /*
-   * Use separate realtime application if the env var exists and is not equal
+   * Use separate realtime application for perspectives if the env var exists and is not equal
    * to "/".
    */
-  enableRealtimeApplication:
-    (pe.hasOwnProperty('REALTIME_APPLICATION') && pe.REALTIME_APPLICATION !== '/')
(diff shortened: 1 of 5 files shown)


jgraff2:
That's based on the toggle, which will be true if either env var is set.

iamigo:
> That's based on the toggle, which will be true if either env var is set.

I don't get it. Shouldn't we only set up an imc subscriber in the api application (redisCache.js lines 76-79) if the imc-specific toggle is enabled?

jgraff2:
I figured it wouldn't do any harm to have an extra subscriber if there are no clients connected, and it was easier this way. 
But yeah, it would be better to only setup the subscriber when necessary. I'll change it.

config/toggles.js
```diff
@@ -118,12 +118,18 @@ const longTermToggles = {
...
-    || (pe.hasOwnProperty('REALTIME_APPLICATION_IMC') && pe.REALTIME_APPLICATION_IMC !== '/'),
+  enableRealtimeApplication: pe.hasOwnProperty('REALTIME_APPLICATION')
+    && pe.REALTIME_APPLICATION !== '/',
+
+  /*
+   * Use separate realtime application for Imc rooms if the env var exists and is not equal
+   * to "/".
+   */
+  enableRealtimeApplicationImc: pe.hasOwnProperty('REALTIME_APPLICATION_IMC')
+    && pe.REALTIME_APPLICATION !== '/',
```
iamigo: do you mean

` . && pe.REALTIME_APPLICATION_IMC !== '/',`

?

PR #1180, opened 2019-05-29
Add publish tests for real-time events for subjects

iamigo:
Fix bugs uncovered by these new tests:
- do not publish an add event when creating a subject with isPublished=false
- when checking whether a subject's "isPublished" attribtue was changed. we were just looking at the current (new) value and the list of fields marked as changed by sequelize... but this causing problems for PUT operations... even if the PUT was NOT changing the isPublished attribute, if the field was included in the request body, then sequelize was including it in the list of changed fields. So I have updated the check to ALSO look at the old value only treat it as changed if the new value !== the old value
- when tags or absolutePath changes, we send a remove event then an add event so that filtering is correct - fix the remove event so that it is populated with the "old" values
- if a non-filtering attribute changes, we should be sending an update event (but we weren't)
- for bulk deletes, we were double-publishing - once from the db's afterDestroy and again from the bulkDeleteSubjects implementation, so remove the publish call from bulkDeleteSubjects and just let the db layer handle it


tests/publish/subjects.js
```diff
@@ -0,0 +1,792 @@
+/**
+ * Copyright (c) 2019, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or
+ * https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * tests/publish.js
```
jgraff2: tests/publish/subjects.js

PR #1181, opened 2019-06-04
Remove "paranoid" setting from db models (exc. Users & Profiles)

iamigo:
Remove isDeleted attribute (exc. Users & Profiles)
Add migration to clean up the soft-deleted data and update the indices
Update tests

Note: we are keeping paranoid on for Users because we can't delete some of the soft-deleted user records due to referential integrity, where another record has a soft-deleted user associated with it as creator and/or owner. And we're keeping it on for Profiles because we're keeping it on for Users.

Changes to /api/v1/helpers/nouns/*:

pull "isDeleted" out of the list of "readOnlyFields"
Changes to /db/model/*:

remove "paranoid" setting on model
remove "isDeleted" field defn
no longer call setIsDeleted in beforeDestroy
remove "isDeleted" from list of fields in any index defns
remove "isDeleted" from list of "ignoreAttributes" when calling publishObject

Migration:

Up
- fixReferences - clear the collectorGroupId col for collectors which reference a soft-deleted collector group
- destroySoftDeleted - deletes any soft-deleted records from the db
- removeOldIndices - deletes the indices which include deletedAt/isDeleted
- removeFields - removes the deletedAt/isDeleted columns

Down
- recreateOldFields - adds back the deletedAt/isDeleted columns
- removeNewIndices - removes the new indices without deletedAt/isDeleted, if they were created


iamigo:
FYI I have tested in devbox with copy of prod db

tests/db/model/sample/timeout.js
```diff
@@ -190,7 +190,7 @@ describe('tests/db/model/sample/timeout.js >', () => {
     .catch(done);
   });
 
-  it('simulate 10 seconds in the past', (done) => {
+  it('simulate 1tests/db/model/subject/delete.js0 seconds in the past', (done) => {
```
pallavi2209: mistyped?

pallavi2209:
@iamigo Just thinking.. do we need any changes for the data existing in Redis?

iamigo:
@pallavi2209 no, I don't think so, since we delete records from redis when the postgres record is soft-deleted

migrations/20190603225918-no-paranoid.js
```diff
@@ -0,0 +1,239 @@
...
+
+  /*
+   * Why this order?
+   * - Generator THEN Collector THEN CollectorGroup so we don't try to delete
+   *   a CollectorGroup which is referenced by a Collector or Generator
+   * - Generator THEN GeneratorTemplate so we don't try to delete a
+   *   GeneratorTemplate which is referenced by a Generator
+   * - Perpsective THEN Lens/Subject so we don't try to delete a Lens or
+   *   Subject which is referenced by a Perspective
+   */
```
jgraff2: >> Promise.all will run the promises in parallel, so it's possible they won't be executed in the right order.

> Yeah at one point I had a sequence of Promise.alls but then I wrapped the whole list in a trx so it all happens atomically.

Does that mean the order doesn't matter here anymore? If so can you update the comment?

PR #1182, opened 2019-06-04
collector integration tests: run every test with both bulk and by-sub…

jgraff2:
…ject collection, track request counts

Changes:
- run all tests twice, once for bulk collection and once for by-subject collection.
- track number of requests made to the mocked endpoints, so we can confirm the collector doesn't make redundant logins.

Corresponding collector changes:
https://github.com/salesforce/refocus-collector/pull/186

Tests won't pass in travis until collector is published.


tests/collector/integration.js
```diff
@@ -212,1238 +276,1559 @@ const nockConfig = [
...
-            .then(() => u.postGenerator(gen2))
-            .then(() => u.expectHeartbeatGens(coll2, { added: [gen2] })),
-        ))
-
-        .then(() => Promise.join(
-          u.expectBulkUpsertSamples(coll1, '60s', ['sub1|asp1']),
-          u.expectBulkUpsertSamples(coll2, '60s', ['sub2|asp2']),
-        ))
-      );
+        it('only gets token once, uses it for all subsequent requests', () => {
```
jgraff2: @pallavi2209 this test verifies that the collector doesn't send multiple login requests.
I confirmed that this fails with the old version of the collector (before your fix), with a login count of 6.

iamigo:
@jgraff2 should this change also update package.json to set minimum collector version to 1.2.1?

jgraff2:
Yeah, I always forget to do that, thanks for the reminder.

tests/collector/integration.js
pallavi2209: wow - these 3 tests! Thank you for the effort @jgraff2

PR #1183, opened 2019-06-06
Add new integer attribute “lensEventApiVersion” to the lens db model

iamigo:
Default to 1
Add migration
Add to swagger
Update tests
Make sure lens installation gets “lensEventApiVersion” attribute from lens.json. If not present, use default value 1 so we don't break anything for someone trying to install a v1 lens before they've had a chance to update it to 2.


tests/api/v1/lenses/post.js
```diff
@@ -208,4 +209,39 @@ describe('tests/api/v1/lenses/post.js >', () => {
...
   });
+
+  it('lens.json sets integer lensEventApiVersion', (done) => {
+    api.post(path)
+      .set('Authorization', token)
+      .field('name', 'testLens')
+      .field('description', 'test description')
+      .attach('library',
+        'tests/api/v1/lenses/lensZips/withValidLensEventApiVersion.zip')
+      .expect((res) => console.log(res.body))
```
jgraff2: console.log

PR #1186, opened 2019-06-11
Add more publish tests for real-time events

iamigo:
Add publish tests for real-time events for samples
Fix bugs found while testing (!!!)
- Use redis publishAsync and always call publishObject expecting a promise to be returned.
- In samples api controller, always call publish inside the promise chain.
- Fix the sampleStoreTimeout so it publishes sample update.
- Stop double-publishing sample delete events when an aspect is deleted.
- In subject db model, always call publish inside the promise chain.

Add publish tests for real-time events for botActions, botData, events
Fix bugs found while testing (!!!)
- When deleting botData, explicitly add something to the "changedKeys" array, otherwise the prepareToPublish fn determines that there's nothing to publish, so we were not emitting an event.

### DIFF
--- a/cache/sampleStoreTimeout.js
+++ b/cache/sampleStoreTimeout.js
@@ -20,6 +20,10 @@ const constants = require('../api/v1/constants');
 const fieldsToStringify = require('./sampleStore').constants.fieldsToStringify;
 const redisErrors = require('./redisErrors');
 const rconf = require('../config').redis;
+const publisher = require('../realtime/redisPublisher');
+const sampleEvent = require('../realtime/constants').events.sample;
+const helper = require('../api/v1/helpers/nouns/samples');
+const model = require('./models/samples');
 const IORedis = require('ioredis');
 const ioredisClient = new IORedis(rconf.instanceUrl.sampleStore);
 const featureToggles = require('feature-toggles');
@@ -76,11 +80,7 @@ function getSampleTimeoutComponents(samples, aspects, curr) {
       const sampleKey = sampleStore
         .toKey(sampleStore.constants.objectType.sample, samp.name);
       logInvalidHmsetValues(sampleKey, objToUpdate);
-      sampCmds.push([
-        'hmset',
-        sampleKey,
-        objToUpdate,
-      ]);
+      sampCmds.push(['hmset', sampleKey, objToUpdate]);
     }
   }
 
@@ -130,9 +130,7 @@ module.exports = {
       });
 
       aspectsSet.forEach((aspName) => {
-        commands.push(
-          ['hgetall', sampleStore.toKey(aspectType, aspName)]
-        );
+        commands.push(['hgetall', sampleStore.toKey(aspectType, aspName)]);
       });
 
       // ioredis use
@@ -180,10 +178,12 @@ module.exports = {
 
       return redisClient.batch(sampCmds).execAsync();
     })
-    .then(() => {
-      const res = { numberEvaluated, numberTimedOut, timedOutSamples };
-      return res;
-    })
+    .then(() => Promise.all(timedOutSamples.map((s) =>
+      model.getSample({ key: { value: s.name } }))))
+    .then((samples) => Promise.all(samples.map((s) =>
+      publisher.publishSample(s, helper.associatedModels.subject,
+        sampleEvent.upd, helper.associatedModels.aspect))))
+    .then(() => ({ numberEvaluated, numberTimedOut, timedOutSamples }))
     .catch((err) => {
       throw err;
     });
--- a/db/model/aspect.js
+++ b/db/model/aspect.js
@@ -47,11 +47,6 @@ module.exports = function aspect(seq, dataTypes) {
       primaryKey: true,
       defaultValue: dataTypes.UUIDV4,
     },
-    isDeleted: {
-      type: dataTypes.BIGINT,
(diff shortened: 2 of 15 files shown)


cache/sampleStoreTimeout.js
```diff
@@ -180,10 +178,12 @@ module.exports = {
 
       return redisClient.batch(sampCmds).execAsync();
     })
-    .then(() => {
-      const res = { numberEvaluated, numberTimedOut, timedOutSamples };
-      return res;
-    })
+    .then(() => Promise.all(timedOutSamples.map((s) =>
+      model.getSample({ key: { value: s.name } }))))
+    .then((samples) => Promise.all(samples.map((s) =>
```
jgraff2: Could simplify to `.map((s) =>`

cache/sampleStoreTimeout.js
```diff
@@ -180,10 +178,12 @@ module.exports = {
...
+      publisher.publishSample(s, helper.associatedModels.subject,
+        sampleEvent.upd, helper.associatedModels.aspect))))
```
jgraff2: Were we really not publishing on timeouts before? How does the multitable show timed-out samples? Do they get published elsewhere?

db/model/aspect.js
```diff
@@ -165,21 +165,7 @@ module.exports = function aspect(seq, dataTypes) {
...
-                samp, seq.models.Subject, sampleEventNames.del,
-                seq.models.Aspect
-              );
-            }
-          })
-        );
-
-        promiseArr.push(inst.checkGeneratorReferences('delete'));
-        return seq.Promise.all(promiseArr);
+        return inst.checkGeneratorReferences('delete');
```
jgraff2: Is the delete event published somewhere else?

tests/publish/samples.js
```diff
@@ -0,0 +1,342 @@
...
+    });
+  });
+
+  describe('patch aspect >', () => {
+    before(() => addSubject(`${tu.namePrefix}S3`, token)
+      .then(() => addAspect(`${tu.namePrefix}A2`, token))
+      .then(() => upsertSample(`${tu.namePrefix}S3|${tu.namePrefix}A2`, token))
+      .then(() => (subscribeTracker = [])));
+
+    it('edit tags, sample.remove followed by sample.add', (done) => {
```
jgraff2: Would be good to have another test to verify that no events are published for a patch without tags.

tests/publish/samples.js
```diff
@@ -0,0 +1,342 @@
...
+          const s1Body = s1[sampleEvents.add];
+          expect(s1Body).to.include.keys('createdAt', 'subjectId', 'aspectId',
+            'user', 'status', 'name', 'relatedLinks', 'provider', 'updatedAt',
+            'previousStatus', 'statusChangedAt', 'aspect', 'subject',
+            'absolutePath');
+          done();
+        });
+    });
+  });
+
```
jgraff2: subject delete/patch?

tests/publish/botActions.js
```diff
@@ -0,0 +1,144 @@
...
+              .end((err, res) => {
+                if (err) {
+                  return done(err);
+                }
+
+                expect(subscribeTracker).to.have.length(1);
+                const evt = JSON.parse(subscribeTracker[0]);
+                expect(evt).to.have.property(botActionEvents.del);
+                const evtBody = evt[botActionEvents.del];
+                console.log(evtBody)
```
pallavi2209: console.log

tests/publish/botActions.js
```diff
@@ -0,0 +1,144 @@
...
+      .catch(done);
+  });
+
+  afterEach(u.forceDelete);
+
+  after(tu.forceDeleteToken);
+
+  afterEach(() => subscribeTracker = []);
+
+  it('POST then PATCH then DELETE, subscriber gets events', (done) => {
```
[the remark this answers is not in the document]
iamigo: There are no PUT endpoints specified for those.

cache/sampleStoreTimeout.js
iamigo: You're right... it was too crazy to think that we weren't publishing sample timeout events... I couldn't trace the code path, but I do observe the real-time events getting sent in sandbox when a timeout happens... I'll have to revisit the code and see why they weren't getting sent in my tests.

db/model/aspect.js
iamigo: we're calling removeAspectRelatedSamples in the aspect afterDestroy hook

db/model/event.js
```diff
@@ -8,7 +8,7 @@
  */
 
 /**
- * db/model/event.js
+ * db/model/events.js
```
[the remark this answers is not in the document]
iamigo: POSTing to bulk upsert follows the same code path as posting to /v1/samples/upsert so I didn't think it was completely necessary (and it adds extra complication to the testing, too, so I just punted)... I can add those but it will take some time

iamigo:
@pallavi2209 @jgraff2 I think I have addressed your comments, please take another look

tests/publish/botActions.js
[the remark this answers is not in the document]
pallavi2209: I see PUT /botActions/{key}, but I guess its fine since PUT and PATCH follow similar code path in db layer.

PR #1187, opened 2019-06-12
Add api endpoints to allow associating collectorGroups with collectors and generators from either side.

jgraff2:
previously, the associations could only be set from these endpoints:
- generator -> collectorGroup (Generator post/patch/put)
- collectorGroup -> collectors (CollectorGroup post/patch/put)

this PR adds the ability to set it up this way as well:
- collector -> collectorGroup (Collector patch)
- collectorGroup -> generators (CollectorGroup post/patch/put)


pallavi2209:
@jgraff2 Tests failing?

PR #1189, opened 2019-06-19
Get samples: Correct sscan param passed

pallavi2209:
Missed the param when calling function recursively.

Found out when testing manually for /samples?status=Critical with prod data.

Hard to surface this in tests because sscan is called recursively when the number of samples are large enough (number of samples in one iteration is random when using sscan)


iamigo:
@jgraff2

PR #1191, opened 2019-06-25
Introduce toggle and logs for find samples

pallavi2209:
Changes:
- Added a toggle for recent optimization changes we did in #1184 
- Even If toggle is NOT enabled, introduced code to get samples by comma separated names.
- Introduced debug logs when toggle is enabled to figure out what is going on with prod data (the filtered gets are not working as expected in prod env (only))

### DIFF
--- a/cache/models/samples.js
+++ b/cache/models/samples.js
@@ -1488,14 +1488,26 @@ module.exports = {
          * that batch of commands.
          */
         else {
-          return redisClient.sortAsync(sortArgs)
+          return redisClient.smembersAsync(
+            sampleStore.constants.indexKey.sample)
+            .then((sampleKeys) => {
+              sampleKeys.sort();
+
+              if (!hasFilters) {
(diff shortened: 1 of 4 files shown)


cache/models/samples.js
```diff
@@ -1250,153 +1269,263 @@ module.exports = {
...
               .then(([samp, asp]) => samp && asp ? [samp, asp] : []);
           }
 
           if (!hasOrder || opts.order === ['name']) {
             nameFilterArr.sort();
-            nameFilterArr = modelUtils.applyLimitAndOffset(opts, nameFilterArr);
           }
 
           // multiple samples
           const sampleKeys = nameFilterArr.map((name) =>
```
pallavi2209: I introduce the option to get samples by comma separated name (even without toggle enabled). Helps with passing all the common tests (which include comma separated names) and it will help with users like Kyle who are only requesting specific number of samples for their integration.

PR #1192, opened 2019-06-26
removing unnecessary alias

cassiodias:
This change will remove the warning message

_Unhandled rejection SequelizeEagerLoadingError: RoomType is associated to Room using an alias. 
You've included an alias (type), but it does not match the alias(es) defined in your association (RoomType)_

FYI: this bug was introduced by the real-time changes it's not an IMC bug.
 
https://github.com/salesforce/refocus/pull/1068


jgraff2:
@cassiodias 
Are you sure this isn't necessary? I'm not sure that it is, but I generally include the "as" to be safe.

Looking at the model definitions, I think the problem is that the association fields are named "type" on both models - the field on the RoomType side should be something like "rooms".

cassiodias:
Roomtype and room are associated without alias, the fk name is already an
alias (don’t like that, but 🤷‍♂️) ... should be something like
room.roomtypeid.

This new method is adding parameters to the query and sequelize is
complaining because there is no alias definition on the original
association.

I have tried adding alias, but sequelize don’t accept alias and foreignKey
at the same time...

I’d suggest you to test that method for the real time feature. All the
tests are working and from IMC perspective rooms and bots are working.

ConalCosgrove:
@jgraff2 is this good to merge?

cassiodias:
👍🏻

PR #1193, opened 2019-06-27
socket client connection - handle auth error

jgraff2:


view/perspective/app.js
```diff
@@ -170,7 +170,10 @@ function setupSocketIOClient(persBody) {
     socket = _io.connect(`${_realtimeApplication}${namespace}`, options)
              .on('connect', function() {
                this.emit('auth', _userSession);
-             });
+             })
+             .on('auth error', (err) =>
```
iamigo: any tests for these here? (mostly I'm curious about whether socket.io supports spaces in the eventName)

jgraff2: This is tested on the realtime side, in the token auth PR.

PR #1196, opened 2019-07-09
Send real-time events on aspect add/edit/delete

iamigo:


realtime/utils.js
```diff
@@ -195,9 +198,16 @@ function perspectiveEmit(nspComponents, obj) {
...
-  if (isThisSubject(obj)) return applyFilter(subjectTagFilter, obj.tags);
+  const objectType = whatAmI(obj);
+  if (objectType === ObjectType.Subject) {
+    return applyFilter(subjectTagFilter, obj.tags);
+  }
+
+  if (objectType === ObjectType.Aspect) {
+    return applyFilter(aspectFilter, obj.aspect.name) &&
+      applyFilter(aspectTagFilter, obj.tags);
+  }
```
jgraff2: Can you also add these changes to the realtime app?

tests/publish/aspects.js
```diff
@@ -0,0 +1,685 @@
...
+                  'ownerId', 'rank', 'relatedLinks', 'tags', 'timeout',
+                  'updatedAt', 'valueLabel', 'valueType', 'warningRange');
+                expect(evtBody.new).to.have.property('name', name);
+                expect(evtBody.new).to.have.property('tags').to.deep.equal(tags);
+                done();
+              });
+          });
+      });
+
+    it('update description of published subject, subscriber gets update ' +
```
jgraff2: "published aspect"

tests/publish/aspects.js
```diff
@@ -0,0 +1,685 @@
...
+                'imageUrl', 'infoRange', 'isPublished', 'name', 'okRange',
+                'ownerId', 'rank', 'relatedLinks', 'tags', 'timeout',
+                'updatedAt', 'valueLabel', 'valueType', 'warningRange');
+              expect(evtBody1).to.have.property('name', newName);
+              done();
+            });
+        });
+    });
+
+    it('update tags of published subject, subscriber gets update event',
```
jgraff2: "published aspect"

tests/publish/aspects.js
```diff
@@ -0,0 +1,685 @@
...
+                'ownerId', 'rank', 'relatedLinks', 'tags', 'timeout',
+                'updatedAt', 'valueLabel', 'valueType', 'warningRange');
+              expect(evtBody1).to.have.property('name', newName);
+              done();
+            });
+        });
+    });
+
+    it('update tags of published aspect, subscriber gets update event',
+      (done) => {
```
jgraff2: For subjects, we send a delete and then add when tags are changed. Do we need to do that for aspects as well?

iamigo: Oh, I think we do need to do that for filtering... if we don't get the remove+add then if a tag change results in the aspect no longer being relevant for that perspective, it would never get the signal to remove. I will change that.

realtime/utils.js
iamigo: https://github.com/salesforce/refocus-real-time/pull/8

tests/publish/aspects.js
jgraff2: Also need to update test description.

PR #1203, opened 2019-07-24
Subject create|update|delete, update sample store with tags

pallavi2209:
On create:
Add tags set in sample store
samsto:tags:subjectAbsolutePath  -> set of tags

On update:
If published -> add tags set
If unpublished -> delete tags set
On tags change -> recreate tags set with new tags
On subject absolute path change -> delete old tags set, create new tags set

On delete:
Delete tags set

Tests for each use cases


pallavi2209:
This PR creates subject tag key in format: `samsto:tags:subjectAbsolutePath`.
Just occurred to me that its possible that aspect name could be same as subject absolute path (root subject). So we should instead have `samsto:tags:subject:subjectAbsolutePath` and `samsto:tags:aspect:aspectName`. Should I go ahead and update the PR?

iamigo:
Hmm collector integration test failures... ?

pallavi2209:
Weird..should be flappers because the merge before that passed. Let me check.

pallavi2209:
All checks passing now.

cassiodias:
@pallavi2209 @jgraff2 @iamigo can we merge this?

PR #1208, opened 2019-07-31
Add 'hasMultipartKey' attribute to botAction and botData

ConalCosgrove:
Since botAction and botData both have non-unique parameter `name`, we need to add the ```hasMultipartKey``` attribute to the helper files to allow the API to return multiple results when we query them by name. 

### Before: 
Screenshot 2019-07-31 at 11 36 53

### After
Screenshot 2019-07-31 at 11 35 43


iamigo:
failing tests?

PR #1219, opened 2019-08-29
Intercept events on perspective page to support v1 lenses

jgraff2:


view/perspective/app.js
```diff
@@ -115,34 +119,61 @@ function handleError(err) {
...
+ */
+function interceptEvent(eventTypeName, eventData) {
+  if (lensEventApiVersion < 2) {
+    if (eventTypeName === eventsQueue.eventType.INTRNL_SMPL_ADD) {
+      eventData.aspect = getTrackedAspect(eventData);
+    } else if (eventTypeName === eventsQueue.eventType.INTRNL_SMPL_DEL) {
+      eventData.aspect = getTrackedAspect(eventData);
+    } else if (eventTypeName === eventsQueue.eventType.INTRNL_SMPL_UPD) {
+      eventData.new.aspect = getTrackedAspect(eventData.new);
+      eventData.old.aspect = getTrackedAspect(eventData.new);
```
cassiodias: can be the case here as ```eventData.old.aspect = eventData.new.aspect``` ?

jgraff2: Yeah, we're only tracking the latest aspect, so we set both old and new to use that one. It's technically not completely the same as it would have been before, but I doubt there's any lens that relies on that behavior, and tracking the previous version of the aspect would significantly complicate things.

cassiodias: 👍🏻

PR #1221, opened 2019-09-09
Added bull p2

BrandonMurnane:


config/toggles.js
```diff
@@ -224,6 +224,12 @@ const shortTermToggles = {
 
   enableBullForBulkDelSubj: environmentVariableTrue(
     pe, 'ENABLE_BULL_FOR_BULK_DEL_SUBJ'),
+
+  enableBullForBulkPostEvents: environmentVariableTrue(
+    pe, 'ENABLE_BULL_FOR_BULK_POST_EVENTS'),
+
+  enableBullForCreateAuditEvents: environmentVariableTrue(
+    pe, 'ENABLE_BULL_FOR_CREATE_AUDIT_EVENTS'),
```
jgraff2: If we're going to have a bunch of these, it would be more manageable to set it to use a single env var with a comma-separated list of jobs to enable, like we do for activity logging.

jobQueue/jobWrapper.js
```diff
@@ -95,6 +97,62 @@ bulkDelSubQueue.on('completed', (job, jobResultObj) => {
...
+        .update(jobResultObj.recordCount, jobResultObj.queueTime);
+    }
+
+    /*
+    * The second argument should match the activity logging type in
+    * /config/activityLog.js
+    */
+    activityLogUtil.printActivityLogString(logObject, 'worker');
+  }
+});
```
jgraff2: This should be exactly the same for each job type. It would be better to define it as a separate function and pass that function in to each handler.

jobQueue/jobWrapper.js
```diff
@@ -265,6 +325,36 @@ function createPromisifiedJob(jobName, data, req) {
...
+    const jobPriority = calculateJobPriority(conf.prioritizeJobsFrom,
+      conf.deprioritizeJobsFrom, req);
+
+    return bulkPostEventsQueue.add(data, { priority: jobPriority })
+      .then((job) => {
+        job.type = jobType.bulkPostEvents;
+        logJobCreate(startTime, job);
+        return job;
+      });
+  }
```
jgraff2: If this is the same for each job, we should refactor to a separate function and pass in the job type.

jobQueue/jobWrapper.js
```diff
@@ -215,7 +273,9 @@ function calculateJobPriority(prioritize, deprioritize, req) {
     low: 'low',
   };
 
-  if (featureToggles.isFeatureEnabled('enableBullForBulkDelSubj')) {
+  if (featureToggles.isFeatureEnabled('enableBullForBulkDelSubj') ||
+   featureToggles.isFeatureEnabled('enableBullForBulkPostEvents') ||
+   featureToggles.isFeatureEnabled('enableBullForCreateAuditEvents')) {
```
jgraff2: If we use a single env var as I recommended above, that could also include a toggle for whether at least one bull job is enabled, so we wouldn't have to check every single one here.

package.json
```diff
@@ -57,6 +57,8 @@
     "test-disablehttp": "REQUIRE_HTTPS=true npm run test-with-coverage -- --recursive tests/disableHttp",
     "test-extra": "npm run test-api-log && npm run test-cache && npm run test-jobQueue && npm run test-view && npm run test-disablehttp && npm run test-cors && npm run test-limiter && npm run test-xfwd && npm run test-routes-enabled && npm run test-routes-disabled && npm run test-realtime-application",
     "test-jobQueue": "ENABLE_BULL_FOR_BULK_DEL_SUBJ=false npm run test-with-coverage -- --recursive tests/jobQueue && ENABLE_BULL_FOR_BULK_DEL_SUBJ=true npm run test-with-coverage -- --recursive tests/jobQueue ",
+    "test-postEventsBulk": "ENABLE_BULL_FOR_BULK_POST_EVENTS=false npm run test-with-coverage -- --recursive tests/api/v1/events && ENABLE_BULL_FOR_BULK_POST_EVENTS=true npm run test-with-coverage -- --recursive tests/api/v1/events ",
+    "test-createAudit": "ENABLE_BULL_FOR_CREATE_AUDIT_EVENTS=false npm run test-with-coverage -- --recursive tests/api/v1/auditEvents && ENABLE_BULL_FOR_CREATE_AUDIT_EVENTS=true npm run test-with-coverage -- --recursive tests/api/v1/auditEvents ",
```
jgraff2: These should be included in the existing test-jobQueue script: set all three toggles, then run all the tests at once.

package.json
```diff
@@ -72,7 +74,7 @@
     "watch": "npm-watch"
   },
   "dependencies": {
-    "@salesforce/refocus-collector": "^1.2.1",
+    "@salesforce/refocus-collector": "^1.3.0",
```
jgraff2: ??? - the collector is currently at version 1.2.1

BrandonMurnane: i dunno i just ran npm update on this

jobQueue/jobWrapper.js
```diff
@@ -93,7 +101,16 @@ bulkDelSubQueue.on('completed', (job, jobResultObj) => {
     */
     activityLogUtil.printActivityLogString(logObject, 'worker');
   }
-});
+}
+
+bulkDelSubQueue.on('completed', (job, jobResultObj) =>
+  logCompletedJob(job, jobResultObj));
```
jgraff2: Could simplify to "bulkDelSubQueue.on('completed', logCompletedJob)".

iamigo:
bulk delete tests failing? @BrandonMurnane

worker/jobProcessor.js
```diff
@@ -23,6 +24,16 @@ module.exports = {
         bulkDelSubQueue.process(Number(concurrency), job);
       }
 
+      if (featureToggles.isFeatureEnabled('enableBullForbulkPostEvents') &&
```
iamigo: type - should be upper-case "b" in "Bulk"

package.json
iamigo: I'm not sure why npm did this, but I think you need to back out this change, since there is no module matching that version spec -- https://www.npmjs.com/package/@salesforce/refocus-collector

BrandonMurnane: this has already been reverted @iamigo

iamigo:
Cool. Ready to approve once the tests are passing:

1) tests/jobQueue/v1/auditEvents/post.js
2) tests/jobQueue/v1/auditEvents/post.js
3) tests/jobQueue/v1/getBulkPostEventStatus.js
4) tests/jobQueue/v1/getBulkPostEventStatus.js

BrandonMurnane:
@iamigo having trouble with the tests as locally they seem to be passing fine when i run npm tests

BrandonMurnane:
@iamigo @jgraff2 if you could help look at this pr i cant seem to figure out even with others help why these two tests wont pass. I also dont wanna spend too much more time getting it working when bulk post events isnt actually used at all and should probably be removed

iamigo:
Hey @BrandonMurnane. I don't want you wasting time on this for no reason - if there is an actual plan to deprecate that API route, please make sure there's a triaged and prioritized story to do so, then go ahead and just `skip` those tests and add a comment that we're skipping them because we are deprecating that API route.

BrandonMurnane:
@jgraff2 @cassiodias can you re review plz

PR #1225, opened 2019-09-25
added catches to botdata updates to fix promise.all

BrandonMurnane:


db/helpers/botDataUtils.js
```diff
@@ -251,7 +251,7 @@ function updateValues(seq, instance) {
           });
 
           // Find bot data to update
-          promises.push(BotData.findOne(whereConst));
+          promises.push(BotData.findOne(whereConst).catch((err) => err));
```
ConalCosgrove: Could be useful to log out the error here just in case we run into issues here in the future

iamigo: `(err) => err` is a no-op... if that is intentional, it is useful to add a comment to indicate that it is intentional. Under what circumstances would this throw?

db/helpers/botDataUtils.js
```diff
@@ -270,7 +270,7 @@ function updateValues(seq, instance) {
       for (let i = ZERO; i < data.length; i += TWO) {
         if (data[i] && data[i].value) {
           newValue = combineValue(data[i].value, data[i + ONE]);
-          updates.push(data[i].update({ value: newValue }));
+          updates.push(data[i].update({ value: newValue }).catch((err) => err));
```
iamigo: Some comment as above.

BrandonMurnane: no-op ? These are to make it so a bad roomtype configuration wont break all of a bots shared contexts

iamigo: "no-op" meaning that function `(err) => err` doesn't actually execute anything

if there are circumstances where a bad roomtype configuration could cause findOne to throw, should the be logged so that the team can better understand how to avoid these "bad roomtype configurations"? or if the team already understands that, is there a user story to prevent such bad roomtype configurations from being persisted in a bad state in the first place?

BrandonMurnane: ah thanks for explaining. Was something i meant to ask in standup as what is the best way to log these things a simple console.log seems like itd be lost in the stream of refocus's logging.

iamigo: do not use console.log -- look at how other places in the app are using the refocus logging module now

PR #1226, opened 2019-09-25
Removing deactivation room from Refocus

cassiodias:
Deactivation job is running now in CaseBot every 5 min (configurable), there is no need for refocus loading deactivationRoom and execute the process.


cassiodias:
@iamigo The use case for Refocus deactivating a room is not valid anymore unless we can give access from Refocus to GUS; One option would be creating a "master" bot, however, that's not a priority, so Case bot was the option since we don't have any RoomType without a case and even not having Case bot in a room it still activating as a master bot checking the rooms.

iamigo:
Yeah, but my point is really that "we **don't** have any RoomType without a case" isn't the same as "we **can't** have any RoomType without a case"

PR #1230, opened 2019-09-27
Turning off room deactivation JOB

cassiodias:
Turning off the job in Refocus because now GUS/ORG62 must be checked before deactivating (TTx metrics). This work has been transferred to Case-Bot.

Added env var ENABLE_DEACTIVATE_ROOMS (already in all refocus environments in Heroku) - this will be false and inside setupIntervals this will get false under toggles for deactivateRooms job.

### DIFF
--- a/clock/scheduledJobs/deactivateRooms.js
+++ b/clock/scheduledJobs/deactivateRooms.js
@@ -11,12 +11,6 @@
  *
  * Gets an array of all rooms which should be deactivated, then deactivates
  * them and does a sync if this is enabled in the room settings.
- *
- * Note:
- * Deactivate room was removed from the job in Refocus because now
- * GUS/ORG62 must be checked before deactivating. This work has been
- * transferred to Case-Bot (even when there is no a case-bot in the room, it
- * still running as a backend service checking rooms to deactivate).
  */
 const moment = require('moment');
 const Op = require('sequelize').Op;
--- a/clock/scheduledJobs/deactivateRooms.js
+++ b/clock/scheduledJobs/deactivateRooms.js
@@ -28,6 +28,7 @@ const conf = require('../../config');
  * @returns {Promise} - Promise that room was deactivated.
  */
 function checkAndDeactivateRoom(room) {
+  console.log('test deactivation');
   // Getting most recent event for this room
   return dbEvent.findOne({
     where: { roomId: room.id },
--- a/clock/scheduledJobs/deactivateRooms.js
+++ b/clock/scheduledJobs/deactivateRooms.js
@@ -11,6 +11,12 @@
  *
  * Gets an array of all rooms which should be deactivated, then deactivates
  * them and does a sync if this is enabled in the room settings.
+ *
+ * Note:
+ * Deactivate room was removed from the job in Refocus because now
+ * GUS/ORG62 must be checked before deactivating. This work has been
+ * transferred to Case-Bot (even when there is no a case-bot in the room, it
+ * still running as a backend service checking rooms to deactivate).
  */
 const moment = require('moment');
 const Op = require('sequelize').Op;
--- a/clock/scheduledJobs/deactivateRooms.js
+++ b/clock/scheduledJobs/deactivateRooms.js
@@ -19,6 +19,7 @@ const dbRoom = require('../../db/index').Room;
 const dbEvent = require('../../db/index').Event;
 const dbBotAction = require('../../db/index').BotAction;
 const conf = require('../../config');
+const logger = require('@salesforce/refocus-logging-client');
 
 /**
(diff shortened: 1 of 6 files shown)


config.js
```diff
@@ -140,7 +140,6 @@ const clockJobConfig = {
   intervals: {
     checkMissedCollectorHeartbeat:
       String(collectorConfig.heartbeatIntervalMillis),
-    deactivateRooms: '5m',
```
cassiodias: Removing this from the list is enough to skip the JOB

clock/scheduledJobs/deactivateRooms.js
```diff
@@ -11,6 +11,12 @@
  *
  * Gets an array of all rooms which should be deactivated, then deactivates
  * them and does a sync if this is enabled in the room settings.
+ *
+ * Note:
+ * Deactivate room was removed from the job in Refocus because now
+ * GUS/ORG62 must be checked before deactivating. This work has been
```
iamigo: can't refer to these private resources in public github - can you state this more generically?

config.js
iamigo: I think we should be leaving all the code here, since it is still a valid use case, even though we don't want to use it in our instance, so we should just set overrides in config which would effectively disable the job

clock/scheduledJobs/deactivateRooms.js
cassiodias: 👍🏻

config.js
cassiodias: Keeping all the code, added env var ENABLE_DEACTIVATE_ROOMS - this will be false and inside setupIntervals this will get false under toggles for deactivateRooms job @iamigo

jgraff2:
It does need a new env var. The way it works most clock jobs are always on, and only the ones that specifically have a toggle defined for them require it to be set.

At the env var level, it probably would make sense to consolidate the toggles into a single env var, like we do for activity logging, but it would require changing the env vars in prod. When I set this up I was trying to keep it compatible with the current setup.

tests/clock/jobs/deactivateRooms.js
```diff
@@ -23,7 +23,7 @@ const ZERO = 0;
 const ONE = 1;
 const MOON_LANDING = '1969-07-20T20:18:00+00:00';
 
-describe('tests/clock/jobs/deactivateRooms.js >', () => {
+describe.skip('tests/clock/jobs/deactivateRooms.js >', () => {
```
jgraff2: No reason to skip if it's still possible to enable via toggle.

iamigo:
I think it would be better to employ a convention like "override the interval to zero to disable the job type" which would be applicable to all job types, rather than have to define a new env var every time we want to disable a job type

cassiodias:
@iamigo make sense - actually that was my first bet, but the way it works the interval will be set as 0 and I was not sure if that would be the best way. 

Would u agree as a separated story?

tests/clock/jobs/deactivateRooms.js
cassiodias: Nice @jgraff2, let me work on that. Thx

iamigo:
If zero already means something, does it mean it just repeats without any delay, i.e. it just runs non-stop? If that's the case, we should override that behavior to mean "do not run ever"

jgraff2:
If the interval is zero, the job will not be started.
We could disable it as-is with no code changes by setting CLOCK_JOB_INTERVAL_deactivateRooms to "0".

cassiodias:
@iamigo @jgraff2 I have to check tomorrow, if zero on CLOCK_JOB_INTERVAL_deactivateRooms works then its a refocus internal solution because ```setInterval``` by design with zero will run *continuously with minimum delay* - that's why I thought setting zero wouldn't be a solution.

jgraff2:
Yes, it's checked in the refocus code. I added a link to the line in the above comment.

jgraff2:
With regard to whether it's better to add a separate toggle or set the interval to zero, the difference is the default behavior. If we add a separate toggle, we're saying that by default this is off, and will only be enabled if needed. Whereas if we just set the interval to zero, we're saying that this should run by default but in this specific deployment it has been disabled temporarily.

So setting to zero isn't a replacement for having toggles - we do still need toggles for certain jobs, because we don't want something like activity stats to run by default. And we do want most other jobs to run by default - something like sample timeout isn't an extra optional job, it's an integral part of the refocus behavior.

cassiodias:
I completely agree. The point here is too far for a feature created by IMC because SRE was leaving rooms active - I am not sure that this is a Refocus public repo feature anyway. 

But I am ok setting to zero - the only concern was that setting as zero in the setInterval would fire that execution in the opposite way.

PR #1231, opened 2019-10-03
fix docker-compose and use alpine linux

esemeniuc:
Features:
-    creates focusdb so database can be initialized (previously missing)
-    alpine cuts download size down to 1/5th of the old size
-    uses sem versioning to keep up with security patches
-    eliminates the finicky sleep env variable so node always starts after postgres is ready


esemeniuc:
New version of old commit

PR #1232, opened 2019-10-08
Fixing bot shared data when its object

cassiodias:
* Fixing when in-depth like ${BotName.someAttrName.name}

### DIFF
--- a/db/helpers/botDataUtils.js
+++ b/db/helpers/botDataUtils.js
@@ -16,6 +16,7 @@
 const _ = require('lodash');
 const serialize = require('serialize-javascript');
 const Op = require('sequelize').Op;
+const logger = require('@salesforce/refocus-logging-client');
 const ZERO = 0;
 const ONE = 1;
 const TWO = 2;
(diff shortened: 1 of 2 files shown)


db/helpers/botDataUtils.js
```diff
@@ -190,98 +191,95 @@ function updateValues(seq, instance) {
...
+                      botId: botNames[syncBot],
+                      roomId: instance.roomId,
+                    },
+                  };
+                }
+              });
             });
+
+            // Find bot data to update
+            promises.push(BotData.findOne(whereConst));
```
BrandonMurnane: why did you remove the catch and logging?

db/helpers/botDataUtils.js
```diff
@@ -190,98 +191,95 @@ function updateValues(seq, instance) {
...
-      for (let i = ZERO; i < data.length; i += TWO) {
-        if (data[i] && data[i].value) {
-          newValue = combineValue(data[i].value, data[i + ONE]);
-          updates.push(data[i].update({ value: newValue }).catch((err) =>
-            logger.error('Shared Context error ', err)));
+        // Update bot data
+        for (let i = ZERO; i < data.length; i += TWO) {
+          if (data[i] && data[i].value) {
+            newValue = combineValue(data[i].value, data[i + ONE]);
+            updates.push(data[i].update({ value: newValue }));
```
BrandonMurnane: same as above

db/helpers/botDataUtils.js
cassiodias: was not here, already added - it was removed by mistake

BrandonMurnane: cool :)

PR #1233, opened 2019-10-09
Redis ops batching

jgraff2:
Set up redisOps to allow methods to be executed either standalone or as part of a batch. 
This makes it possible to always batch sequences of multiple redis ops, without having to set up separate thenable and batchable versions of each method.

The main purpose for this change is that, as part of the sample store format changes, we are adding additional redis commands on aspect/subject create/update/delete, and we want to be able to batch those to minimize the impact on the request times.

There are also several other places where performance could be improved by applying this in the future - see below.

Commits:
- redisOps batching - setup
  - batching logic
  - update existing redisOps methods to use the new format. Consolidate methods that previously had specific command or promise forms. 
- redisOps batching - main usages, test usages, small changes
  - update existing usages of redisOps methods:
    - Places where multiple redisOps commands are sent in parallel using Promise.all are updated to use batching
    - Places where multiple redisOps commands are sent as a batch are updated to use the new format
    - Places where redisOps commands are sent standalone are updated to always use a direct call, rather than having to use executeCommand() or separate methods for Promise and Cmd versions.

The updates to the usages are the minimum set of changes required to apply the new batching functionality to the existing logic; the commands sent to redis should all be exactly the same, but any that were previously sent in parallel will now be sent together as a batch.
They do not include other potential improvements that these changes make possible, such as:
- Places where we are sending direct redis commands as a batch that could be updated to use redisOps batching
- Places where redisOps methods are currently being called sequentially but aren't dependent on each other and could be updated to use batching.
- Places where we are sending direct redis commands that could be updated to use redisOps methods
- Places where we are sending multiple publish commands that could be updated to use batching.

### DIFF
--- a/api/v1/controllers/aspects.js
+++ b/api/v1/controllers/aspects.js
@@ -182,7 +182,7 @@ module.exports = {
         Array.from(userSet).forEach((user) => {
           cachedAspect.writers.push(user);
         });
-        return redisOps.hmSet('aspect', cachedAspect.name, cachedAspect);
+        return redisOps.setHash('aspect', cachedAspect.name, cachedAspect);
       }
 
       // reject the promise if the aspect is not found in the cache
@@ -275,7 +275,7 @@ module.exports = {
     .then((cachedAspect) => {
       if (cachedAspect) {
         cachedAspect.writers = [];
-        return redisOps.hmSet('aspect', cachedAspect.name, cachedAspect);
+        return redisOps.setHash('aspect', cachedAspect.name, cachedAspect);
       }
 
       // reject the promise if the aspect is not found in the cache
@@ -323,7 +323,7 @@ module.exports = {
       if (cachedAspect) {
         cachedAspect.writers = cachedAspect.writers
             .filter((writer) => writer !== userName);
-        return redisOps.hmSet('aspect', cachedAspect.name, cachedAspect);
+        return redisOps.setHash('aspect', cachedAspect.name, cachedAspect);
       }
 
       // reject the promise if the aspect is not found in the cache
--- a/api/v1/controllers/aspects.js
+++ b/api/v1/controllers/aspects.js
@@ -29,6 +29,8 @@ const doPut = require('../helpers/verbs/doPut');
 const u = require('../helpers/verbs/utils');
 const httpStatus = require('../constants').httpStatus;
 const redisOps = require('../../../cache/redisOps');
+const redisStore = require('../../../cache/sampleStore');
+const fieldsToStringify = redisStore.constants.fieldsToStringify.aspect;
 
 /**
  * Validates the given fields from request body or url.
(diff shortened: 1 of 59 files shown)


pallavi2209:
Good stuff!!

cache/redisOps.js
```diff
@@ -16,330 +16,388 @@ const redisStore = require('./sampleStore');
...
-  }
+  /**
+   * Add commands to an active batch, making their result available after the
+   * batch is executed.
+   *
+   * @param  {String} key - property name to be set on the result object
+   * @param  {Function} fn - function that adds commands to the batch
+   * @returns {RedisOps}
+   */
+  return(key, fn) {
```
iamigo: "return" seems like a confusing name for a function, surprised it even lets you use it, consider renaming for to avoid any confusion?

jgraff2: It's allowed because it's a method, so it will always be called like `batch.return('key', () => ...)`, and never like `return('key', () => ...)`. If you tried to do `const { return } = redisOps` it would fail.

Typical usage is:
```
this.batchCmds()
.doStuff()
.doOtherStuff()
.return('things', (batch) =>
  batch.getThings()
)
.exec()
.then(({ things }) => processThings(things))
```

I tried several different names for this and settled on "return" because it was the only one where it really felt clear what it was doing. It's the same idea as a normal function: you do a bunch of stuff, some of which may produce a value, but only the value from the statement marked with "return" will be used by subsequent code.

With something like:
```
.saveResult('things', (batch) =>
  batch.getThings()
)
```
or
```
.doThen('things', (batch) =>
  batch.getThings()
)
```
the intent seems less clear.

Any other name ideas?

jgraff2:
@iamigo

api/v1/controllers/aspects.js
```diff
@@ -165,25 +167,26 @@ module.exports = {
...
-            });
-            return redisOps.hmSet('aspect', cachedAspect.name, cachedAspect);
-          }
-
-          // reject the promise if the aspect is not found in the cache
-          return Promise.reject(new apiErrors.ResourceNotFoundError());
+      return u.findByKey(helper, params);
+    })
+    .then((o) => u.isWritable(req, o))
+    .then((o) => redisOps.getHash('aspect', o.name))
```
BrandonMurnane: just curious why are we hashing this now ?

jgraff2: In order to make all the ops methods work with the batching I had to enforce the rule that they have to return a single-stage promise (no multiple then chains). This was previously being done in two places: getValue and deleteSampleKeys. 

Those methods were deleted, and places that used them were updated to do multiple operations.

Previously getValue did getHash() then arrayObjsStringsToJson(). So here we do both of those directly, instead of calling getValue.

cache/redisOps.js
iamigo: OK, go ahead with it as is

PR #1235, opened 2019-10-30
Add bots to events without botId, publish events on room deactivate

ConalCosgrove:
## db/model/event.js
- added `attachBotsAndPublish` function
- when an event is created/updated, if it does not have a botId parameter, call attachBotsAndPublish

## db/model/room.js
- removed condition that caused events to only be emitted on room activation, and not room deactivation.


db/model/event.js
```diff
@@ -157,7 +167,36 @@ module.exports = function event(seq, dataTypes) {
     }, {
       override: true,
     });
+  };
 
+  Event.prototype.attachBotsAndPublish = function (eventType, changedKeys) {
```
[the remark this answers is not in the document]
ConalCosgrove: Removed 👍🏻

db/model/room.js
```diff
@@ -144,9 +144,7 @@ module.exports = function room(seq, dataTypes) {
         }
 
         if (instance.changed('active')) {
-          if (instance.active) {
-            instance.attachBotsAndPublish();
-          }
```
jgraff2: Is this a required change for using the new format/realtime app, or is it a fix to something that's been wrong all along?
This goes all the way back to when it was first added in https://github.com/salesforce/refocus/pull/939.

ConalCosgrove: This was implemented somewhere else, and I didn't realise. Reverted

PR #1236, opened 2019-10-31
Dynamic room type selection

ConalCosgrove:
Code change in rooms to allow for more granular roomType selection based on provided url parameters


view/rooms/app.js
```diff
@@ -1009,6 +1011,50 @@ function setupColumns() {
...
+ */
+function getRoomTypeFromMapping(urlParams) {
+  if (!_defaultRoomType) {
+    return '';
+  }
+  const urlParamNames = Object.keys(urlParams);
+  for (let i = 0; i < urlParamNames.length; i++){
+    const paramName = urlParamNames[i];
+    const urlParamValue = urlParams[paramName];
+    const newRoomType = urlParamValue && _roomTypeMapping[paramName] &&
```
cassiodias: is this returning a string or a boolean?

view/rooms/app.js
```diff
@@ -1009,6 +1011,50 @@ function setupColumns() {
...
+ * Infers roomType from url parameters other than roomType
+ * (used when it is not specified)
+ * @param {object} urlParams - JSON object of url params
+ * @returns {string} the first roomType name matching one of the url params
+ */
+function getRoomTypeFromMapping(urlParams) {
+  if (!_defaultRoomType) {
+    return '';
+  }
+  const urlParamNames = Object.keys(urlParams);
```
cassiodias: any reason why not using forEach from here?

ConalCosgrove: you can't break out of a for each, so you'd have to loop through the entire array and then return the found value. This way we can exit the loop as soon as a match is found

view/rooms/app.js
ConalCosgrove: A string - if it exists at the index

PR #1239, opened 2019-11-02
use redis instead of db for sub/asp existence check

jgraff2:
In order to improve bulkUpsert performance, use redis keys to check if aspects and subjects exist, instead of going to the db.


jgraff2:
@iamigo @pallavi2209

db/model/subject.js
```diff
@@ -225,9 +225,9 @@ module.exports = function subject(seq, dataTypes) {
                 batch.getSubjAspMapMembers(oldAbsPath)
               );
 
-              batch.removeSubjectTags(inst._previousDataValues);
+              batch.removeKeysForSubject(inst._previousDataValues);
               if (!tagsSetCreated) {
-                batch.setSubjectTags(inst);
+                batch.setupKeysForSubject(inst);
```
pallavi2209: Do we need these corresponding changes for aspect also?

jgraff2: For aspects we are already calling setupKeysForAspect(), all I had to do was add setAspectExists() to that.

PR #1245, opened 2019-11-13
Remove quotes from roomTypeMapping

ConalCosgrove:


view/rooms/index.pug
```diff
@@ -82,7 +82,7 @@ html
     var user = '#{user}';
     var userSession = '#{userSession}';
     var useNewNamespaceFormat = #{useNewNamespaceFormatImc};
-    var roomTypeMapping = '#{roomTypeMapping}';
+    var roomTypeMapping = #{roomTypeMapping};
     var defaultRoomType = '#{defaultRoomType}';
```
cassiodias: what's the difference without ''? user and user session are using ''

ConalCosgrove: It seems to read the variables in differently. With the quotes it ends up as an escaped string. But it's parsed as an object without the quotes (if the variable itself is an object)

PR #1256, opened 2020-07-08
Protecting failure when no aspect name on perspective

cassiodias:
* Added simple code guard
* npm bump for security fix
  * https://github.com/salesforce/refocus/network/alert/package-lock.json/npm/open


LongweiDeng:
That's fast! :) Thanks

PR #1261, opened 2020-07-08
setting return as undefined on beforeunload function

diogo-pessoa:
Description
---

Remove pop-up `leave site` upon user action for back button or refresh triggering the unload event. 

Tested with firefox, Chrome and safari from local Refocus, pop-up behaviour was suppressed after change. 

I've also tested the **TimelineBot** to confirm the post command is still triggered and broadcasts the _user left room_ Event that feeds the Bot with the users exiting room. 

```
User Joined - admin@refocus.admin
10:51 UTC | 07/08/2020

admin@refocus.admin has joined room.
User Left - admin@refocus.admin
10:51 UTC | 07/08/2020

admin@refocus.admin has left room.
User Joined - admin@refocus.admin
10:51 UTC | 07/08/2020

admin@refocus.admin has joined room.
User Left - admin@refocus.admin
10:51 UTC | 07/08/2020

admin@refocus.admin has left room.
```

Line 679:
Setting `undefined` as the return for function set on beforeUnload Event according to developer.mozilla some browsers might ignore the `null` and prompt the user with the `leave site?` confirmation pop-up.

> When this event returns (or sets the returnValue property to) a value other than null or undefined, the user will be prompted to confirm the page unload. In older browsers, the return value of the event is displayed in this dialog. Starting with Firefox 44, Chrome 51, Opera 38, and Safari 9.1, a generic string not under the control of the webpage will be shown instead of the returned string 
...Internet Explorer does not respect the null return value and will display this to users as "null" text. You have to use undefined to skip the prompt.

Line 1074:
> Typically, it is better to use `window.addEventListener()` and the `beforeunload` event, instead of `onbeforeunload`.

### Reference:
https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload


view/rooms/app.js
```diff
@@ -675,8 +675,8 @@ function confirmUserExit() {
     userId: _user.id,
     roomId: parseInt(ROOM_ID, 10),
   };
-
-  return u.postPromiseWithUrl(GET_EVENTS, events);
+  u.postPromiseWithUrl(GET_EVENTS, events);
+  return undefined;
```
LongweiDeng, line 679: Is there another way to handle this? 
Returning undefined doesn't sound so good to me. 
Can this be handled in the caller function?

diogo-pessoa: I could remove the return entirely, which will default to undefined, right? based on the documentation from mozilla, the idea is that the function we bind to the `beforeunload` event does not return anything so that the browser won't ask the user before he leaves the page. We don't really need the return here.  

What do you mean by caller function, sorry I didn't understand what you mean?

From what I saw here, we're biding the event  `beforeunload` to this `confirmUserExit` function and handling the data there, before user exits the room, 

Is there another place I can control the flow before this function?

cassiodias: Sounds weird but that's Mozilla reference 🤷

view/rooms/app.js
```diff
@@ -1071,7 +1071,7 @@ function buildRedirectUrl(url, roomId) {
   return redirectUrl;
 }
 
-window.onbeforeunload = confirmUserExit;
+window.addEventListener('beforeunload', confirmUserExit);
```
cassiodias, line 1074: is this where we generate the event "user has left the room" or here is where the pop message is generated?

view/rooms/app.js
cassiodias: Maybe removing the return will be less confusing if the caller is not expecting any return

diogo-pessoa: @LongweiDeng @cassiodias removed the return as suggested and renamed the function being called from **confirm** to **generateUserExitEvent**, hopefully that makes it clearer to understand what we're doing when the event 'beforeunload' is triggered.

view/rooms/app.js
diogo-pessoa: it happens inside the function `confirmUserExit` which as agreed I refactored to be called`generateUserExitEvent`

PR #1262, opened 2020-07-16
Adding passport-saml cert via env var

cassiodias:
Adding a generic SSO cert to passport-saml strategy when needed.

https://github.com/bergie/passport-saml#security-and-signatures


BrandonMurnane:
isnt there just an endpoint in api to change this ?

cassiodias:
Indeed! The source can be in the DB rather than env var... but the process requires changing the model, db, API doc and this is a quick and dirt for tomorrow if something changes on aloha process.

With better time frame we can get it from DB.

PR #1268, opened 2020-09-22
Call IMC button

diogo-pessoa:
IMC Paging Button
---

Added Paging button for IMC on Rooms view, Current implementation is leveraging the Oncall-Bot integration. Pointing to pager Duty Service :(IMC). 

Screenshot 2020-10-01 at 20 56 10


ConalCosgrove:
the package-lock.json is missing

ConalCosgrove:
is this going to go alongside the current header that includes the imc email and chatter? Or to replace it?

LongweiDeng:
From the screenshot, the button is too close to the deactivation button. It might cause mis-click. 
Could you increase the margin between the two element? @diogo-pessoa

diogo-pessoa:
> is this going to go alongside the current header that includes the imc email and chatter? Or to replace it?

@ConalCosgrove It should go along side, I have updated the image on the preview. (email and Chatter are env Vars, only show up if they're filled.)

view/rooms/app.js
```diff
@@ -873,6 +882,56 @@ function closeConfirmationModal() {
...
+ * @param  {Object} event - Clicked on toggle event.
+ */
+function pagingConfirmationModal(event) {
+  event.preventDefault();
+  pagingIMCConfirmationModal.setAttribute(
+    'style',
+    'display:block;'
+  );
+  pagingIMCConfirmationText.innerText =
+    'Would you like to Page IMC team for this room Incident?';
```
LongweiDeng, line 901: It sounds to me that paging IMC team is for solving the (case) incident. How can we rephrase the message to be more like paging IMC team to solve the technical issue in the room/bots?

cassiodias: yeah, its not room incident, it's any bad experience with IMC

diogo-pessoa: some ideias:
1- _Would you like to Call IMC team?_
2- _Would you like to Page IMC team?_
3-  _ cry for help?_ 😛 
4- _Are you sure you want to Page IMC team?_
5- _please don't page us it's late._ 😛 

Jokes apart, I'm open to suggestions.

ConalCosgrove:
Looks great! One last thing would be maybe having the button text be "Page IMC Team" rather than "Call IMC Team". Although it's open to discussion.

diogo-pessoa:
> Looks great! One last thing would be maybe having the button text be "Page IMC Team" rather than "Call IMC Team". Although it's open to discussion.

Good point, I started with "Page IMC Team" and on the story it suggest "Call". I also prefer the former.

BrandonMurnane:
does this work if for example the pagerduty bot isnt in the room ?

BrandonMurnane:
also missing readme and package.json bumps

diogo-pessoa:
> does this work if for example the pagerduty bot isn't in the room ?

@BrandonMurnane, I've tested from a room without the onCall bot,  as long as the backend of the oncall Bot is running, it receives handles the Action just fine. From what I read the front-end Code is a way to query&filter, which Service to Page, the paging Action itself happens on the back-end.

diogo-pessoa:
> also missing readme and package.json bumps

do we do that on refocus? Not saying we shouldn't. But I couldn't find a **changelog** section on the README and the package.json is `  "version": "1.0.0",`

BrandonMurnane:
Yeah sorry forgot that refocus doesn’t do that

diogo-pessoa:
Deployed to sandbox - Done.

cassiodias (approved):
I still think this button could be a bit far from deactivation and the alert message with a red color but it's fine for now

view/rooms/app.js
LongweiDeng: `Page IMC team if you experience critical issues/blockers using the IMC room and refreshing the browser doesn't help. `
I think we should emphasise that only using the Page button for critical issues. For non-critical issues, they can create a bugs or ping us instead. 
What do you guys think? @diogo-pessoa  @cassiodias  @ConalCosgrove @BrandonMurnane

LongweiDeng: Is it possible to have the message customisable via room type or env var? For example, we can have a generic default message but if the customised message is defined, it overrides it. 

Otherwise, changing the message may require code changes in future.

diogo-pessoa: > Is it possible to have the message customisable via room type or env var? For example, we can have a generic default message but if the customised message is defined, it overrides it.
> 
> Otherwise, changing the message may require code changes in future.

yeah, we can put the String as an EnvVar, I think that's the pattern we use to pass in config to a deployed app without code Change.

diogo-pessoa:
> I still think this button could be a bit far from deactivation and the alert message with a red color but it's fine for now

moved button to the left and set it to be red both on screen and on modal(YES)

view/rooms/app.js
diogo-pessoa: Changes are done I'll deploy on SandBox next week and see how the EnvVar behaves, then we can customize the wording for the modal.

view/rooms/app.js
```diff
@@ -906,11 +906,11 @@ function closePagingModal() {
     'style',
     'display:none;'
   );
-  alert('Thanks, IMC team was paged and it’s working hard to fix the issue');
 }
 
 function pageImcTeam() {
   closePagingModal();
+  alert('Thanks, IMC team was paged and it’s working hard to fix the issue');
```
LongweiDeng, line 913: Should the alert be after the post request? @diogo-pessoa

diogo-pessoa: yeah, I can move that.

view/rooms/app.js
```diff
@@ -931,6 +930,8 @@ function pageImcTeam() {
     ],
   };
   u.postPromiseWithUrl(GET_ACTIONS, serviceReq);
+  closePagingModal();
+  alert('Thanks, IMC team was paged and it’s working hard to fix the issue');
```
LongweiDeng, line 934: (I should have mentioned in the last comment) The message sounds we are already committed to fix the issue at the moment the page sent out. 
How about 'Thanks, IMC team was paged. We will look into it ASAP' or simply 'Thanks, IMC team was paged.'? 
What do others think?

diogo-pessoa: I agree, maybe the `Thanks, IMC team was paged. We will look into it ASAP` which has a sound more along the lines of setting the expectation. Worth hearing what other think.

LongweiDeng (approved):

ConalCosgrove (approved):

PR #1277, opened 2021-03-22
upgrade stack to Heroku-20

LongweiDeng:
Refocus is running in Heroku-16 stack, which will be end-of-life on May 1st, 2021.
We need to upgrade it to Heroku-20.

Ref: 
https://help.heroku.com/0S5P41DC/heroku-16-end-of-life-faq
https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack

### DIFF
--- a/app.json
+++ b/app.json
@@ -16,5 +16,6 @@
       "description": "A secret key for signing the session ID cookie.",
       "generator": "secret"
     }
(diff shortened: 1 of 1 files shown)


LongweiDeng:
closed due to cla issue

PR #1289, opened 2021-07-05
Avoid reparenting

ConalCosgrove:


iamigo (approved):
Looks great - but let's be careful when deploying this that someone is around to rollback if need be, since the node bump is a big one

PR #1296, opened 2021-10-22
Fix reparenting

BrandonMurnane:


iamigo (commented):
Looks pretty good, but can you add a test which shows the new error getting thrown when trying to reparent without perm?

iamigo (approved):
Great--thx!

kriscfoster (approved):

PR #1303, opened 2022-01-25
Validate login return URL

LongweiDeng:
Only accept return URL of protocol `http` or `https` as valid URL to avoid xss (`javascript` protocol). 
If the return URL is invalid, it will redirect to root path '/'.

### DIFF
--- a/view/authentication/login/app.js
+++ b/view/authentication/login/app.js
@@ -14,6 +14,7 @@
 'use strict'; // eslint-disable-line strict
 
 import request from 'superagent';
+import { isValidURL } from '../util';
 
 const input = document.loginform.elements;
 
@@ -33,23 +34,6 @@ function getQueryParams(qs) {
   }
 
   return params;
-}
-
-/**
- * Check if the input string is a valid URL of protocol http or https
- * This should invalidate the XSS URL redirect of protocal 'javascript:'
- * @param {String} url url string to be validated
- * @returns {Boolean} true if the URL is valid
- */
-function isValidURL(string) {
-  const decodedString = decodeURIComponent(string);
-  let url;
-  try {
-    url = new URL(decodedString);
-  } catch (_) {
-    return false;
-  }
-  return url.protocol === 'http:' || url.protocol === 'https:';
 }
 
 /**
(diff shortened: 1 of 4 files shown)


BrandonMurnane (approved):

view/authentication/login/app.js
```diff
@@ -35,6 +35,23 @@ function getQueryParams(qs) {
...
+ */
+function isValidURL(string) {
+  const decodedString = decodeURIComponent(string);
+  let url;
+  try {
+    url = new URL(decodedString);
+  } catch (_) {
+    return false;
+  }
+  return url.protocol === 'http:' || url.protocol === 'https:';
```
iamigo, line 52: This is a *very* small thing and not a blocker at all but just a super-tiny-micro-performance-enhancement: think about how this is going to be called the most... it will almost *always* be called with url.protocol === "https:" so that should be the first check on line 52 rather than the 2nd.

iamigo: also: add some unit tests for the isValidUrl fn with some different strings?

view/authentication/login/app.js
```diff
@@ -55,7 +72,7 @@ function sendData(jsonData) {
       document.getElementById('errorInfo').innerHTML = errorText;
     } else {
       returnUrl = getQueryParams(window.location.search.substring(1));
-      if (returnUrl.ru) {
+      if (isValidURL(returnUrl.ru)) {
```
iamigo, line 75: It might be useful to add some unit tests for getQueryParams because the correctness of that implementation will directly impact the correctness of the isValidUrl implementation

iamigo (approved):

view/authentication/login/app.js
LongweiDeng: I have a quick look at `getQueryParams`, it's probably not an elegant way to extract the query param, but it will work in the use case here, the correctness depends on the correctness of the previous redirect. If the queryParam is manually input and it should consider invalid, anyway. 
I would keep the change of the scope small for now (instead of refactoring it).

BrandonMurnane (approved):

PR #1340, opened 2023-06-21
Update redisCache.js

vswamidass-sfdc:
Enable support for Redis 7


sgundurao6 (approved):

vswamidass-sfdc:
Merging with Shreyas review.  Others are unavailable.  The system is down anyway right now.

mvaldez-sf (commented):
From my very limited knowledge, it certainly looks good per:

* https://github.com/redis/node-redis/blob/HEAD/docs/client-configuration.md#TLS

After it's back up, we may also want to consider a subsequent update to put these values in:

* https://github.com/salesforce/refocus/blob/master/config/redisConfig.js

